PR #14, opened 2021-05-31
Fetch libgccjit.so artifact to use it in CI

antoyo:


antoyo:
@bjorn3: There you go. You can see the errors in the CI now. Thanks for the idea about artifacts.

bjorn3:
Nice!

PR #34, opened 2021-06-18
Run UI tests in CI

antoyo:


test.sh
```diff
@@ -140,10 +140,10 @@ echo "[TEST] rust-lang/rust"
 
 rust_toolchain=$(cat rust-toolchain)
 
-git clone https://github.com/rust-lang/rust.git --depth=1 || true
+git clone https://github.com/rust-lang/rust.git || true
 cd rust
 git fetch
-#git checkout $(rustc -V | cut -d' ' -f3 | tr -d '(')
+git checkout $(rustc -V | cut -d' ' -f3 | tr -d '(')
 export RUSTFLAGS=
```
bjorn3, line 147: This will cause the tests to run with cg_llvm I think.

antoyo: I set `--rustc-args` [later](https://github.com/antoyo/rustc_codegen_gcc/pull/34/files#diff-3722d9ba8feb2d3feac8ce71a209a638d4b404e1c53f937188761181594023e2R176).

PR #36, opened 2021-06-23
Add dummy author info in case none is configured

dkm:
This is currently needed when the script is executed in a virgin environment as
git needs this to create the "initial commit" in the
build_sysroot/prepare_sysroot_src.sh script.


build_sysroot/prepare_sysroot_src.sh
```diff
@@ -20,6 +20,13 @@ git init
 echo "[GIT] add"
 git add .
 echo "[GIT] commit"
+
+# This is needed on virgin system where nothing is configured.
+# git really needs something here, or it will fail.
+# Even using --author is not enough.
+git config user.email || git config user.email "none@none.com"
```
bjorn3, line 27:
```suggestion
git config user.email || git config user.email "info@example.com"
```
none.com is an existing domain. example.com is reserved for example purposes.

dkm: Didn't know about example.com being reserved for this, thanks!

bjorn3: It has been reserved by [RFC2606](https://datatracker.ietf.org/doc/html/rfc2606).

dkm: TIL !

antoyo:
Thanks!

PR #46, opened 2021-07-03
Refactor scripts

antoyo:


antoyo:
@dkm Will you need an option to disable building the sysroot?

dkm:
I think that if we have the binary option disabled in CE, we should skip building the sysroot for nothing. I haven't refreshed my script to use your refactored script yet, but when I do, being able to skip sysroot building would be nice, yes :)

When the issue with too-big binary is fixed, then we'll build sysroot again and this option will be useless.

bjorn3:
> I think that if we have the binary option disabled in CE, we should skip building the sysroot for nothing.

The sysroot needs to be available even if the binary option is disabled. You need the crate metadata for at least libcore unless you want to manually define lang items like `Sized`, `Add`, ... It should be fine to use the cg_llvm built sysroot shipped with rustc though. cg_clif has the `--sysroot llvm` option for this. I recently rewrote most of cg_clif's build system in rust by the way for windows compatibility. Only the testing isn't rewritten yet. `./prepare.sh` and `./build.sh` and `./build/cargo.sh` are now available as `./y.rs prepare`, `./y.rs build` and `./build/cargo`. Feel free to copy the new build system for cg_gcc.

dkm:
I'm testing your new scripts. I'll simply use the `build.sh` script and remove few lines from mine.
Don't bother making the sysroot optional, I'll simply disable binary in compiler explorer until it is fixed and always use cg_gcc sysroot. Easier this way :)

dkm:
@antoyo as you are asking on another issue what's left before being able to push to CE: maybe this script? It's not a blocker as it's mostly a matter of removing few lines and having the same commands in `build.sh`, but it's cleaner this way. So if you merge this (or at least `build.sh`) I can finalize all 3 PRs and ask for reviews.

antoyo:
Ok, tell me when you're done reviewing this PR and I'll merge it!

PR #51, opened 2021-07-18
Remove libgccjit patches to not have them in rustc repository

antoyo:
Since the patches are being removed from this repository, I removed the note in the NOTICE file and the copy of the GPLv3 license.

@nellshamrell: Please tell me if this is ok. Thanks!


test.sh
```diff
@@ -1,5 +1,7 @@
 #!/bin/bash
 
+# TODO: rewrite to cargo-make or something like that to only rebuild the sysroot when needed?
```
bjorn3, line 3: cg_clif has a `config.txt` file with a `keep_sysroot` option that I can quickly enable.

antoyo:
Merging as is.
Still waiting for the confirmation from @nellshamrell, though.

nellshamrell:
Apologies for the delay, forwarding this to our counsel now.

antoyo:
@nellshamrell As a note, the git commit history has been squashed in [my PR in rustc](https://github.com/rust-lang/rust/pull/87260), so those patches will never show up in the git history.

nellshamrell:
@antoyo our counsel confirms that without the patches there's no need to include the NOTICE note or a copy of the GPLv3

PR #72, opened 2021-08-23
Implement basic inline asm support

Commeownist:
I decided to explicitly separate the declaration of out variables from the asm block generation, so now  we first generate the output variables, then register variables (as soon as `gccjit` adds support for them), then the asm block, then output operands, input operands, and clobbers. This system is simpler and less error prone, and also it should be easier to maintain and extend.

I did my best to document the system thoroughly.  

I've concentrated on X86 for now, barring SSE, AVX and so on (the rest of the backend doesn't support yet them anyway). Adding support for other architectures can _ideally_ be done simply by specifying right constraints in the match branches.

I resolved most of the TODOs in the code, and added a few of my own. 

Important questions we still need to resolve:

* Our `.intel_syntax noprefix` hack might break `m` memory constraints. I believe the only realistic way forward is to make gccjit aware of the att/intel dialect difference. I know that you can set the dialect globally (per file) but I'm not really sure if dialect can be set _per block_. Does gcc even support it? Worst case scenario: we might need to forbid different dialects in one project.
* How to align the stack if `nostack` is not specified?

And of course I'm waiting for gccjit to add register variables. Meanwhile, I'll start testing if the system is working correctly.


antoyo (commented):
Thanks!

Do you want me to look into adding register variable support to libgccjit soon or do you want to take a look at it?

src/asm.rs
```diff
@@ -1,254 +1,447 @@
...
-                    match operands[operand_idx] {
-                        InlineAsmOperandRef::Out { reg, place: Some(_), ..  } => {
+                    let mut push_to_template = |modifier, gcc_idx| {
+                        use std::fmt::Write;
+
+                        template_str.push('%');
+                        if let Some(modifier) = modifier {
+                            template_str.push(modifier);
+                        }
+                        write!(template_str, "{}", gcc_idx).unwrap();
```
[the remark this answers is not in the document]
Commeownist: I did it, but I'd like to point out that writing to a string never fails: out-of-memory situation is handled by crashing the program.

Commeownist:
> Do you want me to look into adding register variable support to libgccjit soon or do you want to take a look at it?

I'm loosely working on getting the asm working  (something doesn't work as expected) and covering it with tests ATM, so go ahead. 

May I ask how you debug the backend? I'd like to be able to take a look at the intermediate representation gccjit builds, any way to dump it? Doesn't look like `gcc_jit_context_dump_to_file` is present in the Rust binding.

src/asm.rs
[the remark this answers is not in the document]
antoyo: Yup, I know. I still prefer to avoid `unwrap()`, though.

Commeownist:
I apologize for my silent disappearance, had to take an emergency trip to Siberia (poor comms), forgot to warn you in the chaos. I'm sorry. I may be forced to take a few more in the next a few days, but I'll try to warn you so you won't have to wait for me. 

Now to the matter at hand.

> Change this pattern to dump a reproducer written in C in which you can call gcc_jit_context_dump_to_file. (If it would be simpler for you that I add this function to the binding, please tell me and I'll add it.)

You did add it! I just missed it trying to search for "dump" in `gcclib.rs` github web interface. Sometimes it's dumb as rock. 

The function in question is `gccjit::Context::dump_to_file` which is  easy to access from the Builder. That was the most useful debugging tool, mostly because I could call it from the "approximate troubling place", thus filtering out all the other irrelevant noise.

Now that I've found the error, I'm trying to test if the code actually works, adding a snippet to `tests/run/asm.rs` and failing miserably trying to run it. Did you know that trying `source` an `.sh` file that contains a syntax error (for example because the `CHANNEL` variable had not been defined and the string building pattern expected that it had been) just _crashes your bash process without any message whatsoever_? Lot's of fun, and that's not to mention all sorts of `[...] type: struct struct) has scope limited to function __gccGlobalInitcrc32fast_1becbe98_cgu_1 but was used within function _ZN4core9core_arch3x864sse219_mm_srli_si128_impl17h4a020dcaf28a33dcE [...]` followed by compiler hanging (of course) I'm trying to get through. 

The next thing I'm going to work on is __working__ build and test system, that's settled. 

Bottom line: I think we'll have to support the `llvm_asm!` is a very limited capacity, at least for a while (see the last commit). Is this fine? The alternative is patching sysroot and I'm not sure if it's  a good idea (for now).

antoyo:
No worries. I still haven't had the time to implement register variables anyway.

Yes, this is fine for `llvm_asm!`.

The tests in the `tests/` directory are indeed not working. Sorry about that: I forgot to add them in the CI. I'll fix them soon.

For this error `has scope limited to function __gccGlobalInitcrc32fast_1becbe98_cgu_1 but was used within function`, that's probably caused by some intermediate artifact (local variable in function __gccGlobalInitcrc32fast_1becbe98_cgu_1) used to initialize a global variable that would be used in another function. If you need more help to understandi this or anything else, feel free to ask!

Commeownist:
> The tests in the tests/ directory are indeed not working. 

I swear I _had_ them working at some point, but I forgot what exactly I did :( For now, I'll add a new "example".

> For this error has scope limited to function __gccGlobalInitcrc32fast_1becbe98_cgu_1 but was used within function, that's probably caused by some intermediate artifact (local variable in function __gccGlobalInitcrc32fast_1becbe98_cgu_1) used to initialize a global variable that would be used in another function. If you need more help to understand this or anything else, feel free to ask!

I'll look into it, but I'd rather leave them out of this PR. It's sizeable enough

antoyo:
Oh, actually the tests indeed run. (They didn't run in my custom branch on which I was working locally.) I added them in the CI in #73 .

Oh, I thought you had the global init error because of your changes. So, it's caused by something else?

Commeownist:
> Oh, I thought you had the global init error because of your changes. So, it's caused by something else?

I'm not really sure. I'm 99% sure that no changes to `src/asm.rs` that I've made could have affected anything outside `asm!` blocks. Although, I've made a lot of untracked changes to `*.sh` files (I've already commented on how I totally adore this system)  , so maybe it made the difference.

I'll try to checkout to master and see if `./prepare_build.sh && ./build.sh && cargo test && ./clean_all.sh` works.

Commeownist:
HA!

>  /usr/bin/ld: cannot find -lgccjit

And I do have `gcc_path`, and I'm 100% sure the path is correct , and it does contain `libgccjit.so`, `libgccjit.so.0`, and `libgccjit.so.0.0.1`.

antoyo:
Seems like `LIBRARY_PATH` is somehow not set. (I know the build scripts are a pain, but it would also be a pain to fix and I'm not sure what would be best here.)

How do you get this error? I'd need more details to help you.

antoyo:
Also, it seems there are 3 new failing UI tests in the CI for this branch. There are 104 failure in [this PR](https://github.com/antoyo/rustc_codegen_gcc/pull/73/checks?check_run_id=3450850819) while yours have 107.
Is this expected?

Commeownist:
> How do you get this error? I'd need more details to help you.

```sh
# your latest commit before the last PR
# everything below is done at this commit
git checkout 6f50986667debbfc67776304a8ee23fe0158613f
# my libgcc path
echo -n /home/user/dev/gccjit/build/gcc > gcc_path
./clean_all.sh
# taken from CI
./prepare_build.sh && ./build.sh && cargo test
```

> Seems like LIBRARY_PATH is somehow not set.

That was my guess too, but here's the error.
```
  = note: "cc" "-m64" "/home/user/dev/rustc_codegen_gcc/target/debug/deps/rustc_codegen_gcc-11b7a861feb3f326.104t5a81dgkvoyii.rcgu.o" "/home/user/dev/rustc_codegen_gcc/target/debug/deps/rustc_codegen_gcc-11b7a861feb3f326.11hpavimbvfisz6p.rcgu.o" "/home/user/dev/rustc_codegen_gcc/target/debug/deps/rustc_codegen_gcc-11b7a861feb3f326.12o0wwjrlomk9wqd.rcgu.o" "/home/user/dev/rustc_codegen_gcc/target/debug/deps/rustc_codegen_gcc-11b7a861feb3f326.12wqtqik124vtm77.rcgu.o" "/home/user/dev/rustc_codegen_gcc/target/debug/deps/rustc_codegen_gcc-11b7a861feb3f326.13g73xcnc66syty3.rcgu.o" "/home/user/dev/rustc_codegen_gcc/target/debug/deps/rustc_codegen_gcc-11b7a861feb3f326.13prv1npkmx72w0k.rcgu.o" "/home/user/dev/rustc_codegen_gcc/target/debug/deps/rustc_codegen_gcc-11b7a861feb3f326.1411q1qbzpiep350.rcgu.o" "/home/user/dev/rustc_codegen_gcc/target/debug/deps/rustc_codegen_gcc-11b7a861feb3f326.14m1xn8ukusnznn2.rcgu.o" "/home/user/dev/rustc_codegen_gcc/target/debug/deps/rustc_codegen_gcc-11b7a861feb3f326.14p61tmrbkr4wt44.rcgu.o" "/home/user/dev/rustc_codegen_gcc/target/debug/deps/rustc_codegen_gcc-11b7a861feb3f326.15pnx05xghi2gqdz.rcgu.o" "/home/user/dev/rustc_codegen_gcc/target/debug/deps/rustc_codegen_gcc-11b7a861feb3f326.18o9xa2rxqwnbrgf.rcgu.o" "/home/user/dev/rustc_codegen_gcc/target/debug/deps/rustc_codegen_gcc-11b7a861feb3f326.193m4qcdxrvh6c4s.rcgu.o" "/home/user/dev/rustc_codegen_gcc/target/debug/deps/rustc_codegen_gcc-11b7a861feb3f326.1ahkc8mq1pdf6gz3.rcgu.o" "/home/user/dev/rustc_codegen_gcc/target/debug/deps/rustc_codegen_gcc-11b7a861feb3f326.1b1owotahqebfhlh.rcgu.o" "/home/user/dev/rustc_codegen_gcc/target/debug/deps/rustc_codegen_gcc-11b7a861feb3f326.1cr43dwwu8lw464u.rcgu.o" "/home/user/dev/rustc_codegen_gcc/target/debug/deps/rustc_codegen_gcc-11b7a861feb3f326.1d14ewnmdzwtb9ag.rcgu.o" "/home/user/dev/rustc_codegen_gcc/target/debug/deps/rustc_codegen_gcc-11b7a861feb3f326.1dtf834phlni9zmb.rcgu.o" "/home/user/dev/rustc_codegen_gcc/target/debug/deps/rustc_codegen_gcc-11b7a861feb3f326.1fcrjt4mx8pkh1r6.rcgu.o" "/home/user/dev/rustc_codegen_gcc/target/debug/deps/rustc_codegen_gcc-11b7a861feb3f326.1fsac09x8soi4rgz.rcgu.o" "/home/user/dev/rustc_codegen_gcc/target/debug/deps/rustc_codegen_gcc-11b7a861feb3f326.1h6l1v1977ccsahq.rcgu.o" "/home/user/dev/rustc_codegen_gcc/target/debug/deps/rustc_codegen_gcc-11b7a861feb3f326.1i1rlw0hujbv28ko.rcgu.o" "/home/user/dev/rustc_codegen_gcc/target/debug/deps/rustc_codegen_gcc-11b7a861feb3f326.1i423cw55uty8y4i.rcgu.o" "/home/user/dev/rustc_codegen_gcc/target/debug/deps/rustc_codegen_gcc-11b7a861feb3f326.1ilepc2enygn93ay.rcgu.o" "/home/user/dev/rustc_codegen_gcc/target/debug/deps/rustc_codegen_gcc-11b7a861feb3f326.1j40ysvhp7aa05u2.rcgu.o" "/home/user/dev/rustc_codegen_gcc/target/debug/deps/rustc_codegen_gcc-11b7a861feb3f326.1kavaz4v4wnmryoo.rcgu.o" "/home/user/dev/rustc_codegen_gcc/target/debug/deps/rustc_codegen_gcc-11b7a861feb3f326.1ktdln44cozoa6mb.rcgu.o" "/home/user/dev/rustc_codegen_gcc/target/debug/deps/rustc_codegen_gcc-11b7a861feb3f326.1ky9mxpvwmua1l7q.rcgu.o" "/home/user/dev/rustc_codegen_gcc/target/debug/deps/rustc_codegen_gcc-11b7a861feb3f326.1mcw0smd9c2mrhu2.rcgu.o" "/home/user/dev/rustc_codegen_gcc/target/debug/deps/rustc_codegen_gcc-11b7a861feb3f326.1mejtpg03s52cle5.rcgu.o" "/home/user/dev/rustc_codegen_gcc/target/debug/deps/rustc_codegen_gcc-11b7a861feb3f326.1mokj008f6v66j60.rcgu.o" "/home/user/dev/rustc_codegen_gcc/target/debug/deps/rustc_codegen_gcc-11b7a861feb3f326.1mskeymg095nx2oj.rcgu.o" "/home/user/dev/rustc_codegen_gcc/target/debug/deps/rustc_codegen_gcc-11b7a861feb3f326.1nix9l2qjfmn806a.rcgu.o" "/home/user/dev/rustc_codegen_gcc/target/debug/deps/rustc_codegen_gcc-11b7a861feb3f326.1ocphs1cjv1dtjb2.rcgu.o" "/home/user/dev/rustc_codegen_gcc/target/debug/deps/rustc_codegen_gcc-11b7a861feb3f326.1or4f17il3xfhiy8.rcgu.o" "/home/user/dev/rustc_codegen_gcc/target/debug/deps/rustc_codegen_gcc-11b7a861feb3f326.1pqvdd3m6t1tq604.rcgu.o" "/home/user/dev/rustc_codegen_gcc/target/debug/deps/rustc_codegen_gcc-11b7a861feb3f326.1qcvh5q0c01txftq.rcgu.o" "/home/user/dev/rustc_codegen_gcc/target/debug/deps/rustc_codegen_gcc-11b7a861feb3f326.1r37f4kdqi69hk0j.rcgu.o" "/home/user/dev/rustc_codegen_gcc/target/debug/deps/rustc_codegen_gcc-11b7a861feb3f326.1rtz97ube17mdo2l.rcgu.o" "/home/user/dev/rustc_codegen_gcc/target/debug/deps/rustc_codegen_gcc-11b7a861feb3f326.1seyo1m06r1hsagw.rcgu.o" "/home/user/dev/rustc_codegen_gcc/target/debug/deps/rustc_codegen_gcc-11b7a861feb3f326.1tadrwubrxeo27v.rcgu.o" "/home/user/dev/rustc_codegen_gcc/target/debug/deps/rustc_codegen_gcc-11b7a861feb3f326.1tnxiirp6ev2q31x.rcgu.o" "/home/user/dev/rustc_codegen_gcc/target/debug/deps/rustc_codegen_gcc-11b7a861feb3f326.1uyldubyf84wvaz5.rcgu.o" "/home/user/dev/rustc_codegen_gcc/target/debug/deps/rustc_codegen_gcc-11b7a861feb3f326.1v1r0grly7yd0l8x.rcgu.o" "/home/user/dev/rustc_codegen_gcc/target/debug/deps/rustc_codegen_gcc-11b7a861feb3f326.1w20m4h4zdqdb8j7.rcgu.o" "/home/user/dev/rustc_codegen_gcc/target/debug/deps/rustc_codegen_gcc-11b7a861feb3f326.1wwxtcgyicb3bg5n.rcgu.o" "/home/user/dev/rustc_codegen_gcc/target/debug/deps/rustc_codegen_gcc-11b7a861feb3f326.1xblsz02m67ezhwk.rcgu.o" "/home/user/dev/rustc_codegen_gcc/target/debug/deps/rustc_codegen_gcc-11b7a861feb3f326.1yf7ystq8mlo03bp.rcgu.o" "/home/user/dev/rustc_codegen_gcc/target/debug/deps/rustc_codegen_gcc-11b7a861feb3f326.1zy06lo69mkdroqa.rcgu.o" "/home/user/dev/rustc_codegen_gcc/target/debug/deps/rustc_codegen_gcc-11b7a861feb3f326.20yrkbq5dmd0nmgc.rcgu.o" "/home/user/dev/rustc_codegen_gcc/target/debug/deps/rustc_codegen_gcc-11b7a861feb3f326.21q46sghnbv42grf.rcgu.o" "/home/user/dev/rustc_codegen_gcc/target/debug/deps/rustc_codegen_gcc-11b7a861feb3f326.22g081h8kjzhgknw.rcgu.o" "/home/user/dev/rustc_codegen_gcc/target/debug/deps/rustc_codegen_gcc-11b7a861feb3f326.22p7lpimwb5diidr.rcgu.o" "/home/user/dev/rustc_codegen_gcc/target/debug/deps/rustc_codegen_gcc-11b7a861feb3f326.22ta6p539ua2ab3t.rcgu.o" "/home/user/dev/rustc_codegen_gcc/target/debug/deps/rustc_codegen_gcc-11b7a861feb3f326.231jacbrzglmbg37.rcgu.o" "/home/user/dev/rustc_codegen_gcc/target/debug/deps/rustc_codegen_gcc-11b7a861feb3f326.23bbdgoduxqkla17.rcgu.o" "/home/user/dev/rustc_codegen_gcc/target/debug/deps/rustc_codegen_gcc-11b7a861feb3f326.23yv26ca0u14312l.rcgu.o" "/home/user/dev/rustc_codegen_gcc/target/debug/deps/rustc_codegen_gcc-11b7a861feb3f326.25kmoxzjvl9n4iu1.rcgu.o" "/home/user/dev/rustc_codegen_gcc/target/debug/deps/rustc_codegen_gcc-11b7a861feb3f326.263ugzdwaafsodsp.rcgu.o" "/home/user/dev/rustc_codegen_gcc/target/debug/deps/rustc_codegen_gcc-11b7a861feb3f326.275d7bu9s31cuz7y.rcgu.o" "/home/user/dev/rustc_codegen_gcc/target/debug/deps/rustc_codegen_gcc-11b7a861feb3f326.275fkmpaeje1cl59.rcgu.o" "/home/user/dev/rustc_codegen_gcc/target/debug/deps/rustc_codegen_gcc-11b7a861feb3f326.280sa16t5t9qrs2b.rcgu.o" "/home/user/dev/rustc_codegen_gcc/target/debug/deps/rustc_codegen_gcc-11b7a861feb3f326.2alnzggispniglcz.rcgu.o" "/home/user/dev/rustc_codegen_gcc/target/debug/deps/rustc_codegen_gcc-11b7a861feb3f326.2awks5nz4njbayy6.rcgu.o" "/home/user/dev/rustc_codegen_gcc/target/debug/deps/rustc_codegen_gcc-11b7a861feb3f326.2baqlcb286gyq5cm.rcgu.o" "/home/user/dev/rustc_codegen_gcc/target/debug/deps/rustc_codegen_gcc-11b7a861feb3f326.2c4z8y6q464ht8fn.rcgu.o" "/home/user/dev/rustc_codegen_gcc/target/debug/deps/rustc_codegen_gcc-11b7a861feb3f326.2d3oerjs2knek5q1.rcgu.o" "/home/user/dev/rustc_codegen_gcc/target/debug/deps/rustc_codegen_gcc-11b7a861feb3f326.2du1byty0vbwgosc.rcgu.o" "/home/user/dev/rustc_codegen_gcc/target/debug/deps/rustc_codegen_gcc-11b7a861feb3f326.2ftwhjojjmxwt5jx.rcgu.o" "/home/user/dev/rustc_codegen_gcc/target/debug/deps/rustc_codegen_gcc-11b7a861feb3f326.2gaxpwjlkmjsecms.rcgu.o" "/home/user/dev/rustc_codegen_gcc/target/debug/deps/rustc_codegen_gcc-11b7a861feb3f326.2ggk16lsg9c4acof.rcgu.o" "/home/user/dev/rustc_codegen_gcc/target/debug/deps/rustc_codegen_gcc-11b7a861feb3f326.2gvx798gor730qkt.rcgu.o" "/home/user/dev/rustc_codegen_gcc/target/debug/deps/rustc_codegen_gcc-11b7a861feb3f326.2ha14r13g759ummn.rcgu.o" "/home/user/dev/rustc_codegen_gcc/target/debug/deps/rustc_codegen_gcc-11b7a861feb3f326.2hjiloy1pgpnhhrt.rcgu.o" "/home/user/dev/rustc_codegen_gcc/target/debug/deps/rustc_codegen_gcc-11b7a861feb3f326.2i7uom791suzwu8o.rcgu.o" "/home/user/dev/rustc_codegen_gcc/target/debug/deps/rustc_codegen_gcc-11b7a861feb3f326.2ixk0871jm797onh.rcgu.o" "/home/user/dev/rustc_codegen_gcc/target/debug/deps/rustc_codegen_gcc-11b7a861feb3f326.2jestwvhg0u4aevj.rcgu.o" "/home/user/dev/rustc_codegen_gcc/target/debug/deps/rustc_codegen_gcc-11b7a861feb3f326.2jjogui9azzc63n4.rcgu.o" "/home/user/dev/rustc_codegen_gcc/target/debug/deps/rustc_codegen_gcc-11b7a861feb3f326.2p197gaqe0lpe1z4.rcgu.o" "/home/user/dev/rustc_codegen_gcc/target/debug/deps/rustc_codegen_gcc-11b7a861feb3f326.2phfv3x2raqhpbo1.rcgu.o" "/home/user/dev/rustc_codegen_gcc/target/debug/deps/rustc_codegen_gcc-11b7a861feb3f326.2q262zci6y9wqqm5.rcgu.o" "/home/user/dev/rustc_codegen_gcc/target/debug/deps/rustc_codegen_gcc-11b7a861feb3f326.2q7piwa71yn5egld.rcgu.o" "/home/user/dev/rustc_codegen_gcc/target/debug/deps/rustc_codegen_gcc-11b7a861feb3f326.2qu7fu7j0ked1dmk.rcgu.o" "/home/user/dev/rustc_codegen_gcc/target/debug/deps/rustc_codegen_gcc-11b7a861feb3f326.2qzf8m63fndo85z5.rcgu.o" "/home/user/dev/rustc_codegen_gcc/target/debug/deps/rustc_codegen_gcc-11b7a861feb3f326.2r8ohjj5j1yky6r6.rcgu.o" "/home/user/dev/rustc_codegen_gcc/target/debug/deps/rustc_codegen_gcc-11b7a861feb3f326.2rb503d1pzbrsvfl.rcgu.o" "/home/user/dev/rustc_codegen_gcc/target/debug/deps/rustc_codegen_gcc-11b7a861feb3f326.2rhxj0n9nauhw1zk.rcgu.o" "/home/user/dev/rustc_codegen_gcc/target/debug/deps/rustc_codegen_gcc-11b7a861feb3f326.2rpurohqjl8k415s.rcgu.o" "/home/user/dev/rustc_codegen_gcc/target/debug/deps/rustc_codegen_gcc-11b7a861feb3f326.2rz82nnyawuu8nxt.rcgu.o" "/home/user/dev/rustc_codegen_gcc/target/debug/deps/rustc_codegen_gcc-11b7a861feb3f326.2sl6wk7rcp5fuein.rcgu.o" "/home/user/dev/rustc_codegen_gcc/target/debug/deps/rustc_codegen_gcc-11b7a861feb3f326.2sy5r2pwwyp3pa64.rcgu.o" "/home/user/dev/rustc_codegen_gcc/target/debug/deps/rustc_codegen_gcc-11b7a861feb3f326.2t8nvgzftz0wnofc.rcgu.o" "/home/user/dev/rustc_codegen_gcc/target/debug/deps/rustc_codegen_gcc-11b7a861feb3f326.2tnkw0wj4nglb5vw.rcgu.o" "/home/user/dev/rustc_codegen_gcc/target/debug/deps/rustc_codegen_gcc-11b7a861feb3f326.2v5i4osm5hlkxeig.rcgu.o" "/home/user/dev/rustc_codegen_gcc/target/debug/deps/rustc_codegen_gcc-11b7a861feb3f326.2w1fqsp0katr5yjv.rcgu.o" "/home/user/dev/rustc_codegen_gcc/target/debug/deps/rustc_codegen_gcc-11b7a861feb3f326.2waq2meoxk2ad027.rcgu.o" "/home/user/dev/rustc_codegen_gcc/target/debug/deps/rustc_codegen_gcc-11b7a861feb3f326.2wtj744jjzzm8eqx.rcgu.o" "/home/user/dev/rustc_codegen_gcc/target/debug/deps/rustc_codegen_gcc-11b7a861feb3f326.2xrtz9lorcr11iyg.rcgu.o" "/home/user/dev/rustc_codegen_gcc/target/debug/deps/rustc_codegen_gcc-11b7a861feb3f326.2ymnkewkvi6evtcn.rcgu.o" "/home/user/dev/rustc_codegen_gcc/target/debug/deps/rustc_codegen_gcc-11b7a861feb3f326.2yw2iww5gxr7ymxr.rcgu.o" "/home/user/dev/rustc_codegen_gcc/target/debug/deps/rustc_codegen_gcc-11b7a861feb3f326.3005bj8vysqhpa52.rcgu.o" "/home/user/dev/rustc_codegen_gcc/target/debug/deps/rustc_codegen_gcc-11b7a861feb3f326.30dvo4ly8bsl211a.rcgu.o" "/home/user/dev/rustc_codegen_gcc/target/debug/deps/rustc_codegen_gcc-11b7a861feb3f326.30p6v0hesxc26pwj.rcgu.o" "/home/user/dev/rustc_codegen_gcc/target/debug/deps/rustc_codegen_gcc-11b7a861feb3f326.32ewc5zko0fqypt5.rcgu.o" "/home/user/dev/rustc_codegen_gcc/target/debug/deps/rustc_codegen_gcc-11b7a861feb3f326.3363uuz5g5yt9ym7.rcgu.o" "/home/user/dev/rustc_codegen_gcc/target/debug/deps/rustc_codegen_gcc-11b7a861feb3f326.33fh6hiz3l2ba15n.rcgu.o" "/home/user/dev/rustc_codegen_gcc/target/debug/deps/rustc_codegen_gcc-11b7a861feb3f326.33roqoxh8gkoep7c.rcgu.o" "/home/user/dev/rustc_codegen_gcc/target/debug/deps/rustc_codegen_gcc-11b7a861feb3f326.349960twdqvo5ib3.rcgu.o" "/home/user/dev/rustc_codegen_gcc/target/debug/deps/rustc_codegen_gcc-11b7a861feb3f326.34atpyqaytxmqlgv.rcgu.o" "/home/user/dev/rustc_codegen_gcc/target/debug/deps/rustc_codegen_gcc-11b7a861feb3f326.366s9ct3mmqtq9w8.rcgu.o" "/home/user/dev/rustc_codegen_gcc/target/debug/deps/rustc_codegen_gcc-11b7a861feb3f326.377o6g8238fi8l5s.rcgu.o" "/home/user/dev/rustc_codegen_gcc/target/debug/deps/rustc_codegen_gcc-11b7a861feb3f326.37albshiyc1cwxue.rcgu.o" "/home/user/dev/rustc_codegen_gcc/target/debug/deps/rustc_codegen_gcc-11b7a861feb3f326.37vsoliuoidqre60.rcgu.o" "/home/user/dev/rustc_codegen_gcc/target/debug/deps/rustc_codegen_gcc-11b7a861feb3f326.3be2v8that0dvjxm.rcgu.o" "/home/user/dev/rustc_codegen_gcc/target/debug/deps/rustc_codegen_gcc-11b7a861feb3f326.3c75bgmy1z366y4m.rcgu.o" "/home/user/dev/rustc_codegen_gcc/target/debug/deps/rustc_codegen_gcc-11b7a861feb3f326.3c8nbwtgb42n2aal.rcgu.o" "/home/user/dev/rustc_codegen_gcc/target/debug/deps/rustc_codegen_gcc-11b7a861feb3f326.3crkvodbjknn00ey.rcgu.o" "/home/user/dev/rustc_codegen_gcc/target/debug/deps/rustc_codegen_gcc-11b7a861feb3f326.3eb18hvnutdqexo4.rcgu.o" "/home/user/dev/rustc_codegen_gcc/target/debug/deps/rustc_codegen_gcc-11b7a861feb3f326.3ejp24o8uflgv69q.rcgu.o" "/home/user/dev/rustc_codegen_gcc/target/debug/deps/rustc_codegen_gcc-11b7a861feb3f326.3g5zg8fmrwhtvzen.rcgu.o" "/home/user/dev/rustc_codegen_gcc/target/debug/deps/rustc_codegen_gcc-11b7a861feb3f326.3jchbh5i0bf7tika.rcgu.o" "/home/user/dev/rustc_codegen_gcc/target/debug/deps/rustc_codegen_gcc-11b7a861feb3f326.3lc159dsgdrdudp7.rcgu.o" "/home/user/dev/rustc_codegen_gcc/target/debug/deps/rustc_codegen_gcc-11b7a861feb3f326.3lqz5nywgpffx9mj.rcgu.o" "/home/user/dev/rustc_codegen_gcc/target/debug/deps/rustc_codegen_gcc-11b7a861feb3f326.3muaa4f0rsk6afel.rcgu.o" "/home/user/dev/rustc_codegen_gcc/target/debug/deps/rustc_codegen_gcc-11b7a861feb3f326.3o350zvppkn0eijn.rcgu.o" "/home/user/dev/rustc_codegen_gcc/target/debug/deps/rustc_codegen_gcc-11b7a861feb3f326.3o4wco8mdn2lugmc.rcgu.o" "/home/user/dev/rustc_codegen_gcc/target/debug/deps/rustc_codegen_gcc-11b7a861feb3f326.3o5p62kaxsxw8jqy.rcgu.o" "/home/user/dev/rustc_codegen_gcc/target/debug/deps/rustc_codegen_gcc-11b7a861feb3f326.3oseyjx1s35ojr15.rcgu.o" "/home/user/dev/rustc_codegen_gcc/target/debug/deps/rustc_codegen_gcc-11b7a861feb3f326.3p49a706307wcqjq.rcgu.o" "/home/user/dev/rustc_codegen_gcc/target/debug/deps/rustc_codegen_gcc-11b7a861feb3f326.3rfn719l4hz38dx3.rcgu.o" "/home/user/dev/rustc_codegen_gcc/target/debug/deps/rustc_codegen_gcc-11b7a861feb3f326.3rrl4wocp79d1oz5.rcgu.o" "/home/user/dev/rustc_codegen_gcc/target/debug/deps/rustc_codegen_gcc-11b7a861feb3f326.3t7rl8y7dn6501y2.rcgu.o" "/home/user/dev/rustc_codegen_gcc/target/debug/deps/rustc_codegen_gcc-11b7a861feb3f326.3tgtd6qerpv7yyro.rcgu.o" "/home/user/dev/rustc_codegen_gcc/target/debug/deps/rustc_codegen_gcc-11b7a861feb3f326.3uac1ttze6poelqz.rcgu.o" "/home/user/dev/rustc_codegen_gcc/target/debug/deps/rustc_codegen_gcc-11b7a861feb3f326.3w8csodb49k1yvy8.rcgu.o" "/home/user/dev/rustc_codegen_gcc/target/debug/deps/rustc_codegen_gcc-11b7a861feb3f326.3wgd29tu523khpry.rcgu.o" "/home/user/dev/rustc_codegen_gcc/target/debug/deps/rustc_codegen_gcc-11b7a861feb3f326.3whfjr5w628gpqa6.rcgu.o" "/home/user/dev/rustc_codegen_gcc/target/debug/deps/rustc_codegen_gcc-11b7a861feb3f326.3wvd07vy0pehign9.rcgu.o" "/home/user/dev/rustc_codegen_gcc/target/debug/deps/rustc_codegen_gcc-11b7a861feb3f326.3z1tk1xc5u9jtt00.rcgu.o" "/home/user/dev/rustc_codegen_gcc/target/debug/deps/rustc_codegen_gcc-11b7a861feb3f326.411yb9hvjjoyey17.rcgu.o" "/home/user/dev/rustc_codegen_gcc/target/debug/deps/rustc_codegen_gcc-11b7a861feb3f326.4151xshqmfeb3uv4.rcgu.o" "/home/user/dev/rustc_codegen_gcc/target/debug/deps/rustc_codegen_gcc-11b7a861feb3f326.42ybbnz3wzj6ajy0.rcgu.o" "/home/user/dev/rustc_codegen_gcc/target/debug/deps/rustc_codegen_gcc-11b7a861feb3f326.436gs1i5bl99f7kb.rcgu.o" "/home/user/dev/rustc_codegen_gcc/target/debug/deps/rustc_codegen_gcc-11b7a861feb3f326.45km4h7de0t5jv43.rcgu.o" "/home/user/dev/rustc_codegen_gcc/target/debug/deps/rustc_codegen_gcc-11b7a861feb3f326.46gnkfeeuc0whpk5.rcgu.o" "/home/user/dev/rustc_codegen_gcc/target/debug/deps/rustc_codegen_gcc-11b7a861feb3f326.47f56e471j8i4t1q.rcgu.o" "/home/user/dev/rustc_codegen_gcc/target/debug/deps/rustc_codegen_gcc-11b7a861feb3f326.47m2gck85ief0qvr.rcgu.o" "/home/user/dev/rustc_codegen_gcc/target/debug/deps/rustc_codegen_gcc-11b7a861feb3f326.487r7gw8l4pkjcgx.rcgu.o" "/home/user/dev/rustc_codegen_gcc/target/debug/deps/rustc_codegen_gcc-11b7a861feb3f326.48bhu2xhrcxe7xqb.rcgu.o" "/home/user/dev/rustc_codegen_gcc/target/debug/deps/rustc_codegen_gcc-11b7a861feb3f326.4a45mwfmfacy2u1t.rcgu.o" "/home/user/dev/rustc_codegen_gcc/target/debug/deps/rustc_codegen_gcc-11b7a861feb3f326.4alq6scb8daojays.rcgu.o" "/home/user/dev/rustc_codegen_gcc/target/debug/deps/rustc_codegen_gcc-11b7a861feb3f326.4btqugk76uxe92ib.rcgu.o" "/home/user/dev/rustc_codegen_gcc/target/debug/deps/rustc_codegen_gcc-11b7a861feb3f326.4c2ofynku1m4vhoo.rcgu.o" "/home/user/dev/rustc_codegen_gcc/target/debug/deps/rustc_codegen_gcc-11b7a861feb3f326.4ccue4ezi7zthguu.rcgu.o" "/home/user/dev/rustc_codegen_gcc/target/debug/deps/rustc_codegen_gcc-11b7a861feb3f326.4dcj7ge2106xujxs.rcgu.o" "/home/user/dev/rustc_codegen_gcc/target/debug/deps/rustc_codegen_gcc-11b7a861feb3f326.4ea2ufj11dobv1ln.rcgu.o" "/home/user/dev/rustc_codegen_gcc/target/debug/deps/rustc_codegen_gcc-11b7a861feb3f326.4efncd2c4ljw5xmp.rcgu.o" "/home/user/dev/rustc_codegen_gcc/target/debug/deps/rustc_codegen_gcc-11b7a861feb3f326.4g6tuz16txuyhb93.rcgu.o" "/home/user/dev/rustc_codegen_gcc/target/debug/deps/rustc_codegen_gcc-11b7a861feb3f326.4g7el3g4d7zhtkyz.rcgu.o" "/home/user/dev/rustc_codegen_gcc/target/debug/deps/rustc_codegen_gcc-11b7a861feb3f326.4h1fshhi4tyaavzw.rcgu.o" "/home/user/dev/rustc_codegen_gcc/target/debug/deps/rustc_codegen_gcc-11b7a861feb3f326.4h3hyr0g4s47ku03.rcgu.o" "/home/user/dev/rustc_codegen_gcc/target/debug/deps/rustc_codegen_gcc-11b7a861feb3f326.4k24uutyignptcnw.rcgu.o" "/home/user/dev/rustc_codegen_gcc/target/debug/deps/rustc_codegen_gcc-11b7a861feb3f326.4k34ae2i3ak7jthn.rcgu.o" "/home/user/dev/rustc_codegen_gcc/target/debug/deps/rustc_codegen_gcc-11b7a861feb3f326.4kdbrz4lvrr1h0wc.rcgu.o" "/home/user/dev/rustc_codegen_gcc/target/debug/deps/rustc_codegen_gcc-11b7a861feb3f326.4kjnyetdpo655p29.rcgu.o" "/home/user/dev/rustc_codegen_gcc/target/debug/deps/rustc_codegen_gcc-11b7a861feb3f326.4llxxpa2ozwntjsa.rcgu.o" "/home/user/dev/rustc_codegen_gcc/target/debug/deps/rustc_codegen_gcc-11b7a861feb3f326.4lrcg5xeunizdk26.rcgu.o" "/home/user/dev/rustc_codegen_gcc/target/debug/deps/rustc_codegen_gcc-11b7a861feb3f326.4mp013llfx3eg5mc.rcgu.o" "/home/user/dev/rustc_codegen_gcc/target/debug/deps/rustc_codegen_gcc-11b7a861feb3f326.4nztq9sm2it8zb44.rcgu.o" "/home/user/dev/rustc_codegen_gcc/target/debug/deps/rustc_codegen_gcc-11b7a861feb3f326.4pvnytho6vc00ls3.rcgu.o" "/home/user/dev/rustc_codegen_gcc/target/debug/deps/rustc_codegen_gcc-11b7a861feb3f326.4q730omw9xjaxaf7.rcgu.o" "/home/user/dev/rustc_codegen_gcc/target/debug/deps/rustc_codegen_gcc-11b7a861feb3f326.4qiyhcaeuqvol4uq.rcgu.o" "/home/user/dev/rustc_codegen_gcc/target/debug/deps/rustc_codegen_gcc-11b7a861feb3f326.4qo9z0aeyuonk57s.rcgu.o" "/home/user/dev/rustc_codegen_gcc/target/debug/deps/rustc_codegen_gcc-11b7a861feb3f326.4rqb9xy6gb7gksdp.rcgu.o" "/home/user/dev/rustc_codegen_gcc/target/debug/deps/rustc_codegen_gcc-11b7a861feb3f326.4s3sryeyyilormwz.rcgu.o" "/home/user/dev/rustc_codegen_gcc/target/debug/deps/rustc_codegen_gcc-11b7a861feb3f326.4s7ox6wso1zk0g8w.rcgu.o" "/home/user/dev/rustc_codegen_gcc/target/debug/deps/rustc_codegen_gcc-11b7a861feb3f326.4sdehu284hxmnimh.rcgu.o" "/home/user/dev/rustc_codegen_gcc/target/debug/deps/rustc_codegen_gcc-11b7a861feb3f326.4ta8007zp1wmcysb.rcgu.o" "/home/user/dev/rustc_codegen_gcc/target/debug/deps/rustc_codegen_gcc-11b7a861feb3f326.4vq2sfrpoq8kv4ce.rcgu.o" "/home/user/dev/rustc_codegen_gcc/target/debug/deps/rustc_codegen_gcc-11b7a861feb3f326.4w4mrf6ybiovz9vz.rcgu.o" "/home/user/dev/rustc_codegen_gcc/target/debug/deps/rustc_codegen_gcc-11b7a861feb3f326.4xu4mnv02pqr6h6j.rcgu.o" "/home/user/dev/rustc_codegen_gcc/target/debug/deps/rustc_codegen_gcc-11b7a861feb3f326.4xylm464epjwz9vu.rcgu.o" "/home/user/dev/rustc_codegen_gcc/target/debug/deps/rustc_codegen_gcc-11b7a861feb3f326.4y0o7595qcmw3xj3.rcgu.o" "/home/user/dev/rustc_codegen_gcc/target/debug/deps/rustc_codegen_gcc-11b7a861feb3f326.4yzh0ydga86f7fbk.rcgu.o" "/home/user/dev/rustc_codegen_gcc/target/debug/deps/rustc_codegen_gcc-11b7a861feb3f326.4zjhy70ov79ezbxj.rcgu.o" "/home/user/dev/rustc_codegen_gcc/target/debug/deps/rustc_codegen_gcc-11b7a861feb3f326.50bvlxkjn9qx0059.rcgu.o" "/home/user/dev/rustc_codegen_gcc/target/debug/deps/rustc_codegen_gcc-11b7a861feb3f326.50dnfxts9jn9b2se.rcgu.o" "/home/user/dev/rustc_codegen_gcc/target/debug/deps/rustc_codegen_gcc-11b7a861feb3f326.50j1owdh0h2dqh1o.rcgu.o" "/home/user/dev/rustc_codegen_gcc/target/debug/deps/rustc_codegen_gcc-11b7a861feb3f326.50jkz4k07pj0b7x.rcgu.o" "/home/user/dev/rustc_codegen_gcc/target/debug/deps/rustc_codegen_gcc-11b7a861feb3f326.512hxta5178qc5v.rcgu.o" "/home/user/dev/rustc_codegen_gcc/target/debug/deps/rustc_codegen_gcc-11b7a861feb3f326.51ghrep98k1fuird.rcgu.o" "/home/user/dev/rustc_codegen_gcc/target/debug/deps/rustc_codegen_gcc-11b7a861feb3f326.537yqbw2w6n6va75.rcgu.o" "/home/user/dev/rustc_codegen_gcc/target/debug/deps/rustc_codegen_gcc-11b7a861feb3f326.53bblcy3ijoyhif3.rcgu.o" "/home/user/dev/rustc_codegen_gcc/target/debug/deps/rustc_codegen_gcc-11b7a861feb3f326.53gxt1jg3zltg99n.rcgu.o" "/home/user/dev/rustc_codegen_gcc/target/debug/deps/rustc_codegen_gcc-11b7a861feb3f326.541eclj392jdmvxj.rcgu.o" "/home/user/dev/rustc_codegen_gcc/target/debug/deps/rustc_codegen_gcc-11b7a861feb3f326.54nolxg2weukejx3.rcgu.o" "/home/user/dev/rustc_codegen_gcc/target/debug/deps/rustc_codegen_gcc-11b7a861feb3f326.552spneai8h21r7w.rcgu.o" "/home/user/dev/rustc_codegen_gcc/target/debug/deps/rustc_codegen_gcc-11b7a861feb3f326.5579i98on6q5asqu.rcgu.o" "/home/user/dev/rustc_codegen_gcc/target/debug/deps/rustc_codegen_gcc-11b7a861feb3f326.55ke5lrmwfpqb2qs.rcgu.o" "/home/user/dev/rustc_codegen_gcc/target/debug/deps/rustc_codegen_gcc-11b7a861feb3f326.56a8wbaxq1wmsgeo.rcgu.o" "/home/user/dev/rustc_codegen_gcc/target/debug/deps/rustc_codegen_gcc-11b7a861feb3f326.57p297153r35wavf.rcgu.o" "/home/user/dev/rustc_codegen_gcc/target/debug/deps/rustc_codegen_gcc-11b7a861feb3f326.57vsz4wffmqtfc3h.rcgu.o" "/home/user/dev/rustc_codegen_gcc/target/debug/deps/rustc_codegen_gcc-11b7a861feb3f326.58lugedi03dtuuf0.rcgu.o" "/home/user/dev/rustc_codegen_gcc/target/debug/deps/rustc_codegen_gcc-11b7a861feb3f326.59chyhl23yqr9i9d.rcgu.o" "/home/user/dev/rustc_codegen_gcc/target/debug/deps/rustc_codegen_gcc-11b7a861feb3f326.5ab23bke0rw181n7.rcgu.o" "/home/user/dev/rustc_codegen_gcc/target/debug/deps/rustc_codegen_gcc-11b7a861feb3f326.5ajewofrh45j6bq5.rcgu.o" "/home/user/dev/rustc_codegen_gcc/target/debug/deps/rustc_codegen_gcc-11b7a861feb3f326.5amg6kduljnblvj5.rcgu.o" "/home/user/dev/rustc_codegen_gcc/target/debug/deps/rustc_codegen_gcc-11b7a861feb3f326.5b9e08g6rhb0yglk.rcgu.o" "/home/user/dev/rustc_codegen_gcc/target/debug/deps/rustc_codegen_gcc-11b7a861feb3f326.5ck3s4t02ga4dfze.rcgu.o" "/home/user/dev/rustc_codegen_gcc/target/debug/deps/rustc_codegen_gcc-11b7a861feb3f326.5ckg55r6t63lvwtw.rcgu.o" "/home/user/dev/rustc_codegen_gcc/target/debug/deps/rustc_codegen_gcc-11b7a861feb3f326.5crab0bb06az9ebh.rcgu.o" "/home/user/dev/rustc_codegen_gcc/target/debug/deps/rustc_codegen_gcc-11b7a861feb3f326.5dwj7h6kof5eixb3.rcgu.o" "/home/user/dev/rustc_codegen_gcc/target/debug/deps/rustc_codegen_gcc-11b7a861feb3f326.5e9x7kwxqzmokw35.rcgu.o" "/home/user/dev/rustc_codegen_gcc/target/debug/deps/rustc_codegen_gcc-11b7a861feb3f326.5eir7rgo7v0ii2d.rcgu.o" "/home/user/dev/rustc_codegen_gcc/target/debug/deps/rustc_codegen_gcc-11b7a861feb3f326.5g1bs28qa3d1i7o9.rcgu.o" "/home/user/dev/rustc_codegen_gcc/target/debug/deps/rustc_codegen_gcc-11b7a861feb3f326.5grd7soya3c1ue5t.rcgu.o" "/home/user/dev/rustc_codegen_gcc/target/debug/deps/rustc_codegen_gcc-11b7a861feb3f326.5n0mbhowrgtusuf.rcgu.o" "/home/user/dev/rustc_codegen_gcc/target/debug/deps/rustc_codegen_gcc-11b7a861feb3f326.67los8815f8c9sx.rcgu.o" "/home/user/dev/rustc_codegen_gcc/target/debug/deps/rustc_codegen_gcc-11b7a861feb3f326.720papmmlwk5xmh.rcgu.o" "/home/user/dev/rustc_codegen_gcc/target/debug/deps/rustc_codegen_gcc-11b7a861feb3f326.7hduu91c82o6e3v.rcgu.o" "/home/user/dev/rustc_codegen_gcc/target/debug/deps/rustc_codegen_gcc-11b7a861feb3f326.8ru5cxtuuxvqyh4.rcgu.o" "/home/user/dev/rustc_codegen_gcc/target/debug/deps/rustc_codegen_gcc-11b7a861feb3f326.9q5ltgzx51s52qm.rcgu.o" "/home/user/dev/rustc_codegen_gcc/target/debug/deps/rustc_codegen_gcc-11b7a861feb3f326.9z8z4siah7sn0r5.rcgu.o" "/home/user/dev/rustc_codegen_gcc/target/debug/deps/rustc_codegen_gcc-11b7a861feb3f326.b4eg3wd2s0i1y2t.rcgu.o" "/home/user/dev/rustc_codegen_gcc/target/debug/deps/rustc_codegen_gcc-11b7a861feb3f326.bfxglpxxxlvtnpj.rcgu.o" "/home/user/dev/rustc_codegen_gcc/target/debug/deps/rustc_codegen_gcc-11b7a861feb3f326.c0jscoq5masys3b.rcgu.o" "/home/user/dev/rustc_codegen_gcc/target/debug/deps/rustc_codegen_gcc-11b7a861feb3f326.c3uqz2eqlc7vkwk.rcgu.o" "/home/user/dev/rustc_codegen_gcc/target/debug/deps/rustc_codegen_gcc-11b7a861feb3f326.ckiut3rv5f93iim.rcgu.o" "/home/user/dev/rustc_codegen_gcc/target/debug/deps/rustc_codegen_gcc-11b7a861feb3f326.clp2thh4ciabovo.rcgu.o" "/home/user/dev/rustc_codegen_gcc/target/debug/deps/rustc_codegen_gcc-11b7a861feb3f326.di1mi4x7d2ben7o.rcgu.o" "/home/user/dev/rustc_codegen_gcc/target/debug/deps/rustc_codegen_gcc-11b7a861feb3f326.dwrgjr2nwlag64a.rcgu.o" "/home/user/dev/rustc_codegen_gcc/target/debug/deps/rustc_codegen_gcc-11b7a861feb3f326.ebb3ugj8wlhh5i7.rcgu.o" "/home/user/dev/rustc_codegen_gcc/target/debug/deps/rustc_codegen_gcc-11b7a861feb3f326.hnlv7pjeh5gsbfh.rcgu.o" "/home/user/dev/rustc_codegen_gcc/target/debug/deps/rustc_codegen_gcc-11b7a861feb3f326.jp2u799wic30q66.rcgu.o" "/home/user/dev/rustc_codegen_gcc/target/debug/deps/rustc_codegen_gcc-11b7a861feb3f326.klk33fqb850cwjh.rcgu.o" "/home/user/dev/rustc_codegen_gcc/target/debug/deps/rustc_codegen_gcc-11b7a861feb3f326.mkujcb48halt4lg.rcgu.o" "/home/user/dev/rustc_codegen_gcc/target/debug/deps/rustc_codegen_gcc-11b7a861feb3f326.nc8gw1wijy2g8de.rcgu.o" "/home/user/dev/rustc_codegen_gcc/target/debug/deps/rustc_codegen_gcc-11b7a861feb3f326.o2357ntysamxq0n.rcgu.o" "/home/user/dev/rustc_codegen_gcc/target/debug/deps/rustc_codegen_gcc-11b7a861feb3f326.pxt9ff9xtni0q6m.rcgu.o" "/home/user/dev/rustc_codegen_gcc/target/debug/deps/rustc_codegen_gcc-11b7a861feb3f326.qdkv4zjtyizje1t.rcgu.o" "/home/user/dev/rustc_codegen_gcc/target/debug/deps/rustc_codegen_gcc-11b7a861feb3f326.qrh8v1l3mita4zk.rcgu.o" "/home/user/dev/rustc_codegen_gcc/target/debug/deps/rustc_codegen_gcc-11b7a861feb3f326.qyrx70saf95n49l.rcgu.o" "/home/user/dev/rustc_codegen_gcc/target/debug/deps/rustc_codegen_gcc-11b7a861feb3f326.ru8r0ou3pa7654n.rcgu.o" "/home/user/dev/rustc_codegen_gcc/target/debug/deps/rustc_codegen_gcc-11b7a861feb3f326.rv43bzzob28qf79.rcgu.o" "/home/user/dev/rustc_codegen_gcc/target/debug/deps/rustc_codegen_gcc-11b7a861feb3f326.squ4axac9if2c1b.rcgu.o" "/home/user/dev/rustc_codegen_gcc/target/debug/deps/rustc_codegen_gcc-11b7a861feb3f326.srhmkt048dp0wra.rcgu.o" "/home/user/dev/rustc_codegen_gcc/target/debug/deps/rustc_codegen_gcc-11b7a861feb3f326.ttbait1v2850dm.rcgu.o" "/home/user/dev/rustc_codegen_gcc/target/debug/deps/rustc_codegen_gcc-11b7a861feb3f326.txcuwcnponi2k4o.rcgu.o" "/home/user/dev/rustc_codegen_gcc/target/debug/deps/rustc_codegen_gcc-11b7a861feb3f326.uokgqwlath2grnl.rcgu.o" "/home/user/dev/rustc_codegen_gcc/target/debug/deps/rustc_codegen_gcc-11b7a861feb3f326.vl519qsajtvyq6c.rcgu.o" "/home/user/dev/rustc_codegen_gcc/target/debug/deps/rustc_codegen_gcc-11b7a861feb3f326.vyaeskud2wxnwph.rcgu.o" "/home/user/dev/rustc_codegen_gcc/target/debug/deps/rustc_codegen_gcc-11b7a861feb3f326.x4an3xfv9oj3bi1.rcgu.o" "/home/user/dev/rustc_codegen_gcc/target/debug/deps/rustc_codegen_gcc-11b7a861feb3f326.x7hw7u8wtblwbmw.rcgu.o" "/home/user/dev/rustc_codegen_gcc/target/debug/deps/rustc_codegen_gcc-11b7a861feb3f326.xs0v66lpn9hxxja.rcgu.o" "/home/user/dev/rustc_codegen_gcc/target/debug/deps/rustc_codegen_gcc-11b7a861feb3f326.ygzih56ku5ohwak.rcgu.o" "/home/user/dev/rustc_codegen_gcc/target/debug/deps/rustc_codegen_gcc-11b7a861feb3f326.ylgz7km6o4xuho8.rcgu.o" "/home/user/dev/rustc_codegen_gcc/target/debug/deps/rustc_codegen_gcc-11b7a861feb3f326.ze3e7tyd4ynhueo.rcgu.o" "-Wl,--as-needed" "-L" "/home/user/dev/rustc_codegen_gcc/target/debug/deps" "-L" "/home/user/.rustup/toolchains/nightly-2021-08-12-x86_64-unknown-linux-gnu/lib/rustlib/x86_64-unknown-linux-gnu/lib" "-Wl,-Bstatic" "/home/user/dev/rustc_codegen_gcc/target/debug/deps/libar-8ec0ad9e88618f2c.rlib" "-L" "/home/user/.rustup/toolchains/nightly-2021-08-12-x86_64-unknown-linux-gnu/lib/rustlib/x86_64-unknown-linux-gnu/lib" "-Wl,-Bdynamic" "-ltest-8f0ddf5305823a79" "-Wl,-Bstatic" "/home/user/dev/rustc_codegen_gcc/target/debug/deps/libgccjit-662ae032528221e1.rlib" "/home/user/dev/rustc_codegen_gcc/target/debug/deps/libgccjit_sys-f1a4143f18332c66.rlib" "/home/user/dev/rustc_codegen_gcc/target/debug/deps/liblibc-0797a6c8ec8b4139.rlib" "-L" "/home/user/.rustup/toolchains/nightly-2021-08-12-x86_64-unknown-linux-gnu/lib/rustlib/x86_64-unknown-linux-gnu/lib" "-Wl,-Bdynamic" "-lrustc_driver-49ba2d0ecf6014f9" "-Wl,--start-group" "-L" "/home/user/.rustup/toolchains/nightly-2021-08-12-x86_64-unknown-linux-gnu/lib/rustlib/x86_64-unknown-linux-gnu/lib" "-lstd-5665011a98b2dd1d" "-Wl,--end-group" "-Wl,-Bstatic" "/home/user/.rustup/toolchains/nightly-2021-08-12-x86_64-unknown-linux-gnu/lib/rustlib/x86_64-unknown-linux-gnu/lib/libcompiler_builtins-96219fb718f2f3e8.rlib" "-Wl,-Bdynamic" "-lgccjit" "-lc" "-lm" "-lLLVM-12-rust-1.56.0-nightly" "-lgcc_s" "-lutil" "-lrt" "-lpthread" "-lm" "-ldl" "-lc" "-Wl,--eh-frame-hdr" "-Wl,-znoexecstack" "-L" "/home/user/.rustup/toolchains/nightly-2021-08-12-x86_64-unknown-linux-gnu/lib/rustlib/x86_64-unknown-linux-gnu/lib" "-o" "/home/user/dev/rustc_codegen_gcc/target/debug/deps/rustc_codegen_gcc-11b7a861feb3f326" "-Wl,--gc-sections" "-pie" "-Wl,-zrelro" "-Wl,-znow" "-nodefaultlibs"
  = note: /usr/bin/ld: cannot find -lgccjit
```

My gcc_path is `/home/user/dev/gccjit/build/gcc` (no whitespace). It seems like the -L flag is somehow not passed to the linker. It kinda makes sense since `cargo test` doesn't know anything about variables set in `build.sh`. I presume the CI environment is set independently.

Running `./cargo.sh test` panics at

https://github.com/antoyo/rustc_codegen_gcc/blob/e228f0c16ea8c34794a6285bf57aab627c26b147/src/lib.rs#L237-L240

Makes sense. Now let's add `-Clto=off` to RUSTFLAGS:

```
libgccjit.so: error: gcc_jit_block_add_assignment: rvalue unsigned int_unsigned int_unsigned int_unsigned int_unsigned int_unsigned int_unsigned int_unsigned int_unsigned int_unsigned int_unsigned int_unsigned int_unsigned int_unsigned int_unsigned int_unsigned int (type: struct struct) has scope limited to function __gccGlobalInitcrc32fast_1becbe98_cgu_1 but was used within function _ZN4core9core_arch3x864sse219_mm_srli_si128_impl17h4a020dcaf28a33dcE (in statement: *(bitcast(&stack_var_1, signed char  __attribute__((vector_size(sizeof (signed char) * 16)))  __attribute__((aligned(16))) *)) = _mm_shuffle_epi8 (loadedValue352, loadedValue350, unsigned int_unsigned int_unsigned int_unsigned int_unsigned int_unsigned int_unsigned int_unsigned int_unsigned int_unsigned int_unsigned int_unsigned int_unsigned int_unsigned int_unsigned int_unsigned int);)
thread 'rustc' panicked at 'gcc_jit_block_add_assignment: rvalue unsigned int_unsigned int_unsigned int_unsigned int_unsigned int_unsigned int_unsigned int_unsigned int_unsigned int_unsigned int_unsigned int_unsigned int_unsigned int_unsigned int_unsigned int_unsigned int (type: struct struct) has scope limited to function __gccGlobalInitcrc32fast_1becbe98_cgu_1 but was used within function _ZN4core9core_arch3x864sse219_mm_srli_si128_impl17h4a020dcaf28a33dcE (in statement: *(bitcast(&stack_var_1, signed char  __attribute__((vector_size(sizeof (signed char) * 16)))  __attribute__((aligned(16))) *)) = _mm_shuffle_epi8 (loadedValue352, loadedValue350, unsigned int_unsigned int_unsigned int_unsigned int_unsigned int_unsigned int_unsigned int_unsigned int_unsigned int_unsigned int_unsigned int_unsigned int_unsigned int_unsigned int_unsigned int_unsigned int);)', /home/user/.cargo/git/checkouts/gccjit.rs-13c2e290f2fb9e4d/0572117/src/block.rs:144:13
```

Here we go again.  At least now I know it wasn't my change that broke it.

Ugh. I'll figure it out tomorrow.

Commeownist:
> Also, it seems there are 3 new failing UI tests in the CI for this branch. There are 104 failure in this PR while yours have 107.
Is this expected?

Not at all. I'll see tomorrow.

antoyo:
Yeah, you need to set `LIBRARY_PATH` (and possibly `LD_LIBRARY_PATH`) manually for `cargo test` to work.

antoyo:
I added support for register variables in [this commit](https://github.com/antoyo/gcc/commit/18d948aa4de76bac8c9650c144bd286ac0ac2ce4). ([here](https://github.com/antoyo/gccjit.rs/commit/54be27e41fff7b6ab532e2e21a82df50a12b9ad3) for the rust binding)

Commeownist:
I've figured why I couldn't run the tests and I added support for explicit registers. I'm on mobile now, will report in an hour.

Commeownist:
(Got distracted again, sorry).

OK, now I can finally confirm that `asm!` blocks on x86-64 work just as intended. We did it! I have also been able to figure out how to run tests from `tests/run/*.rs`. The test coverage is rudimentary for now, but I think we can adopt some asm tests from rustc later.

In order to run the tests you need to build the `tests/lib.rs` executable with llvm backend. It's simply a test runner, and it can't be built with GCC just yet (the  `__gccGlobalInit` error from above, I've also seen some sse-related issues). 
  * `export LD_LIBRARY_PATH="$(cat gcc_path)"` - needed to start the runner since it has a dynamic dependency on `libgccjit.so`, _even though it doesn't use it_. 
  * `export RUSTFLAGS= "-L $(cat gcc_path)"` - required by the cc linker, I presume for the same reason.
  * `cargo test` - this will build and start the runner, that will in turn build and run the tests.  There are lots of warnings when using `--nocapture`, tests need to be cleaned.

Moderate issues:
* Unlike LLVM, GCC doesn't have a cross-platform constraint for symbols like `fn` and `static`, the symbol's name must be embedded into the template (in fact, [the spec](https://doc.rust-lang.org/beta/unstable-book/library-features/asm.html#operand-type) mandates it). Even though it's relatively easy with `fn`- just push the mangled name of the `fn`, it may not be so easy with `statics` because they may need some special treatment (`@plt`) in exceptional cases (thread locals and I presume weak symbols in shared libraries).

  I've implemented the simple case (mangled name for both `fn` and `static`). Should be good enough for now.

Minor issues:

* `Context::dump_to_file` does not properly dump register variables. It prints them as normal variables (`type name;`), but I would expect to see `register type name("reg_name");`.
 
Unresolved questions:
* Aligning stack. If gcc makes sure the stack is aligned upon `asm` invocations, the docs don't mention it. If it doesn't, this might generate incorrect code in some circumstances (thought I'm not sure which exactly). 

Resolved questions:
* The `m` memory operands and its relation to the difference between `AT&T` and `Intel` dialects doesn't seem to concern us because [the spec](https://doc.rust-lang.org/nightly/unstable-book/library-features/asm.html#memory-address-operands) places the responsibility of using the right addressing syntax on the programmer, not the backend. Really good news.

antoyo:
I indeed did not implement the dump yet and those tests are indeed expected to be compiled using LLVM (sorry if it wasn't clear).

For the stack alignment, as far as I know, it is for SSE code to work. The LLVM doc mentions "The compiler should make conservative assumptions about what the asm might contain and should generate its usual stack alignment code in the prologue if the ‘alignstack’ keyword is present", but it's unclear to me how to do that in GCC.

I'll look at the rest later. Thanks for your work!

Commeownist:
> Also, it seems there are 3 new failing UI tests in the CI for this branch. There are 104 failure in [this PR](https://github.com/antoyo/rustc_codegen_gcc/pull/73/checks?check_run_id=3450850819) while yours have 107.
> Is this expected?

Those are `ui/llvm-asm/issue-14936.rs`, `ui/llvm-asm/llvm-asm-literal-escaping.rs`, and `ui/numbers-arithmetic/saturating-float-casts.rs`. 

The first two use `llvm_asm!` in  a nontrivial way, and since we're not going to support it, I think it's settled. I excluded them.

The last one uses `llvm_asm!` as a black box. The test fails because the `f32 -> u64` runtime conversion does not work as expected. It used to pass previously because the black box is a stub, and GCC is smart enough to optimize the conversion at compile-time, thus turning the test into a no-op. Now that you have a real black box (thanks to this PR), the test is forced to work at runtime and actually test something and detect the bug. So yeah, my PR helped to discover a bug in the backend, you're welcome :wink:

src/asm.rs
```diff
@@ -1,254 +1,544 @@
...
+                InlineAsmOperandRef::SymStatic { def_id } => {
+                    constants_len += self.tcx.symbol_name(Instance::mono(self.tcx, def_id)).name.len();
                 }
-                _ => {}
             }
         }
 
-        // All output operands must come before the input operands, hence the 2 loops.
-        for (idx, op) in operands.iter().enumerate() {
+        for (rust_idx, op) in rust_operands.iter().enumerate() {
```
antoyo, line 275: Since this function is big, I'd see value in adding a comment for each step. Here, that would be:
```rust
// 2. Register variables.
```
before this for loop.

Please do the same for the other steps.

src/asm.rs
```diff
@@ -1,254 +1,544 @@
...
+                            constraint, 
+                            rust_idx,
+                            late,
+                            readwrite: false,
+                            tmp_var, 
+                            out_place,
+                        });
+
+                        let out_gcc_idx = outputs.len() - 1;
+                        let constraint = Cow::Owned(out_gcc_idx.to_string());
```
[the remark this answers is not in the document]
antoyo: It seems like the constraint field is only ever assigned a scalar value (either a single char or an integer). Do you think it would be worth using another type than `Cow<str>` to avoid the allocation in this case.

antoyo (commented):
Nice work!

I'm half-way through reviewing this file, but I'm sending you those comments now.

By the way, the tests directory was only meant as a way to start testing when I couldn't run Rust tests suite. It's okay that you used it in this PR, but I don't think we should make it test the whole language. If anything is missing in the Rust tests suite, I believe it would be better to add it there.

src/asm.rs
```diff
@@ -1,254 +1,544 @@
...
+            // Fortunately, it's used as a simple black box to make sure that inputs are not optimized away.
+            // Let's just emulate it.
+            let block = self.llbb();
+            let extended_asm = block.add_extended_asm(None, "");
+            for input in inputs {
+                extended_asm.add_input_operand(None, "r", input);
+            }
+            extended_asm.add_clobber("memory");
+            extended_asm.set_volatile_flag(true);
+        } else {
```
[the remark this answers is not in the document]
Commeownist: I could do that with no problem at all, but I would like to point out that the code style of main rust monorepo differs from yours. Since you're aiming at merging this project into the monorepo, I think we could avoid some of extra churn if we were to accept their code style. 

`rustfmt` is a good idea. With my previous point in mind, I propose those of of the Rust repo.

(I don't care what code style we'll end up with, just saving effort).

src/asm.rs
```diff
@@ -1,254 +1,544 @@
...
+                InlineAsmOperandRef::In { reg, value } => {
+                    if let Ok(constraint) = reg_to_gcc(reg).map(Cow::Borrowed) {
+                        inputs.push(AsmInOperand { 
+                            constraint, 
+                            rust_idx, 
+                            val: value.immediate()
+                        });
+                    } else {
+                        // left for the next pass
+                        continue
```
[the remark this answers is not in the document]
Commeownist: This was left as a guard in case we ever need to add some code after the `if else`. In my experience, this actually helps to avoid silly mistakes.

src/asm.rs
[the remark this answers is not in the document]
antoyo: That's not a problem. My PR is doing what some other project subtrees is doing: adding a ignore on rustfmt for my own project.

I'll add the rustfmt config, then!

src/asm.rs
```diff
@@ -485,13 +485,21 @@ impl<'a, 'gcc, 'tcx> AsmBuilderMethods<'tcx> for Builder<'a, 'gcc, 'tcx> {
...
+        //     (especially with current `rustc_backend_ssa` API).
+        //  2. Not every output operand has an `out_place`, and it's required by `add_output_operand`.
+        //
+        // Instead, we generate a temporary output variable for each output operand, and then this loop,
+        // generates `out_place = tmp_var;` assignments if out_place exists, or `(void)tmp_var;` if it doesn't.
         for op in &outputs {
             if let Some(place) = op.out_place {
                 OperandValue::Immediate(op.tmp_var.to_rvalue()).store(self, place);                
+            } else {
+                block.add_eval(None, op.tmp_var.to_rvalue());
```
antoyo, line 501: `add_eval()` is sometimes necessary to make something known to the backend, for instance, to execute something that does not return a value (for instance, a call to `printf` where you don't care about the result).

Here, I don't know if such a thing happens. The only thing I could see is telling to the backend the register name, but even then, it might already work without calling `add_eval()` (I believe this is the case, but I might be wrong).

Am I missing something here?

Commeownist: Not really. I'm just getting familiar with the code. I saw this thing is used when calling `void` functions, and made a guess, a wrong one. Do you want me to undo?

antoyo: Yes please!

src/asm.rs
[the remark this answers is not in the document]
antoyo: The LLVM probably has the same issue, so that was probably already looked at!

antoyo (commented):
I'm done revewing this file. Do you have other things to do besides addressing the reviews?w

src/asm.rs
```diff
@@ -1,254 +1,556 @@
...
+        if !self.late {
+            res.push('&');
+        }
+
+        res.push_str(&self.constraint);
+        res
+    }
+}
+
+enum ConstraintOrRegister<'a> {
```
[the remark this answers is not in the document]
Commeownist: Probably not, I did it automatically. I don't think there's any difference, but if you think so...

src/asm.rs
```diff
@@ -1,254 +1,556 @@
...
+            extended_asm.add_clobber("cc");
+        }
+        if !options.contains(InlineAsmOptions::NOMEM) {
+            extended_asm.add_clobber("memory");
+        }
+        if !options.contains(InlineAsmOptions::PURE) {
+            extended_asm.set_volatile_flag(true);
+        }
+        if !options.contains(InlineAsmOptions::NOSTACK) {
+            // TODO(@Commeownist): figure out how to align stack
```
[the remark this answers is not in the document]
Commeownist: Almost definitely no, sorry. This attribute applies only to structs, enums and other kinds of types, not statements, that's my experience. What does it even mean? "Align the generated asm code on X bytes boundary in the text section"? We need to align stack (`rsp`), not code.

One idea would be to declare a variable with this attribute just before the asm block, but I'm not sure if that would work. I'll start a thread on GCC mailing list.

Since it only concerns SSE-related code and we don't support it yet at all, I propose to leave this as TODO for now.

Fun fact: there is indeed a list of attributes that's applicable to statements, but [the list](https://gcc.gnu.org/onlinedocs/gcc-11.2.0/gcc/Statement-Attributes.html#Statement-Attributes) is depressingly short and not helping.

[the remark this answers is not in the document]
antoyo: It's okay to leave it as a TODO: the user can always manually align the stack anyway.

Commeownist:
> Do you have other things to do besides addressing the reviews?

Nothing immediately addressable: 

- [ ] Is one `cc` clobber is sufficient on all architectures to inform the compiler that FLAGS, FP stack, etc will be clobbered; for example, llvm backend clobbers all `st` registers if one of them (or `x87`)  is clobbered "due to the way LLVM handles the FP stack in inline assembly". I'm not sure if GCC has quirks around this as well.  I'm pretty sure it's working good for now, but am not 100% sure. 
- [ ] Stack aligning. I'm clearly not qualified to answer this, and I'm going to ask on GCC mailing list.
- [ ]  `sym STATIC` operands. Current code works well for all kinds of stable statics, but will break with , for example,  

    ```rust
    #[thread_local] static FOO: u64 = 0;  // this attribute is unstable
    asm!("", sym FOO);
    ```

I think we should merge this in as is - it's in pretty good shape IMHO.

Commeownist:
I've now excluded the  `tests/ui/llvm-asm/` folder entirely, so the only extra failing test must now be `ui/numbers-arithmetic/saturating-float-casts.rs`. I think the PR is ready.

src/asm.rs
```diff
@@ -1,254 +1,558 @@
...
-        for (idx, op) in operands.iter().enumerate() {
+        // Clobbers collected from `out("explicit register") _` and `inout("expl_reg") var => _`
+        let mut clobbers = vec![];
+
+        // We're trying to preallocate space for the template
+        let mut constants_len = 0;
+
+        // There are rules we must adhere to if we want GCC to do the right thing:
+        // 
+        // * Every local variable the asm block uses as output must be declared *before*
```
antoyo, line 158:
```suggestion
        // * Every local variable the asm block used as output must be declared *before*
```

src/asm.rs
```diff
@@ -1,254 +1,558 @@
...
+                    // Rustc frontend guarantees that input and output types are "compatible",
+                    // so we can just use input var's type for the output variable.
+                    //
+                    // This decision is also backed by the fact that LLVM needs in and out 
+                    // values to be of *exactly the same type*, not just "compatible". 
+                    // I'm not sure if GCC is so picky too, but better safe than sorry.
+                    let ty = in_value.layout.gcc_type(self.cx, false);
+                    let tmp_var = self.current_func().new_local(None, ty, "output_register");
+
+                    // If the out_place is None (i.e `inout(reg) var` syntax was used), we translate
```
antoyo, line 240: Did you mean the following?

```suggestion
                    // If the out_place is None (i.e `inout(reg) _` syntax was used), we translate
```

src/asm.rs
```diff
@@ -1,254 +1,558 @@
...
-        for (idx, op) in operands.iter().enumerate() {
+        // 2. Register variables.
+        for (rust_idx, op) in rust_operands.iter().enumerate() {
             match *op {
-                InlineAsmOperandRef::In { .. } | InlineAsmOperandRef::InOut { .. } => {
-                    operand_numbers.insert(idx, current_number);
-                    current_number += 1;
-                },
-                _ => (),
+                // `out("explicit register") _`
```
antoyo, line 281: Did you mean the following?

```suggestion
                // `out("explicit register") var`
```

src/asm.rs
```diff
@@ -1,254 +1,558 @@
...
+                            constraint: "r".into(), 
+                            rust_idx, 
+                            val: reg_var.to_rvalue()
+                        });
+                    }
+
+                    // processed in the previous pass
+                }
+
+                // `inout("explicit register") in_var => out_var/_`
```
antoyo, line 327: Did you mean the following?

```suggestion
                // `inout("explicit register") in_var => out_var`
```

src/asm.rs
```diff
@@ -1,254 +1,558 @@
...
+                        });
+                    }
+
+                    // processed in the previous pass
+                }
+
+                InlineAsmOperandRef::Const { .. } 
+                | InlineAsmOperandRef::SymFn { .. } 
+                | InlineAsmOperandRef::SymStatic { .. } => {
+                    // processed on the previous pass
```
antoyo, line 366:
```suggestion
                    // processed in the previous pass
```

antoyo (approved):
Sorry for the nitpicking ;) .
We can merge after those are fixed.

src/asm.rs
Commeownist: I think the grammar is correct :)

Commeownist: I kinda tend to omit extra "that"s and some articles. Lot's of my friends, including native speakers (British) do the same, so I guess it's just one of my peculiarities I have developed. Or a local thing. Or something.

Rephrased.

Commeownist:
For your information: https://gcc.gnu.org/pipermail/gcc-help/2021-September/140679.html

I guess we need to bring it up with rustc devs and see what we can do. I don't think something like this would be a good solution:
```asm
push rsp
and rsp, 0xffffffffffffff00
/* Code */
pop rsp
```

Because this push would be useless in 99% cases.

antoyo:
Yeah, I looked at the function prelude code generation in gcc, and while I couldn't find the one for x86, it looks like the stack alignment code is not separated from it, so it would be impossible to only call it.

antoyo:
Thanks a lot for your amazing work!

PR #79, opened 2021-09-11
Update to nightly-2021-09-11

Commeownist:
There are quite a few changes in rustc I had to address, most notably `black_box` is now a backend-driven intrinsic.  No more hacks for `llvm_asm`!

I haven't yet looked at the codegen tests: no spare time, sorry. Hopefully, this evening.


src/asm.rs
```diff
@@ -107,26 +107,12 @@ enum ConstraintOrRegister {
...
-        else {
-            // TODO(@Commeownist): switch to `struct_span_err_with_code` 
-            // once we get this merged into rustc
-            self.sess().struct_span_err(span, "GCC backend does not support `llvm_asm!`")
-                .help("consider using the `asm!` macro instead")
-                .emit();
-        }
+    fn codegen_llvm_inline_asm(&mut self, _ia: &LlvmInlineAsmInner, _outputs: Vec<PlaceRef<'tcx, RValue<'gcc>>>, _inputs: Vec<RValue<'gcc>>, span: Span) -> bool {
+        // TODO(@Commeownist): switch to `struct_span_err_with_code` 
+        // once we get this merged into rustc
```
bjorn3, line 112: I don't think that is worth it given that `llvm_asm!` will be removed in the future.

src/intrinsic/mod.rs
```diff
@@ -307,6 +308,21 @@ impl<'a, 'gcc, 'tcx> IntrinsicCallMethods<'tcx> for Builder<'a, 'gcc, 'tcx> {
                     }
                 }
 
+                sym::black_box => {
+                    // TODO(@Commeownist): I suspect this can be implemented more efficiently via volatile reads and writes.
```
bjorn3, line 312: That is likely less efficient as an empty asm block won't emit any instructions, while volatile loads and stores do emit instructions as the volatile prevents them from being removed.

src/asm.rs
Commeownist: Do you want me to remove the comment? I don't understand, could you please explain?

bjorn3: > Do you want me to remove the comment?

Yes

src/intrinsic/mod.rs
Commeownist: Fair enough, but I think it depends on specific use case: those optimizations inhibited by the asm block may be well worth the extra load. Anyway, I don't think this matters much, and programmers can reinvent their own versions of `black_box` just fine. 

I wish we had black box embedded in GCC and LLVM.

antoyo (approved):
Thanks!

src/allocator.rs
```diff
@@ -77,6 +77,9 @@ pub(crate) unsafe fn codegen(tcx: TyCtxt<'_>, mods: &mut GccContext, kind: Alloc
         else {
             block.end_with_void_return(None);
         }
+
+        // TODO(@Commeownist): Check if we need to emit some extra debugging info in certain circumstances
```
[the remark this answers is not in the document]
Commeownist: You know what's funny? Codegen tests apparently need debuginfo to work :) Also, related to #81

Commeownist:
(There are not enough hours in the day)

I have good news, bad news, and bad news (it's midnight and I'm a little bit unimaginative at midnights; I'll catch up another day).

Good news: I've been able to verify that the `clobber_api` is indeed correctly translated to `out("reg")_` operands, transparently to the backend (I compiled a simple separate file). Also, turns out that not every register's name as given by rustc is a valid GCC name for the clobber list. I'm sending a separate PR for it.

Bad news: running codegen tests in requires being able to save intermediate artifacts to a file ([judging by this](https://github.com/rust-lang/rust/blob/2c7bc5e33c25e29058cbafefe680da8d5e9220e9/compiler/rustc_codegen_llvm/src/back/write.rs#L747-L752)). I haven't find a way to save GIMPLE to a separate file, [or anywhere but stderr](https://gcc.gnu.org/onlinedocs/jit/topics/contexts.html#c.GCC_JIT_BOOL_OPTION_DUMP_INITIAL_GIMPLE). (I checked your fork as well). Related to #81. 

Bad news: `clobber_api("C")` clobbers `k[1-7]` AVX-512 registers, which [the docs](https://doc.rust-lang.org/beta/unstable-book/library-features/asm.html#abi-clobbers) _of course_ do not mention. I'm starting to love open-source world, and it's very weird, kiss-kill kind of love. Unfortunately, GCC does not allow to clobber these registers unless the `-mavx512f` option is passed, and I see `gccjit::context` does not support testing for added options. Another path to libgccjit...

bjorn3:
> Bad news: clobber_api("C") clobbers k[1-7] AVX-512 registers

I think it should be changed to only do so when AVX-512 is actually enabled for the current function I think. It makes sense that GCC doesn't allow it when AVX-512 isn't enabled. I am slightly suprised that LLVM does allow it. cc @Amanieu.

> which the docs of course do not mention.

The actual code for this is https://github.com/rust-lang/rust/blob/3ed6c1d23fd40f4367259a531465e809eb00ec27/compiler/rustc_target/src/asm/mod.rs#L802 which does indeed have k1-k7. Guess the docs need to be updated.

Amanieu:
Yes, LLVM simply ignores any invalid clobbers. So for GCC you would just have to gate the clobbers based on what target features are enabled.

Amanieu:
Note that this only applies to clobbers, LLVM still requires outputs (even dummy ones) to be valid registers for the current target.

This is the code that handles this in the LLVM codegen backend: https://github.com/rust-lang/rust/blob/master/compiler/rustc_codegen_llvm/src/asm.rs#L135

Commeownist:
>  I see `gccjit::context` does not support testing for added options. Another patch to libgccjit...

I think this was wrong. We can just check features provided by the frontend as suggested by @Amanieu .

antoyo:
@Commeownist: Is it okay for you if I fix the remaining reviews in order to merge this PR? I'd like to merge it soon to update the PR in rustc to move things forward.

Commeownist:
@antoyo Cool.

src/allocator.rs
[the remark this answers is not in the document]
Commeownist: Actually, I propose to leave the comment. This kind of information may be pretty hard to find in future.

Commeownist:
@antoyo I've resolved reviews (except I'd like to leave the TODO comment) and updated to `nightly-2021-09-17`. I think you can just merge it as is.

I'll leave codegen tests for another day.

antoyo:
There's an error in the CI which seems unrelated: it's about the bootstrap.

Commeownist:
@antoyo Are you sure that `rust-toolchain` file is respected? These lines from the log make me somewhat concerned:

```text
downloading https://static.rust-lang.org/dist/2021-09-08/rustfmt-nightly-x86_64-unknown-linux-gnu.tar.xz
...
extracting /home/runner/work/rustc_codegen_gcc/rustc_codegen_gcc/rust/build/cache/2021-09-08/rustfmt-nightly-x86_64-unknown-linux-gnu.tar.xz
```

Commeownist:
I suspect it's relate to those wonderful `git checkout <hash>` commands all over `test.sh`.

antoyo:
Does this work locally? Perhaps the CI is having issue…

bjorn3:
> @antoyo Are you sure that `rust-toolchain` file is respected? These lines from the log make me somewhat concerned:
> 
> ```
> downloading https://static.rust-lang.org/dist/2021-09-08/rustfmt-nightly-x86_64-unknown-linux-gnu.tar.xz
> ...
> extracting /home/runner/work/rustc_codegen_gcc/rustc_codegen_gcc/rust/build/cache/2021-09-08/rustfmt-nightly-x86_64-unknown-linux-gnu.tar.xz
> ```

Rustc uses a nightly rustfmt instead of the version associated with the bootstrap compiler. I don't quite remember why though.

bjorn3:
Bootstrap failed due to a new warning about an unused field that doesn't exist on beta. You could add `deny-warnings = false` to the `[rust]` section of `config.toml`:

https://github.com/rust-lang/rustc_codegen_gcc/blob/8ec7976ced61c31dbcaeda3f52d7ecc197764f80/test.sh#L150

Commeownist:
@antoyo Ready

antoyo:
Thanks!

PR #82, opened 2021-09-15
Impove handling of registers in inline asm

Commeownist:
Also, we no longer translate `ax`, `bx`, `cx` etc. register to `a`, `b`, `c` constraints and just create register variables for them. It's easier this way and should work just as good..

### DIFF
--- a/src/asm.rs
+++ b/src/asm.rs
@@ -173,7 +173,9 @@ impl<'a, 'gcc, 'tcx> AsmBuilderMethods<'tcx> for Builder<'a, 'gcc, 'tcx> {
                             continue
                         },
                         (Register(reg_name), None) => {
-                            clobbers.push(reg_name);
+                            if !clobbers.contains(&reg_name) {
+                                clobbers.push(reg_name);
+                            }    
                             continue
                         }
                     };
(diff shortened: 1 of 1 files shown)


src/asm.rs
```diff
@@ -542,16 +544,20 @@ fn reg_to_gcc(reg: InlineAsmRegOrRegClass) -> ConstraintOrRegister {
...
-                // For registers like r11, we have to create a register variable: https://stackoverflow.com/a/31774784/389119
-                name => return ConstraintOrRegister::Register(name), 
+            match reg {
+                InlineAsmReg::X86(_) => {
+                    // TODO(antoyo): add support for vector register.
+                    //
+                    // // For explicit registers, we have to create a register variable: https://stackoverflow.com/a/31774784/389119
+                    return ConstraintOrRegister::Register(match reg.name() {
+                        // Some of registers' names does not map 1-1 from rust to gcc
+                        "st(0)" => "st",
```
bjorn3, line 554: This is a case of divergence of register names between LLVM and GCC. Should the LLVM name be used as canonical name in `reg.name()`, the GCC name or should `reg.name()` get an option to switch between the two? @Amanieu what do you think?

Amanieu: `reg.name()` is the name used by rustc (for `in`/`out` register operands). It should be converted to the appropriate register name for LLVM/GCC.

antoyo:
@Commeownist One thing I was wondering was whether register variables are as efficient as constraint code. Do you have any idea?

Also, can you confirm that all the register names except `st` are the same between GCC and LLVM?

Commeownist:
>  One thing I was wondering was whether register variables are as efficient as constraint code. Do you have any idea?

So was I. Not sure, but I suspect there's no difference. I'm not qualified to answer this question, and I propose to move this discussion to `gcc-dev` mailing list.

> Also, can you confirm that all the register names except st are the same between GCC and LLVM?

I didn't check every single x86_64 register, but I can confirm that all registers clobbered by `clobber_abi` are handled correctly.

antoyo:
There are more UI tests failing with this PR.
Is this expected?

Commeownist:
No. I thin it's because it's not rebased onto master.

src/asm.rs
```diff
@@ -173,7 +173,20 @@ impl<'a, 'gcc, 'tcx> AsmBuilderMethods<'tcx> for Builder<'a, 'gcc, 'tcx> {
                             continue
                         },
                         (Register(reg_name), None) => {
-                            clobbers.push(reg_name);
+                            // `clobber_abi` can add lots of clobbers that are not supported by the target,
+                            // such as AVS-512 registers, so we just ignore unsupported registers
```
antoyo, line 177:
```suggestion
                            // such as AVX-512 registers, so we just ignore unsupported registers
```

antoyo (approved):
Sorry, I missed that typo the first time.

antoyo:
Thanks!

PR #84, opened 2021-09-18
Add required rustup components to rust-toolchain file

pymongo:
rustc_codegen_gcc and rust-clippy use same rustup components:

<https://github.com/rust-lang/rust-clippy/blob/master/rust-toolchain#L3>

move this code `rustup component add rust-src rustc-dev llvm-tools-preview` from `prepare.sh` to `rust-toolchain` is convenient for CI build (e.g. CI doesn't need to run prepare.sh before cargo test)

rust-toolchain file components require rustup minimal version [1.23.0 - 2020-11-27](https://github.com/rust-lang/rustup/blob/master/CHANGELOG.md#user-content-changed-3)

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -77,6 +77,9 @@ jobs:
         git config --global user.name "User"
         ./prepare.sh
 
+    - name: modified rust-toolchain to satisfy rustup version 1.20.1 in actions-rs
+      run: awk -F'[ ="]+' '$1 == "channel" { print $2 }' rust-toolchain > tmp && mv tmp rust-toolchain
+
     # Compile is a separate step, as the actions-rs/cargo action supports error annotations
     - name: Compile
       uses: actions-rs/cargo@v1.0.3
--- a/prepare_build.sh
+++ b/prepare_build.sh
@@ -1,5 +1,4 @@
 #!/bin/bash --verbose
 set -e
 
-rustup component add rust-src rustc-dev llvm-tools-preview
 ./build_sysroot/prepare_sysroot_src.sh
--- a/rust-toolchain
+++ b/rust-toolchain
@@ -1,1 +1,3 @@
-nightly-2021-09-17
+[toolchain]
+channel = "nightly-2021-09-17"
+components = ["rust-src", "rustc-dev", "llvm-tools-preview"]


pymongo:
```
~/work/rustc_codegen_gcc/rustc_codegen_gcc/build_sysroot/sysroot_src/library/core/tests ~/work/rustc_codegen_gcc/rustc_codegen_gcc
[TEST] libcore
rm: cannot remove './target': No such file or directory
error: toolchain '[toolchain]' is not installed
Error: error: toolchain '[toolchain]' is not installed
rustc_codegen_gcc is build for  but the default rustc version is rustc 1.57.0-nightly (e4828d5b7 2021-09-16).
error: toolchain '[toolchain]' is not installed
Using .
Error: error: toolchain '[toolchain]' is not installed
Error: Process completed with exit code 1.
```

seems the CI rustup version is less than 1.23.0 and not recognize toml format rust-toolchain, I would fix this

pymongo:
## actions-rs CI rustup version is 1.20.1 not support components field on rust-toolchain file

<https://github.com/actions-rs/core/blob/4487ec12063a411d19993339db585c95b354ffb4/src/commands/rustup.ts#L11>

> const PROFILES_MIN_VERSION = '1.20.1';

rust-clippy repo just delete the rust-toolchain file to make clippy_dev.yml CI works

<https://github.com/rust-lang/rust-clippy/blob/6ca00075eea400cac0b003bf2b75f5f36e840884/.github/workflows/clippy_dev.yml#L28-L38>

```
    - name: remove toolchain file
      run: rm rust-toolchain

    - name: rust-toolchain
      uses: actions-rs/toolchain@v1.0.6
      with:
        toolchain: nightly
        target: x86_64-unknown-linux-gnu
        profile: minimal
        components: rustfmt
        default: true
```

antoyo:
The CI was probably failing because the `test.sh` script reads this file, for instance [here](https://github.com/rust-lang/rustc_codegen_gcc/blob/master/test.sh#L136).

Unless you find a simple way to make this work, I believe it will be simpler to keep this file as it currently is.

bjorn3:
You could copy https://github.com/bjorn3/rustc_codegen_cranelift/blob/0ddb937624265f167f66b034422252d00801cd29/scripts/cargo.sh#L7

PR #101, opened 2021-10-12
Fix FIXME in `Builder::and` and `Builder::or` impls

### LINKED ISSUE #80: Don't put results in variables when unneeded
There was an issue in libgccjit that is now fixed, but required that some results were put unnecessarily in variables.
Examples are in the methods [`and()` and `or()`](https://github.com/antoyo/rustc_codegen_gcc/blob/7c707e4b9535b70de8c6beac24b7bc05595b1932/src/builder.rs#L594-L606).

fisherdarling:
Since a bug in gccjit was resolved: https://gcc.gnu.org/bugzilla//show_bug.cgi?id=95498 the `Builder::and` and `Builder::or` impl can be fixed and made much cleaner.

I think I did this right? Is there a list somewhere with the proper casting rules? I was a able to run `./test.sh` and it made it through testing rust tests:
`test result: FAILED. 4352 passed; 86 failed; 50 ignored; 0 measured; 0 filtered out; finished in 93.79s`

closes #80

### DIFF
--- a/src/builder.rs
+++ b/src/builder.rs
@@ -1,4 +1,4 @@
-use std::{borrow::Cow, ops::BitAnd};
+use std::borrow::Cow;
 use std::cell::Cell;
 use std::convert::TryFrom;
 use std::ops::Deref;


fisherdarling:
Glad to hear that was right! I was a little confused if I had to do anything with int promotion (?) or something like that, but I guess not.

Also, is there a reason why `rustfmt` is not being used?

fisherdarling:
Thanks! I'll definitely pick up some other low-hanging fruit in the future.

PR #102, opened 2021-10-17
Remove unused dependency on object

bjorn3:


antoyo:
Sorry for not looking at this earlier.
I wanted to check what was using this crate before.
I couldn't find it, so do you happen to know what was using this crate?

bjorn3:
Did you maybe copy it from cg_clif? Cg_clif uses it for creating the metadata object and used to use it for reading metadata from dylibs (now moved into cg_ssa).

antoyo:
You mean the archive code that was removed in this [PR](https://github.com/rust-lang/rustc_codegen_gcc/pull/79/files)?

bjorn3:
No, the metadata loader and dylib metadata writing.

antoyo:
Oh well. I guess it's not that important to find anyway.
Thanks!

PR #103, opened 2021-10-20
Support 128-bit integers on platforms without native support

antoyo:


src/lib.rs
```diff
@@ -91,6 +91,9 @@ impl CodegenBackend for GccCodegenBackend {
             sess.warn("LTO is not supported. You may get a linker error.");
         }
 
+        // FIXME: this way of checking if 128-bit integers are supported will overwrite the global
+        // state of libgccjit regarding types, causing some issues for bitcast.
+        // While waiting for a better solution, try using fork to avoid this overwrite.
```
bjorn3, line 96: With fork do you mean the `fork()` libc function? That may cause a deadlock or corruption as rustc may be embedded in another program. If this program is multi-threaded there may be a lock that will never be unlocked in the new process. The program embedding rustc may also try to send data from within rustc to a part running after rustc is done. Also even the regular rustc executable uses at least two threads. The rustc driver spawns a new thread to run the entirety of rustc on. This is to increase the stack size and to ensure some thread local variables can't be accessed after compilation is done (accessing some of them after compilation is done could cause UB).

antoyo: Yeah, I know this is not safe, but for now, I don't think I can do much as there is an issue within libgccjit with using multiple contexts and it's not possible right now to check if 128-bit integers are supported (except using this hack).

antoyo: An alternative would be to run an external process to check if 128-bit integers are supported, but I'm not sure which process I could rely on.

antoyo: I was able to fix this issue [on the libgccjit](https://github.com/antoyo/gcc/commit/014202a910ee1f0e87f7bfa74d58148a708dccf7), so those hacks won't be necessary.

PR #115, opened 2022-01-24
Correctly import foreign statics

bjorn3:
This is a fix for the environment variable issue I reported in a PM.

Previously foreign statics would actually cause a local static to be defined and exported. This issue was found because std::env::vars() was found to return no env vars despite many being defined. This was caused by libstd importing environ as foreign static. The accidental definition of environ caused libstd to read a null pointer which was interpreted as there being no environment variables at all.

I also move rustup component installation to rust-toolchain This allows cargo check to function correctly without having to first
run prepare_build.sh. cg_clif has been using rust-toolchain too for a while now.

### DIFF
--- a/src/declare.rs
+++ b/src/declare.rs
@@ -22,7 +22,7 @@ impl<'gcc, 'tcx> CodegenCx<'gcc, 'tcx> {
             global
         }
         else {
-            self.declare_global(name, ty, is_tls, link_section)
+            self.declare_global(name, ty, GlobalKind::Exported, is_tls, link_section)
         }
     }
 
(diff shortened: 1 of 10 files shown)


src/declare.rs
```diff
@@ -47,8 +47,8 @@ impl<'gcc, 'tcx> CodegenCx<'gcc, 'tcx> {
         unsafe { std::mem::transmute(func) }
     }*/
 
-    pub fn declare_global(&self, name: &str, ty: Type<'gcc>, is_tls: bool, link_section: Option<Symbol>) -> LValue<'gcc> {
-        let global = self.context.new_global(None, GlobalKind::Exported, ty, name);
+    pub fn declare_global(&self, name: &str, ty: Type<'gcc>, is_import: bool, is_tls: bool, link_section: Option<Symbol>) -> LValue<'gcc> {
```
antoyo, line 50: I'd rather make this function take a `GlobalKind` directly. That looks cleaner to me than a `true/false` value that you don't know what it is at the call site.

antoyo (commented):
Thanks a lot!

src/declare.rs
bjorn3: DOne

antoyo:
This seems to break `STDOUT`, according to the CI.

bjorn3:
https://github.com/rust-lang/rustc_codegen_gcc/blob/e690fb12731eb5f86d4c88400a344f65607af385/tests/run/assign.rs#L54

This is trying to import a global named `STDOUT` and then call `fflush` on it. This is not a thing libc exports. There is an export called `stdout` though. Previously the code would pass a null pointer to `fflush`. Will fix the test code tomorrow.

bjorn3:
Fixed tests and squashed.

antoyo:
Thanks a lot!

PR #122, opened 2022-01-30
Rustup to rustc 1.60.0-nightly (a00e130da 2022-01-29)

bjorn3:
This PR must not be squashed when merging.


bjorn3:
I'm locally getting several errors about traits not being implemented despite there clearly being a derive for them. For example

```rust
error[E0308]: mismatched types
  --> iter/adapters/intersperse.rs:79:30
   |
79 |         .intersperse_with(|| NotClone { u: 2 })
   |                              ^^^^^^^^^^^^^^^^^
   |                              |
   |                              expected `&intersperse::test_intersperse_with::NotClone`, found struct `intersperse::test_intersperse_with::NotClone`
   |                              help: consider borrowing here: `&NotClone { u: 2 }`
```

antoyo:
Strange.
I'm not sure what is going on.

bjorn3:
Fixed it. Had to switch coretests to the 2021 edition.

src/type_.rs
```diff
@@ -125,6 +125,12 @@ impl<'gcc, 'tcx> BaseTypeMethods<'tcx> for CodegenCx<'gcc, 'tcx> {
         else if typ.dyncast_vector().is_some() {
             TypeKind::Vector
         }
+        else if typ.is_compatible_with(self.float_type) {
+            TypeKind::Float
+        }
+        else if typ.is_compatible_with(self.double_type) {
+            TypeKind::Double
+        }
```
antoyo, line 133: Please rebase or remove this if not needed anymore. Your code might actually be better than what I wrote because I just did a comparison instead of calling `is_compatible_with()` so feel free to keep yours!

antoyo (approved):
Great job!
Only one thing to fix.

src/type_.rs
bjorn3: Rebased and removed the comparison in favor of `is_compatible_with()`. The comparison does a pointer comparison. I don't know if libgccjit guarantees that there is only a single float and double type value.

antoyo: Well, yes, but a float aligned won't have the same pointer as normal float, so it's best to use `is_compatible_with()`.

antoyo:
Thanks a lot!

PR #127, opened 2022-02-04
README: Add compiler-rt cloning step

JBBgameich:
I hope this is correct, at least it makes it build again for me.

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -21,6 +21,8 @@ You can also use my [fork of gcc](https://github.com/antoyo/gcc) which already i
 ```bash
 $ git clone https://github.com/rust-lang/rustc_codegen_gcc.git
 $ cd rustc_codegen_gcc
+$ git clone https://github.com/llvm/llvm-project llvm --depth 1 --single-branch
+$ export RUST_COMPILER_RT_ROOT="$PWD/llvm/compiler-rt"
 $ ./prepare_build.sh # download and patch sysroot src
 $ ./build.sh --release
 ```


Readme.md
```diff
@@ -21,6 +21,8 @@ You can also use my [fork of gcc](https://github.com/antoyo/gcc) which already i
 ```bash
 $ git clone https://github.com/rust-lang/rustc_codegen_gcc.git
 $ cd rustc_codegen_gcc
+$ git clone https://github.com/llvm/llvm-project llvm --depth 1 --single-branch $
+$ export RUST_COMPILER_RT_ROOT="$PWD/llvm/compiler-rt" $
```
bjorn3, line 25: This shouldn't be needed. compiler-builtins which is built as part of the rust standard library should already provide all necessary intrinsics. What error do you get without?

bjorn3: Looks like compiler-builtins itself reads this env var if the "c" package feature is enabled, which it shouldn:t at least on x86. What arch do you use?

JBBgameich: I'm on x86_64. The error I'm getting is
```
error: failed to run custom build command for `compiler_builtins v0.1.67`

Caused by:
  process didn't exit successfully: `/home/jonahb/dev/rust/rustc_codegen_gcc/build_sysroot/target/debug/build/compiler_builtins-8535d9b204370609/build-script-build` (exit status: 101)
  --- stdout
  cargo:rerun-if-changed=build.rs
  cargo:compiler-rt=/home/jonahb/.cargo/registry/src/github.com-1ecc6299db9ec823/compiler_builtins-0.1.67/compiler-rt
  cargo:rustc-cfg=feature="unstable"
  cargo:rustc-cfg=feature="mem-unaligned"

  --- stderr
  thread 'main' panicked at 'RUST_COMPILER_RT_ROOT is not set', /home/jonahb/.cargo/registry/src/github.com-1ecc6299db9ec823/compiler_builtins-0.1.67/build.rs:491:21
  note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace

```

bjorn3: I'm on x86_64 too. I have never seen this error before. I have no clue why this would happen on x86_64.

antoyo: That's related to my changes to support non-native integers: some builtins were missing in `compiler-builtins`, so I had to use `compiler-rt` to get them.

antoyo, line 25: I believe the `$` at the end of the lines should be removed.

antoyo:
Thanks!

PR #130, opened 2022-02-25
Rustup to rustc 1.61.0-nightly (4b043faba 2022-02-24)

bjorn3:
Please don't squash this PR when merging.


bjorn3:
The only new test failure is ui/attributes/main-removed-2/main.rs which is a newly introduced tests that uses a proc macro.

antoyo:
Thanks a lot!

PR #131, opened 2022-02-25
Fix miscompilation when cg_ssa is using multiple builders at the same time

bjorn3:


src/builder.rs
```diff
@@ -116,8 +116,7 @@ impl<'a, 'gcc, 'tcx> Builder<'a, 'gcc, 'tcx> {
 
         // NOTE: since jumps were added and compare_exchange doesn't expect this, the current blocks in the
```
antoyo, line 117: Please update `blocks` to `block` in this comment and the others as there is only one block now.

src/builder.rs
```diff
@@ -911,11 +901,12 @@ impl<'a, 'gcc, 'tcx> BuilderMethods<'a, 'tcx> for Builder<'a, 'gcc, 'tcx> {
...
 
     fn ptrtoint(&mut self, value: RValue<'gcc>, dest_ty: Type<'gcc>) -> RValue<'gcc> {
-        self.cx.ptrtoint(self.block.expect("block"), value, dest_ty)
+        let usize_value = self.cx.const_bitcast(value, self.cx.type_isize());
+        self.intcast(usize_value, dest_ty, false)
     }
 
     fn inttoptr(&mut self, value: RValue<'gcc>, dest_ty: Type<'gcc>) -> RValue<'gcc> {
-        self.cx.inttoptr(self.block.expect("block"), value, dest_ty)
+        self.cx.const_bitcast(value, dest_ty)
```
antoyo, line 909: Perhaps we should add an assert to make sure `value.get_type() == self.isize_type`.

src/common.rs
```diff
@@ -202,11 +181,7 @@ impl<'gcc, 'tcx> ConstMethods<'tcx> for CodegenCx<'gcc, 'tcx> {
                 }
 
                 let value = self.const_uint_big(self.type_ix(bitsize), data);
-                if layout.value == Pointer {
-                    self.inttoptr(self.current_block.borrow().expect("block"), value, ty)
-                } else {
-                    self.const_bitcast(value, ty)
-                }
+                self.const_bitcast(value, ty)
```
antoyo, line 184: This does not look right. `bitsize` could potentially be `1`, so are you sure the type will be the same size?

antoyo (commented):
That's a great quality of life improvement! Thanks!

A few nits and perhaps an issue:

src/common.rs
bjorn3: Only for a bool can bitsize be 1. A bool is represented as self.bool_type which is also 1 bit.

antoyo: Oh, right, `type_ix()` will return `self.i8_type` which is 8 bits, like `self.bool_type`.
In any case, it seems like the size is always equal, since the tests pass, but I'd also like an assert for the size.

src/builder.rs
bjorn3: Looks like the type doesn't have to be isize for the inttoptr used by the atomic_load and atomic_store intrinsics, so I changed this to do an intcast first like ptrtoint.

src/common.rs
antoyo: `get_size()` indeed only works on integers. If needed, I can make it work for pointer types.

That's a difference between LLVM and gcc: LLVM has a type `i1` of "1 bit" for bool while GCC's bool is 8 bits.

bjorn3: `type_ix(1)` needs to be a `bool_type`. Seems like it is currently implemented as `i8_type`.

src/context.rs
```diff
@@ -31,8 +31,7 @@ pub struct CodegenCx<'gcc, 'tcx> {
     pub codegen_unit: &'tcx CodegenUnit<'tcx>,
     pub context: &'gcc Context<'gcc>,
 
-    // TODO(antoyo): First set it to a dummy block to avoid using Option?
-    pub current_block: RefCell<Option<Block<'gcc>>>,
+    // TODO(bjorn3): First set it to a dummy function to avoid using Option?
```
antoyo, line 34: Do you think that's feasible?

bjorn3: I have no clue. You know more about gccjit than me.

bjorn3:
Yes, it is ready. Got two more PR's queued up after this.

src/common.rs
antoyo: Ok, can you add a TODO here saying to add an assert for the sizes, please?

src/context.rs
antoyo: Oh, I meant on the rustc side. It looks like we can probably be in some context where we're not in a function, right?

I guess we can still create a dummy function, but I like the solution of this PR which don't require a dummy block.

bjorn3: It may be possible to entirely remove `current_func` from the context in favor of only having it on the builder. I think this would be the best solution if it isn't too much work.

src/common.rs
bjorn3: Done

antoyo (approved):

PR #132, opened 2022-02-26
Support -Cpanic=unwind without unwinding

bjorn3:
This is necessary for bootstrapping rustc as compiling rustc with -Cpanic=abort is rather hard if at all possible. This does not implement actual unwinding on panics.

### DIFF
--- a/src/intrinsic/mod.rs
+++ b/src/intrinsic/mod.rs
@@ -1086,7 +1086,9 @@ impl<'a, 'gcc, 'tcx> Builder<'a, 'gcc, 'tcx> {
 }
 
 fn try_intrinsic<'gcc, 'tcx>(bx: &mut Builder<'_, 'gcc, 'tcx>, try_func: RValue<'gcc>, data: RValue<'gcc>, _catch_func: RValue<'gcc>, dest: RValue<'gcc>) {
-    if bx.sess().panic_strategy() == PanicStrategy::Abort {
+    // NOTE: the `|| true` here is to use the panic=abort strategy with panic=unwind too
+    if bx.sess().panic_strategy() == PanicStrategy::Abort || true {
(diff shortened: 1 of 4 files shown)


src/intrinsic/mod.rs
```diff
@@ -1086,7 +1086,8 @@ impl<'a, 'gcc, 'tcx> Builder<'a, 'gcc, 'tcx> {
 }
 
 fn try_intrinsic<'gcc, 'tcx>(bx: &mut Builder<'_, 'gcc, 'tcx>, try_func: RValue<'gcc>, data: RValue<'gcc>, _catch_func: RValue<'gcc>, dest: RValue<'gcc>) {
-    if bx.sess().panic_strategy() == PanicStrategy::Abort {
+    if bx.sess().panic_strategy() == PanicStrategy::Abort || true {
+        // TODO(bjorn3): Properly implement unwinding and remove the `|| true` once this is done.
```
antoyo, line 1090: Is the `|| true` there to make it use this code when unwinding is enabled?
If so, can you specify this in the comment, please, as this is not clear to me?

antoyo (approved):

src/intrinsic/mod.rs
bjorn3: Correct. Will add a comment.

bjorn3: Done

antoyo:
Thanks a lot!

bjorn3:
This PR fixed 12 rust tests it seems.

antoyo:
> This PR fixed 12 rust tests it seems.

Yeah, I was wondering whether it was this PR or the bug for int vs bool.

bjorn3:
That is also a possibility.

PR #133, opened 2022-02-27
Don't export global allocs which are not statics

bjorn3:
They aren't be referenced outside of the current cgu anyway. This should make optimizations a bit more effective.

### DIFF
--- a/src/consts.rs
+++ b/src/consts.rs
@@ -170,11 +170,9 @@ impl<'gcc, 'tcx> CodegenCx<'gcc, 'tcx> {
             match kind {
                 Some(kind) if !self.tcx.sess.fewer_names() => {
                     let name = self.generate_local_symbol_name(kind);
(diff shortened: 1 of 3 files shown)


src/consts.rs
```diff
@@ -170,11 +170,9 @@ impl<'gcc, 'tcx> CodegenCx<'gcc, 'tcx> {
             match kind {
                 Some(kind) if !self.tcx.sess.fewer_names() => {
                     let name = self.generate_local_symbol_name(kind);
-                    // TODO(antoyo): check if it's okay that TLS is off here.
```
bjorn3, line 173: Only statics can be thread local. Allocations referenced by other allocations can't ever be thread local as a single pointer has to be used independent of the thread.

antoyo (approved):

antoyo:
Thanks!

PR #135, opened 2022-02-28
Add support for on_stack parameters

antoyo:


bjorn3:
I can confirm this fixed proc macro support.

PR #136, opened 2022-03-08
Stop removing the llvm-asm tests in test.sh since they don't exist anymore

### LINKED ISSUE #126: Stop removing the llvm-asm tests in test.sh since they don't exist anymore

light4:
Fixes #126


antoyo (approved):
Thanks!

PR #142, opened 2022-03-23
Fix the channel parameter passed to `build_sysroot.sh`

yvt:
`build_sysroot.sh` takes `--release`/(none), not `release`/`debug`.

https://github.com/rust-lang/rustc_codegen_gcc/blob/44c0204af3ac285474a779f777f0adc158a8344b/build_sysroot/build_sysroot.sh#L20-L26

(A certain architecture causes ICE when compiling without optimization due to [a bug](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=96233) in GCC, hence this change is necessary.)

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -36,4 +36,9 @@ rm -r target/out || true
 mkdir -p target/out/gccjit
 
 echo "[BUILD] sysroot"
-time ./build_sysroot/build_sysroot.sh $CHANNEL
+if [[ "$CHANNEL" == "release" ]]; then
+    time ./build_sysroot/build_sysroot.sh --release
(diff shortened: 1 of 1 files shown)


bjorn3:
Maybe codegen backend and sysroot optimization level should be decoupled? For cg_clif I personally use optimized codegen backend with unoptimized sysroot as this is a net win for the amount if time it takes to test changes. An optimized backend sames more time on sysroot compilation than optimizing the backend costs.

yvt:
That sounds like a good idea. I think I'll try adding a new flag to `build.sh`.

PR #147, opened 2022-03-26
Don't pass --target in cargo.sh

bjorn3:
This was a workaround for compiling proc macros resulting in an abi
incompatibility. By passing --target proc macros will be built by the
llvm backend. This is no longer necessary as the abi incompatibility has
since been fixed.

### DIFF
--- a/cargo.sh
+++ b/cargo.sh
@@ -20,4 +20,4 @@ fi
 cmd=$1
 shift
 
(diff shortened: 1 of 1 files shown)


bjorn3:
On further thought this will regress error messages when proc macros panic as proc macros will now be built with `-Cpanic=abort`. See https://github.com/bjorn3/rustc_codegen_cranelift/issues/1174. @antoyo I will let it up to you if you want to merge this before implementing unwinding or if you want to wait on this.

antoyo:
I'm ok with merging this without unwinding support.

PR #151, opened 2022-03-30
Add intrinsic translation for x86 arch

GuillaumeGomez:


src/intrinsic/x86.rs
```diff
@@ -0,0 +1,766 @@
+"llvm.x86.avx512.mask.pcmpeq.q.256" => "__builtin_ia32_pcmpeqq256_mask"
```
bjorn3, line 1: Could this list be sorted?

GuillaumeGomez: I was also thinking the same. I'm currently updating the script generating the instructions.

bjorn3 (approved):

src/intrinsic/llvm.rs
```diff
@@ -3,134 +3,16 @@ use gccjit::Function;
...
-            "llvm.x86.sha1msg1" => "__builtin_ia32_sha1msg1",
-            "llvm.x86.sha1msg2" => "__builtin_ia32_sha1msg2",
-            "llvm.x86.sha1nexte" => "__builtin_ia32_sha1nexte",
-            "llvm.x86.sha1rnds4" => "__builtin_ia32_sha1rnds4",
-            "llvm.x86.sha256msg1" => "__builtin_ia32_sha256msg1",
-            "llvm.x86.sha256msg2" => "__builtin_ia32_sha256msg2",
-            "llvm.x86.sha256rnds2" => "__builtin_ia32_sha256rnds2",
-
             "llvm.sqrt.v2f64" => "__builtin_ia32_sqrtpd",
             _ => unimplemented!("***** unsupported LLVM intrinsic {}", name),
```
antoyo, line 13: What do you think of putting the include here and moving the `unimplemented!()` in the generated file?

PR #153, opened 2022-03-30
Sort archs and make the generate predictable

GuillaumeGomez:
Sorry for the noise, I didn't pay attention to the order of the archs. It's now predictable so this shouldn't happen again.

This PR adds the "general" intrinsics (not bound to any arch).

Last part of https://github.com/rust-lang/rustc_codegen_gcc/pull/151.


src/intrinsic/archs.rs
```diff
@@ -2935,31 +1508,1475 @@ match name {
...
+    "llvm.objectsize.i32.p0i8" => "__builtin_object_size",
+    "llvm.objectsize.i64.p0i8" => "__builtin_object_size",
+    "llvm.objectsize.i8.p0i8" => "__builtin_object_size",
+    "llvm.objectsize.v16i8.p0i8" => "__builtin_object_size",
+    "llvm.objectsize.v2i64.p0i8" => "__builtin_object_size",
+    "llvm.objectsize.v4i32.p0i8" => "__builtin_object_size",
+    "llvm.objectsize.v8i16.p0i8" => "__builtin_object_size",
+    "llvm.stackrestore" => "__builtin_stack_restore",
+    "llvm.stacksave" => "__builtin_stack_save",
+    "llvm.trap" => "__builtin_trap",
```
bjorn3, line 2980: I don't think any of these are used by stdarch or any other part of the standard library.

bjorn3: It is only used inside cg_llvm as implementation for the `abort` builder function. cg_gcc directly calls the `abort` builtin function there.

antoyo: `llvm.debugtrap` is needed by the `breakpoint` intrinsic.

bjorn3: Wouldn't the implementation of said intrinsic directly call the right gcc builtin?

GuillaumeGomez:
It seems like these functions shouldn't be added then. Is it fine to still have the sorting predictable?

antoyo:
> It seems like these functions shouldn't be added then. Is it fine to still have the sorting predictable?

Indeed. We can add them later if needed.

Yes, please, go ahead for the sorting.

GuillaumeGomez:
Done!

PR #160, opened 2022-04-14
Don't assume /bin/bash is available on every system.

MikaelUrankar:


antoyo:
Thanks!
It seems the CI doesn't like `/usr/bin/env bash --verbose`. Can you please fix it?

PR #163, opened 2022-04-24
Add inline assembly `sym` operands as GCC input operands

### LINKED ISSUE #157: `fn`s referenced by `sym` operands get DCE-ed, causing linker errors
`tests/run/asm.rs` fails to compile after increasing the optimization level.

```diff
--- a/tests/lib.rs
+++ b/tests/lib.rs
@@ -39,6 +39,8 @@ fn main() {
                 "-Zno-parallel-llvm",
                 "-C", "panic=abort",
                 "-C", "link-arg=-lc",
+                "-C", "codegen-units=1",
+                "-C", "opt-level=3",
                 "-o", exe.to_str().expect("to_str"),
                 path.to_str().expect("to_str"),
             ]);
```

```
$ env LIBRARY_PATH=(cat gcc_path) cargo test asm
   ⋮
  |                                                                                                                                                                  
  = note: "cc" "-m64" ...
  = note: .../ld: .../rustc_codegen_gcc/build_sysroot/sysroot/lib/rustlib/x86_64-unknown-linux-gnu/lib/libstd.rlib(std.std.5ab0d951-cgu.5.rcgu.o): warning: relocation in read-only section `.rodata.global.1G'
          .../ld: .../.tmphnejVm/asm.asm.b19220b4-cgu.0.rcgu.o: in function `asm::main':
          fake.c:(.text._ZN3asm4main17h6b1c2af27aca2c4fE+0x114): undefined reference to `asm::main::foo'
          .../ld: fake.c:(.text._ZN3asm4main17h6b1c2af27aca2c4fE+0x128): undefined reference to `asm::main::foo'
          .../ld: warning: creating DT_TEXTREL in a PIE
```

yvt:
Updates `<Builder as AsmBuilderMethods>::codegen_inline_asm` to convert `sym` operands into GCC `"X" (&func_or_static)` input operands to indicate the dependency on the referenced symbols and prevent them from being eliminated.

Fixes #157.

### DIFF
--- a/tests/lang_tests_common.rs
+++ b/tests/lang_tests_common.rs
@@ -1,3 +1,4 @@
+//! The common code for `tests/lang_tests_*.rs`
 use std::{
     env::{self, current_dir},
     path::PathBuf,
@@ -7,7 +8,15 @@ use std::{
 use lang_tester::LangTester;
 use tempfile::TempDir;
 
-fn main() {
+/// Controls the compile options (e.g., optimization level) used to compile
+/// test code.
+#[allow(dead_code)] // Each test crate picks one variant
+pub enum Profile {
+    Debug,
+    Release,
+}
+
+pub fn main_inner(profile: Profile) {
     let tempdir = TempDir::new().expect("temp dir");
     let current_dir = current_dir().expect("current dir");
     let current_dir = current_dir.to_str().expect("current dir").to_string();
@@ -42,6 +51,15 @@ fn main() {
                 "-o", exe.to_str().expect("to_str"),
                 path.to_str().expect("to_str"),
             ]);
+            match profile {
+                Profile::Debug => {}
+                Profile::Release => {
+                    compiler.args(&[
+                        "-C", "opt-level=3",
+                        "-C", "lto=no",
+                    ]);
+                }
+            }
             // Test command 2: run `tempdir/x`.
             let runtime = Command::new(exe);
             vec![("Compiler", compiler), ("Run-time", runtime)]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -9,8 +9,12 @@ license = "MIT OR Apache-2.0"
 crate-type = ["dylib"]
 
 [[test]]
-name = "lang_tests"
-path = "tests/lib.rs"
+name = "lang_tests_debug"
+path = "tests/lang_tests_debug.rs"
+harness = false
+[[test]]
+name = "lang_tests_release"
+path = "tests/lang_tests_release.rs"
 harness = false
 
 [features]
--- a/src/asm.rs
+++ b/src/asm.rs
@@ -13,6 +13,7 @@ use std::borrow::Cow;
 use crate::builder::Builder;
 use crate::context::CodegenCx;
 use crate::type_of::LayoutGccExt;
+use crate::callee::get_fn;
 
 
 // Rust asm! and GCC Extended Asm semantics differ substantially.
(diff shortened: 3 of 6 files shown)


antoyo:
I think it would be better to do it the same way as the LLVM codegen, that is by setting the attribute `__attribute__((__used__))`.

Is this a blocker for you?

yvt:
The LLVM codegen does `add_compiler_used_global` only for `sym` operands in `global_asm!`, not those in `asm!`, unless I'm missing something.

https://github.com/rust-lang/rust/blob/d8e59edbfa47ff38e23e6dedab6bedd3b41895e0/compiler/rustc_codegen_llvm/src/asm.rs#L345-L347

antoyo:
Oh, you're right indeed!

Have you tried using the same constraint (`s`) as the LLVM codegen to see if it fixes this issue?

antoyo:
Also, could you please add an option (maybe an environment variable) in the tests to allow running the tests with optimizations in the CI?

yvt:
> Have you tried using the same constraint (`s`) as the LLVM codegen to see if it fixes this issue?

GCC seems to employ some sort of check that rejects this when position independent code is enabled. <https://cpp.godbolt.org/z/jdGbxKzcq>

yvt:
> Also, could you please add an option (maybe an environment variable) in the tests to allow running the tests with optimizations in the CI?

Since `lang_tester` doesn't provide a good support for test matrices, I split `tests/lib.rs` into `tests/lang_tests_{debug, release}.rs` so that  `cargo test` would test both with delimited outputs. Does this look good for you? <https://github.com/rust-lang/rustc_codegen_gcc/pull/163/commits/c5b80404d299160d4b941182f0c2eea578a1610e>

```
$ cargo test
    ⋮
     Running tests/lang_tests_debug.rs (target/debug/deps/lang_tests_debug-9d19ab0362e92e44) 

running 21 tests                                                                                                                                            
test lang_tests::slice.rs ... ok
test lang_tests::static.rs ... ok
    ⋮

test result: ok. 21 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out

     Running tests/lang_tests_release.rs (target/debug/deps/lang_tests_release-bff42339b6b0d08f)

running 21 tests
test lang_tests::fun_ptr.rs ... ok
test lang_tests::static.rs ... ok
    ⋮

test result: ok. 21 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out
```

tests/lang_tests_common.rs
```diff
@@ -16,15 +16,21 @@ fn main() {
     env::set_var("LD_LIBRARY_PATH", gcc_path);
     LangTester::new()
         .test_dir("tests/run")
-        .test_file_filter(|path| path.extension().expect("extension").to_str().expect("to_str") == "rs")
+        .test_file_filter(|path| {
```
antoyo, line 19: I'd appreciate if you remove unrelevant changes (formatting in this case).

I tried creating a config for rustfmt that would look like what I wanted, but I was unable to do so, so I gave up, but if that would be important to you and others, I could create issues in rustfmt's repo to request adding the options I'm missing.

tests/lang_tests_common.rs
```diff
@@ -7,7 +7,7 @@ use std::{
 use lang_tester::LangTester;
 use tempfile::TempDir;
 
-fn main() {
+pub fn main_inner(modify_command: fn(&mut Command) -> &mut Command) {
```
antoyo, line 10: I'd prefer to use a config argument instead of a function. That argument can be a simple enum with `Debug/Release` variants for now. We could convert that later to a struct if needed.

antoyo (commented):
Yes, I like having debug and release mode separate.

A few nits.

tests/lang_tests_common.rs
yvt: Have you considered adding a `.rustfmt.toml` file with `ignore = ...` or `disable_all_formatting = true`?

A number of people (including me) have auto-format-on-save enabled, and there's at least [one editor plugin](https://github.com/mitranim/sublime-rust-fmt/blob/86b3498adee26a278908df00f307e601e253dbf0/RustFmt.sublime-settings#L32) that enables this by default. Adding this would save churns for everyone and for you.

antoyo: Good idea, I did it [here](https://github.com/rust-lang/rustc_codegen_gcc/commit/dc8da94d56435f95547ac02cf634dea223668a78).

yvt:
Addressed your feedback in 63ffdfdd1776afa2e82bbd3d2ff8ff7b7f0d5b67.

antoyo (approved):
Thanks a lot!

PR #164, opened 2022-04-24
Don't emit `.intel_syntax` for non-x86 targets

yvt:
Cherry-picked from <https://github.com/yvt/rustc_codegen_gcc/commit/5f8660156af3721f79b96d084c0e28bf421ea92f>


src/asm.rs
```diff
@@ -436,7 +435,7 @@ impl<'a, 'gcc, 'tcx> AsmBuilderMethods<'tcx> for Builder<'a, 'gcc, 'tcx> {
             }
         }
 
-        if !intel_dialect {
+        if att_dialect {
             template_str.push_str(INTEL_SYNTAX_INS);
```
antoyo, line 439: Won't this emit `.intel_syntax` on non-x86 targets?

antoyo (commented):
Thanks!

src/asm.rs
yvt:
```rust
let att_dialect = is_x86 && options.contains(InlineAsmOptions::ATT_SYNTAX);
//                ^^^^^^^^^
```
;)

PR #165, opened 2022-04-24
Define immutable statics with const qualified types

### LINKED ISSUE #161: Immutable static data is placed in `.data` section
[Complete source code](https://gist.github.com/yvt/565b8c8b33744be6c2ee0e3b7b1d63a7)

```rust
#[no_mangle]
pub extern "C" fn ultimate_answer() -> *const u32 {
    (&[42u32; 42 * 42 * 42] as &'static [u32]).as_ptr()
}
```

In LLVM codegen, the above constant `[42u32; 42 * 42 * 42]` is placed in `.text` section, whereas in GCC codegen, it's placed in `.data` section.

```
$ cargo build --release
$ size target/release/librcg_const.so
   text    data     bss     dec     hex filename
 297166     480       8  297654   48ab6 target/release/librcg_const.so
```

```
$ rustc_codegen_gcc/cargo.sh build --release
$ size target/release/librcg_const.so
   text    data     bss     dec     hex filename
    958  296856       8  297822   48b5e target/release/librcg_const.so
```

In another case, a constant value compared in `assert_eq!` is placed in `.data` section when it would be much more efficient to inline the constant as an immediate operand.

```rust
// r3_kernel::task::wait_until_woken_up
assert_eq!(*running_task.st.read(&*lock), TaskSt::Running);
```

```
fff81a81:   fb 22 44 03 00 00   mov.l   #836, r2       // &TaskSt::Running
fff81a87:   b2 c3               movu.b  10[r4], r3     // task.st
fff81a89:   58 2e               movu.b  [r2], r14      // *&TaskSt::Running
      ⋮
fff81a8e:   47 e3               cmp r14, r3
fff81a90:   20 18               beq.b   fff81aa8       // (branch if assertion passed)
      ⋮
```

yvt:
This PR adds `const` qualification to the types of immutable statics to allow them to be placed in a read-only section.

Fixes #161.

```rust
#[no_mangle]
pub extern "C" fn ultimate_answer_concise() -> u32 {
    const CANDIDATES: &[u32] = &[41, 42, 43];
    CANDIDATES[1]
}
```

```
Before:
    0000000000001100 <ultimate_answer>:
        1100:       48 8d 05 19 2f 00 00    lea    0x2f19(%rip),%rax        # 4020 <ultimate_answer_concise+0x2f10>
        1107:       c3                      ret
        1108:       0f 1f 84 00 00 00 00    nopl   0x0(%rax,%rax,1)
        110f:       00

After:
    0000000000001110 <ultimate_answer_concise>:
        1110:       b8 2a 00 00 00          mov    $0x2a,%eax
        1115:       c3
```

### DIFF
--- a/src/consts.rs
+++ b/src/consts.rs
@@ -47,7 +47,10 @@ impl<'gcc, 'tcx> StaticMethods for CodegenCx<'gcc, 'tcx> {
             }
         }
         let global_value = self.static_addr_of_mut(cv, align, kind);
-        // TODO(antoyo): set global constant.
+        #[cfg(feature = "master")]
+        self.global_lvalues.borrow().get(&global_value)
+            .expect("`static_addr_of_mut` did not add the global to `self.global_lvalues`")
+            .global_set_readonly();
         self.const_globals.borrow_mut().insert(cv, global_value);
         global_value
     }
@@ -88,7 +91,8 @@ impl<'gcc, 'tcx> StaticMethods for CodegenCx<'gcc, 'tcx> {
         // mutability are placed into read-only memory.
         if !is_mutable {
             if self.type_is_freeze(ty) {
-                // TODO(antoyo): set global constant.
+                #[cfg(feature = "master")]
+                global.global_set_readonly();
             }
         }
 
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -41,7 +41,7 @@ dependencies = [
 [[package]]
 name = "gccjit"
 version = "1.0.0"
-source = "git+https://github.com/antoyo/gccjit.rs#6c2af0cf733a26740f01a7c679afc20431165a54"
+source = "git+https://github.com/antoyo/gccjit.rs#bdb86fb5092895ff5589726b33250010c64d93f6"
 dependencies = [
  "gccjit_sys",
 ]
@@ -49,7 +49,7 @@ dependencies = [
 [[package]]
 name = "gccjit_sys"
 version = "0.0.1"
-source = "git+https://github.com/antoyo/gccjit.rs#6c2af0cf733a26740f01a7c679afc20431165a54"
+source = "git+https://github.com/antoyo/gccjit.rs#bdb86fb5092895ff5589726b33250010c64d93f6"
 dependencies = [
  "libc 0.1.12",
 ]


bjorn3:
What is the size of an optimized hello world before and after this PR?

yvt:
@bjorn3 

```
$ cargo new --bin hello
$ cd hello

---- cg_llvm ----
$ cargo +nightly-2022-03-26 build --release
$ size target/release/hello
   text    data     bss     dec     hex filename
 296079   11984     536  308599   4b577 target/release/hello

---- cg_gcc master ----
$ rustc_codegen_gcc/cargo.sh build --release
$ size target/release/hello
   text    data     bss     dec     hex filename
 358162   34944     112  393218   60002 target/release/hello

---- cg_gcc #165 ----
$ rustc_codegen_gcc/cargo.sh build --release
$ size target/release/hello
   text    data     bss     dec     hex filename
 389882    1880       8  391770   5fa5a target/release/hello
```

antoyo:
Perhaps comparing the sizes with only one Rust codegen unit would give a better comparison.

yvt:
With `-Ccodegen-units=1` for app:

```
$ echo 'codegen-units = 1' >> Cargo.toml

---- cg_gcc master ----
   text    data     bss     dec     hex filename
 358018   34944     112  393074   5ff72 target/release/hello

---- cg_gcc #165 ----
   text    data     bss     dec     hex filename
 389738    1880       8  391626   5f9ca target/release/hello

---- cg_llvm ----
   text    data     bss     dec     hex filename
 291793   11792     536  304121   4a3f9 target/release/hello
```

With `-Ccodegen-units=1` for app and sysroot:

```
$ echo 'codegen-units = 1' >> Cargo.toml
$ echo 'codegen-units = 1' >> rustc_codegen_gcc/build_sysroot/Cargo.toml

---- cg_gcc master ----
   text    data     bss     dec     hex filename
 311747   28808      88  340643   532a3 target/release/hello

---- cg_gcc #165 ----
   text    data     bss     dec     hex filename
 337443    1848       8  339299   52d63 target/release/hello
```

src/builder.rs
```diff
@@ -661,7 +661,7 @@ impl<'a, 'gcc, 'tcx> BuilderMethods<'a, 'tcx> for Builder<'a, 'gcc, 'tcx> {
         // dereference after a drop, for instance.
         // TODO(antoyo): handle align of the load instruction.
         let deref = ptr.dereference(None).to_rvalue();
-        let value_type = deref.get_type();
+        let value_type = deref.get_type().unqualified();
```
antoyo, line 664: Why was this added?

antoyo (commented):
LGTM.
Thanks!

src/builder.rs
yvt: This strips `const` qualification to make the defined local variable writable. Writing to a `const` local variable causes a libgccjit error (which is presumably due to [`memento_of_get_const::accepts_writes_from`](https://github.com/antoyo/gcc/blob/141e034e736dd375a1236d5bc0a5b9ba0c6a274c/gcc/jit/jit-recording.h#L757-L761) returning `false`):

```
[BUILD] mini_core
   ⋮
libgccjit.so: error: gcc_jit_block_add_assignment: mismatching types: assignment to loadedValue50 (type: const __uint8_t) from *&_RNvNvCs2VSKAzIc0pl_9mini_core7get_tls1A (type: const __uint8_t)
```

antoyo: Hum, that could possibly remove other important qualifiers (like volatile, aligned).
I'll write a patch to libgccjit to have the possibility to only remove const.

yvt: Actually, that might not be necessary: `volatile` never appears in this place. The alignment is explicitly specified by the caller (`align: Align`), so we just have to apply it by `get_aligned`.

yvt: Never mind, `value_type` should have the natural alignment of the pointed type.

antoyo: With your edits, I'm not sure what's needed now.

Do you wait for my patch to only remove const?

yvt: Oh, sorry for leaving you hanging. Yes, I'm awaiting your patch.

antoyo: I added it [here](https://github.com/antoyo/gcc/commit/72cb89f82a6c9c5471f0b2c122b21dbe387c7904) and [there](https://github.com/antoyo/gccjit.rs/commit/383523c29c9dde572d2079ad21decb572da24cc9).

antoyo:
The call to `is_const()` will require to be feature-gated for libgccjit12.
I'll add it to `libgccjit_without_int128.so`.

yvt:
I realized that a `PlaceRef` having a `const` qualifier in its pointee type may be created in violation of the following invariant. I'd like to poke around a bit more.

<https://github.com/rust-lang/rustc_codegen_gcc/blob/0405aa0065f567ac11756c8ce5845a5bc9604f94/src/type_of.rs#L160-L161>

> The call to `is_const()` will require to be feature-gated for libgccjit12.

Do you want me to disable this PR's behavior entirely for libgccjit12, or ...?

antoyo:
> > The call to `is_const()` will require to be feature-gated for libgccjit12.
> 
> Do you want me to disable this PR's behavior entirely for libgccjit12, or ...?

Just adding `#[cfg(feature="master")]` before the line using `is_const()` should do it.
To avoid the warning of unused `mut`, it might be better to create a new variable with the same name instead.

yvt:
The const qualifier is now stripped as soon as a static's address is taken. This prevents const-qualified pointers, which shouldn't appear otherwise, from circulating and may fix bugs that we are not aware of.

src/mono_item.rs
```diff
@@ -16,6 +16,18 @@ impl<'gcc, 'tcx> PreDefineMethods<'tcx> for CodegenCx<'gcc, 'tcx> {
...
+
+        // As an optimization, all shared statics which do not have interior
+        // mutability are placed into read-only memory.
+        let gcc_type =
+            if !is_mutable && self.type_is_freeze(ty) {
+                gcc_type.make_const()
+            }
+            else {
+                gcc_type
+            };
```
antoyo, line 29: The libgccjit12 CI failed and I have the feeling it could be due to this change.
Can you please try to feature-gate this as well?

yvt: Done in 42dee9393c7f945c6805140f879f56a6a07b6149

src/consts.rs
```diff
@@ -299,6 +288,33 @@ impl<'gcc, 'tcx> CodegenCx<'gcc, 'tcx> {
...
+/// returned `RValue` has a canonical type.
+fn canonicalize_static_addr<'gcc, 'tcx>(cx: &CodegenCx<'gcc, 'tcx>, addr: RValue<'gcc>) -> RValue<'gcc> {
+    // Remove a const qualifier if it has one
+    #[cfg(feature = "master")]
+    let addr = {
+        let pointee_ty = addr.get_type().get_pointee().expect("get_pointee");
+        if let Some(stripped_pointee_ty) = pointee_ty.is_const() {
+            // Cast `addr` to a non-const pointer
+            let new_addr_ty = stripped_pointee_ty.make_pointer();
+            cx.context.new_bitcast(None, addr, new_addr_ty)
```
antoyo, line 311: Since it's a pointer, can this be a normal cast instead of a bitcast?

antoyo (commented):
Can you explain to me why you canonicalize the rvalue?
I don't understand why it's done.

yvt:
> Can you explain to me why you canonicalize the rvalue? I don't understand why it's done.

Basically, when taking a pointer of a static, we don't want to get different pointer types depending on whether the static is mutable (`T *`) or immutable (`T const *`). Otherwise we'll get libgccjit errors like <https://github.com/rust-lang/rustc_codegen_gcc/pull/165#discussion_r857233579> and <https://github.com/rust-lang/rustc_codegen_gcc/pull/165#discussion_r862778579>. Also that's a violation of an internal invariant (https://github.com/rust-lang/rustc_codegen_gcc/pull/165#issuecomment-1114267909).

yvt:
> Do you mean that was added as a workaround to libgccjit, to be able to load the address of a global variable?

To be more precise, it's to make the local variable to store the loaded data assignable. (A const-qualified local variable [cannot be assigned to](https://c.godbolt.org/z/aGM58vG3r).) I'd rather see this as a symptom of [inconsistent typing of rvalues](https://github.com/rust-lang/rustc_codegen_gcc/pull/165#issuecomment-1115720912), which this "canonicalization" fixes.

Meanwhile, #170 has changed the situation; the local variable created in `load` is now explicitly typed by the caller, and the source pointer is automatically bit-cast, which removes the possibility of this libgccjit error. In my quick testing, there seems to be no remaining ways by which `const`-qualified pointers can cause visible issues. This gives us another option: proceed without canonicalization. I recommend against doing so, but I guess it's your call.

> I worry about the effect of only having non-const pointer (could that change the code generation?).

Assuming that the semantics of libgccjit instructions is similar to that of C, the presence of const qualifier only matters for the mutability of variables (`T var` vs `T const var`). Const qualifiers in other positions can be cast in or away anytime and (afaik) mean nothing.

antoyo:
> This gives us another option: proceed without canonicalization. I recommend against doing so, but I guess it's your call.

Why would you recommend to still do canonicalization?
I try to keep the same structure as the LLVM codegen in order to keep up with the changes there and also in the hope that I can refactor common code into `rustc_codegen_ssa` some day. This PR changes the structure a bit with this canonicalization and `static_addr_of_inner`, so that's why I want to know your opinion on this.

> Const qualifiers in other positions can be cast in or away anytime and (afaik) mean nothing.

Good to know. I worried about TBAA doing something here, but it's disabled in Rust anyway.

yvt:
> Why would you recommend to still do canonicalization? I try to keep the same structure as the LLVM codegen in order to keep up with the changes there [...]

I'm starting to understand your design intent. But as you probably already know, there are differences between LLVM and libgccjit's APIs, and sometimes we have to make compromise and take an approach unique to cg_gcc. And when we do, the best thing we can do to mitigate the impact is to contain it to a narrow scope.

First, why did I add `static_addr_of_inner`? libgccjit doesn't have a counterpart of `LLVMSetGlobalConstant`, which means the immutability of a global variable must be realized by applying a const qualifier on its type. The variable type must be specified at the definition time, which means `static_addr_of` can't just call `static_addr_of_mut` and change the variable's immutability later, hence the addition of `static_addr_of_inner`.

Realizing immutability in this way has a side effect: the pointer taken from an immutable global variable now has a different type. In one case (https://github.com/rust-lang/rustc_codegen_gcc/pull/165#discussion_r857233579) this actually caused a type mismatch in a module that is seemingly unconnected to this change. This particular case was mitigated by #170, but we don't know to what extent this will cause problems in the present or the future. This is a problem specific to libgccjit, so cg_llvm will continue to assume that the pointer types of immutable global variables and mutable ones are identical, and whenever we port its changes to cg_gcc, we risk breakage because this assumption does not hold for libgccjit. This brings us to a solution: Cast the taken pointer immediately ("canonicalization"), thus localizing the problem to a module that is responsible for dealing with global variables.

In conclusion, doing canonicalization here keeps the code free of latent bugs, improves the code quality, and minimizes risks, which is why I'm still in favor of doing this.

antoyo:
That makes a lot of sense, so let's keep it that way.

You could please add these explanations in the code as comments?

antoyo:
I believe I could make the equivalent of `LLVMSetGlobalConstant` in libgccjit.
Do you think it would be worth it?

(Sorry for all the back-and-forth.)

Thanks for all your work!

yvt:
> I believe I could make the equivalent of `LLVMSetGlobalConstant` in libgccjit.
> Do you think it would be worth it?

Yes, I think so. Though, I'm curious how willing the libgccjit maintainers are to accept the patch, especially considering that it's not strictly necessary and doesn't directly map to any of [the existing GCC variable attributes](https://gcc.gnu.org/onlinedocs/gcc/Common-Variable-Attributes.html#Common-Variable-Attributes).

yvt:
In that case, I'd say go for it.

antoyo:
Ok, I added the function in [this commit](https://github.com/antoyo/gcc/commit/6c00b2a3b02d67cdf009d248103e821d784f4ace) and in [the bindings](https://github.com/antoyo/gccjit.rs/commit/ecd3560dc7b9626529855c2f29379dcdbfac92c4) as well.

antoyo:
I just wanted to make sure that using the new readonly API it gives the same result.
Is it the case?

yvt:
Yes, it's a perfect match `size`-wise.

```
$ cargo new --bin hello
$ cd hello
$ echo '[profile.release]' >> Cargo.toml
$ echo 'panic = "abort"' >> Cargo.toml
$ echo 'codegen-units = 1' >> Cargo.toml
$ echo 'codegen-units = 1' >> rustc_codegen_gcc/build_sysroot/Cargo.toml

---- cg_gcc master ----
$ rustc_codegen_gcc/cargo.sh build --release
$ size target/release/hello
   text    data     bss     dec     hex filename
 311995   30920     392  343307   53d0b target/release/hello

---- cg_gcc 7bb9fd8 (old) ----
$ size target/release/hello
   text    data     bss     dec     hex filename
 339755    1848       8  341611   5366b target/release/hello

---- cg_gcc b8c82b1 (new) ----
$ size target/release/hello
   text    data     bss     dec     hex filename
 339755    1848       8  341611   5366b target/release/hello
```

src/consts.rs
```diff
@@ -47,7 +47,8 @@ impl<'gcc, 'tcx> StaticMethods for CodegenCx<'gcc, 'tcx> {
             }
         }
         let global_value = self.static_addr_of_mut(cv, align, kind);
-        // TODO(antoyo): set global constant.
+        #[cfg(feature = "master")]
+        self.global_lvalues.borrow().get(&global_value).unwrap().global_set_readonly();
```
antoyo, line 51: Please use `expect()` instead of `unwrap()`.

bjorn3 (approved):

antoyo (approved):
Thanks a lot!
And again, sorry for all the back and forth.

PR #170, opened 2022-05-02
Use the provided pointee type in `<Builder as BuilderMethods>::load`

yvt:
Updates the internal calls to `<Builder as BuilderMethods>::load` so that its `ty` parameter has consistent semantics, enabling this method's implementation to make use of it.

### DIFF
--- a/src/builder.rs
+++ b/src/builder.rs
@@ -652,7 +652,7 @@ impl<'a, 'gcc, 'tcx> BuilderMethods<'a, 'tcx> for Builder<'a, 'gcc, 'tcx> {
         unimplemented!();
     }
 
-    fn load(&mut self, _ty: Type<'gcc>, ptr: RValue<'gcc>, _align: Align) -> RValue<'gcc> {
+    fn load(&mut self, _pointee_ty: Type<'gcc>, ptr: RValue<'gcc>, _align: Align) -> RValue<'gcc> {
         // TODO(antoyo): use ty.
         let block = self.llbb();
         let function = block.get_function();
@@ -715,7 +715,11 @@ impl<'a, 'gcc, 'tcx> BuilderMethods<'a, 'tcx> for Builder<'a, 'gcc, 'tcx> {
                 OperandValue::Ref(place.llval, Some(llextra), place.align)
             }
             else if place.layout.is_gcc_immediate() {
-                let load = self.load(place.llval.get_type(), place.llval, place.align);
+                let load = self.load(
+                    place.layout.gcc_type(self, false),
+                    place.llval,
+                    place.align,
+                );
                 if let abi::Abi::Scalar(ref scalar) = place.layout.abi {
                     scalar_load_metadata(self, load, scalar);
                 }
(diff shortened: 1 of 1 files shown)


src/builder.rs
```diff
@@ -652,18 +652,16 @@ impl<'a, 'gcc, 'tcx> BuilderMethods<'a, 'tcx> for Builder<'a, 'gcc, 'tcx> {
...
-    fn load(&mut self, _ty: Type<'gcc>, ptr: RValue<'gcc>, _align: Align) -> RValue<'gcc> {
-        // TODO(antoyo): use ty.
+    fn load(&mut self, pointee_ty: Type<'gcc>, ptr: RValue<'gcc>, _align: Align) -> RValue<'gcc> {
         let block = self.llbb();
         let function = block.get_function();
         // NOTE: instead of returning the dereference here, we have to assign it to a variable in
         // the current basic block. Otherwise, it could be used in another basic block, causing a
         // dereference after a drop, for instance.
         // TODO(antoyo): handle align of the load instruction.
         let deref = ptr.dereference(None).to_rvalue();
```
bjorn3, line 662: I think you will have to do a pointer cast before dereferencing. Otherwise you may get an assignment of two different types.

antoyo:
What do you mean by "so that its `ty` parameter has consistent semantics"?

Does this fix any bug? I did not use this parameter as it didn't seem required and I thought that could be required by the LLVM API, but maybe not for libgccjit.

bjorn3:
cg_ssa currently has a bunch of pointercasts to satisfy LLVM. With LLVM migrating to opaque pointers, these pointercasts will likely be removed as LLVM doesn't need them, or at the very least they will likely regress. This means that cg_gcc will have to insert pointercasts as necessary during loads and stores based on the given pointee type to satisfy GCC.

src/builder.rs
yvt: Addressed in 76f575608a2a3b3ebf1aa11fdfa3997de6b95777.

yvt:
> What do you mean by "so that its `ty` parameter has consistent semantics"?

When called from `rustc_codegen_ssa`, it contains a **pointee** type. When called from `rustc_codegen_gcc`, it contains a **pointer** type. Commit b8498e4b00b87b300531be562a0fb1f3a48a7baa fixes this discrepancy.

src/builder.rs
```diff
@@ -715,7 +714,11 @@ impl<'a, 'gcc, 'tcx> BuilderMethods<'a, 'tcx> for Builder<'a, 'gcc, 'tcx> {
                 OperandValue::Ref(place.llval, Some(llextra), place.align)
             }
             else if place.layout.is_gcc_immediate() {
-                let load = self.load(place.llval.get_type(), place.llval, place.align);
+                let load = self.load(
+                    place.llval.get_type().get_pointee().expect("get_pointee"),
```
antoyo, line 718: One thing I'm wondering is whether we should get the type the same way it is done in the LLVM codegen, that is:
```suggestion
                    place.layout.llvm_type(self),
```

yvt: Done. Had to pass an extra parameter (`set_fields`) to `gcc_type`.

antoyo (approved):
Thanks!
This looks good to me.

Is this ready to be merged?

yvt:
It's ready to merge.

PR #171, opened 2022-05-03
Update intrinsics conversion generation

GuillaumeGomez:


tools/generate_intrinsics.py
```diff
@@ -0,0 +1,119 @@
...
+                print("Skipping repository update.")
+                return
+            elif choice.lower() == "y":
+                print("Updating repository...")
+                run_command(["git", "pull", "origin"], cwd="llvm-project")
+                return
+            else:
+                print("Didn't understand answer...")
+    print("Cloning LLVM repository...")
+    run_command(["git", "clone", "https://github.com/llvm/llvm-project", "--depth", "1", LLVM_PATH])
```
bjorn3, line 34: Maybe use a blobless clone with a sparse checkout? Something like

```sh
$ git clone https://github.com/llvm/llvm-project.git --filter=tree:0 --no-checkout
$ cd llvm-project
$ git sparse-checkout init
$ git sparse-checkout set add llvm/include/llvm/IR
```

This downloads 125MB resulting in a 160MB checkout without github spending any time to create a new packfile from scratch as `--depth 1` requires. For reference `--depth 1` downloads 170MB resulting in a 1.6GB checkout while taking longer.

src/intrinsic/archs.rs
```diff
@@ -1,917 +1,234 @@
...
+    "llvm.amdgcn.s.getpc" => "__builtin_amdgcn_s_getpc",
+    "llvm.amdgcn.s.getreg" => "__builtin_amdgcn_s_getreg",
+    "llvm.amdgcn.s.memrealtime" => "__builtin_amdgcn_s_memrealtime",
+    "llvm.amdgcn.s.memtime" => "__builtin_amdgcn_s_memtime",
+    "llvm.amdgcn.s.sendmsg" => "__builtin_amdgcn_s_sendmsg",
+    "llvm.amdgcn.s.sendmsghalt" => "__builtin_amdgcn_s_sendmsghalt",
+    "llvm.amdgcn.s.setprio" => "__builtin_amdgcn_s_setprio",
+    "llvm.amdgcn.s.setreg" => "__builtin_amdgcn_s_setreg",
+    "llvm.amdgcn.s.sleep" => "__builtin_amdgcn_s_decperflevel",
+    "llvm.amdgcn.s.sleep" => "__builtin_amdgcn_s_incperflevel",
```
bjorn3, line 94: As indicated by the ci warning this is a duplicate of the above line except with a different rhs.

src/intrinsic/archs.rs
```diff
@@ -1,917 +1,234 @@
...
+    "llvm.amdgcn.s.sleep" => "__builtin_amdgcn_s_sleep",
```
bjorn3, line 95: Same

src/intrinsic/archs.rs
```diff
@@ -2167,53 +1084,221 @@ match name {
...
+    "llvm.ve.vl.extract.vm512l" => "__builtin_ve_vl_extract_vm512l",
+    "llvm.ve.vl.extract.vm512u" => "__builtin_ve_vl_extract_vm512u",
+    "llvm.ve.vl.insert.vm512l" => "__builtin_ve_vl_insert_vm512l",
+    "llvm.ve.vl.insert.vm512u" => "__builtin_ve_vl_insert_vm512u",
+    "llvm.ve.vl.pack.f32a" => "__builtin_ve_vl_pack_f32a",
+    "llvm.ve.vl.pack.f32p" => "__builtin_ve_vl_pack_f32p",
     // x86
-    "llvm.x86.addcarry.u32" => "__builtin_ia32_addcarry_u32",
-    "llvm.x86.addcarry.u64" => "__builtin_ia32_addcarry_u64",
-    "llvm.x86.addcarryx.u32" => "__builtin_ia32_addcarryx_u32",
```
antoyo, line 2195: Unfortunately, this one is used by rustc and I guess there are a few others in this case (like `llvm.x86.avx.cmp.pd.256`). Could you please somehow combine this new script with the old one to include the intrinsics from both sources?

tools/generate_intrinsics.py
GuillaumeGomez: I didn't know about this. Let me give it a try.

src/intrinsic/archs.rs
GuillaumeGomez: What should we do in such a case?

tools/generate_intrinsics.py
GuillaumeGomez: The end result is great! Thanks for this. :)

GuillaumeGomez:
So for now, if I have a same LLVM intrinsic with multiple GCC conversions, I pick the first one in alpha sort and comment the following ones.

src/intrinsic/archs.rs
bjorn3: How can a single llvm intrinsic correspond to multiple gcc intrinsics which look like they have different semantics?

tools/generate_intrinsics.py
```diff
@@ -60,18 +60,20 @@ def extract_instrinsics(intrinsics, file):
...
             if len(entries) > 0:
-                intrinsic = content.split(":")[0].split(" ")[1].strip()
+                intrinsic = content.split("def ")[1].strip().split(":")[0].strip()
                 intrinsic = intrinsic.split("_")
                 if len(intrinsic) < 2 or intrinsic[0] != "int":
                     continue
                 intrinsic[0] = "llvm"
                 intrinsic = ".".join(intrinsic)
+                # if intrinsic == "llvm.amdgcn.s.sleep":
+                #     import pdb;pdb.set_trace()
```
bjorn3, line 76: nit: leftover debug comment

GuillaumeGomez: Thanks!

GuillaumeGomez:
After a discussion with @antoyo, if we have an LLVM intrinsic with different translations, I'll pick the one from `llvm-project` over the one from `llvmint`.

antoyo (approved):
Thanks a lot!

src/intrinsic/archs.rs
GuillaumeGomez: llvmint and llvm can define the same intrinsic with different translations. The rest was bug.

PR #175, opened 2022-05-15
Update llvmint

GuillaumeGomez:
I updated my llvmint repository by including changes from https://github.com/aweinstock314/llvmint which resulted into one new intrinsic.


GuillaumeGomez:
The problem was that I wasn't regenerating the `json` file from `llvmint`. It's now fixed and the list is complete.

GuillaumeGomez:
To fix the removal issue, I keep both `llvmint` repositories.

Issue #169, opened 2022-05-01
Investigate making the CI faster


antoyo:
We can probably run all these tests in parallel (in different CI jobs):

 * std_tests
 * test_libcore
 * extended_sysroot_tests
 * test_rustc

KirtiGhugtyal6:
Where are these tests located

bjorn3:
These are the functions defined in https://github.com/rust-lang/rustc_codegen_gcc/blob/master/test.sh

PR #184, opened 2022-06-15
setup: simplify build procedure

D3PSI:
removed redundant manual configuration of environment
added llvm as git submodule (requires recursive clone)

this PR aims to reduce the amount of manual configuration required to build this project. i am working on addressing https://github.com/rust-lang/rustc_codegen_gcc/issues/169 at the same time as well. also, the build setup seems a bit too complicated in general, trying to simplify.

for the moment this build will fail.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -15,11 +15,8 @@ jobs:
 
     steps:
     - uses: actions/checkout@v2
-
-    - uses: actions/checkout@v2
       with:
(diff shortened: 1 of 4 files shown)


bjorn3:
This would force a full clone of the huge llvm-project repo even if you already have a copy of it somewhere else you can use. (exact version doesn't matter at all anyway) This is especially bad for rust-lang/rust which would now get two different llvm-project submodules.

PR #188, opened 2022-06-24
Simplify CI

GuillaumeGomez:
Part of #169.


.github/workflows/ci.yml
```diff
@@ -106,8 +100,29 @@ jobs:
...
-    - name: Test
-      if: matrix.libgccjit_version != 'libgccjit12.so'
+    - name: Test std_tests
+      run: |
+        # Enable backtraces for easier debugging
+        export RUST_BACKTRACE=1
+
+        # Reduce amount of benchmark runs as they are slow
+        export COMPILE_RUNS=2
+        export RUN_RUNS=2
```
antoyo, line 110: If you look in `test.sh`, you can see those environment variables are unused for now. I'm not sure if you have an idea about how to organize benchmark runs in the file, but I'm assuming those will run in parallel as well in the future.
So, if that's the case, please remove all references of them in the CI and we'll add them back in their own parallel job when we actually run them in the CI.

.github/workflows/ci.yml
```diff
@@ -106,8 +100,29 @@ jobs:
         command: build
         args: --release
 
-    - name: Test
-      if: matrix.libgccjit_version != 'libgccjit12.so'
+    - name: Test std_tests
+      run: |
+        # Enable backtraces for easier debugging
+        export RUST_BACKTRACE=1
```
antoyo, line 106: Perhaps we should enable this at the top level to avoid having to specify it multiple times?

.github/workflows/ci.yml
```diff
@@ -128,7 +142,7 @@ jobs:
         export COMPILE_RUNS=2
         export RUN_RUNS=2
 
-        ./test.sh --release --no-default-features
+        ./test.sh --release --test-rustc ${{ matrix.libgccjit_version.extra }}
```
antoyo, line 145: It seems like the `mini_test` function from the `test.sh` script is not ran anymore. Please add a function and CI job for it.

GuillaumeGomez: Oh indeed, adding it.

PR #201, opened 2022-07-09
Add cache for rust repository

GuillaumeGomez:


test.sh
```diff
@@ -290,7 +290,9 @@ function test_rustc() {
 
     rust_toolchain=$(cat rust-toolchain | grep channel | sed 's/channel = "\(.*\)"/\1/')
 
-    git clone https://github.com/rust-lang/rust.git || true
+    if [ ! -d "rust" ]; then
```
antoyo, line 293: Is this condition necessary, since the following line uses `|| true`?

GuillaumeGomez: This is a good point. Reverting this one.

PR #209, opened 2022-08-27
2022/08/26 sync from rust

antoyo:


src/builder.rs
```diff
@@ -1185,6 +1185,15 @@ impl<'a, 'gcc, 'tcx> BuilderMethods<'a, 'tcx> for Builder<'a, 'gcc, 'tcx> {
     fn atomic_cmpxchg(&mut self, dst: RValue<'gcc>, cmp: RValue<'gcc>, src: RValue<'gcc>, order: AtomicOrdering, failure_order: AtomicOrdering, weak: bool) -> RValue<'gcc> {
         let expected = self.current_func().new_local(None, cmp.get_type(), "expected");
         self.llbb().add_assignment(None, expected, cmp);
+        // NOTE: gcc doesn't support a failure memory model that is stronger than the success
+        // memory model.
+        let failure_order =
+            if failure_order as i32 > order as i32 {
+                AtomicOrdering::Relaxed
```
bjorn3, line 1192: Shouldn't it strengthen the success case rather than weaken the failure case?

antoyo: Thanks! Fixed.

PR #214, opened 2022-08-31
simd: implement simd_fmin/fmax

sadlerap:
This implements simd_fmin/fmax in a largely-optimal method.

### DIFF
--- a/failing-ui-tests12.txt
+++ b/failing-ui-tests12.txt
@@ -8,6 +8,7 @@ src/test/ui/packed/packed-struct-size.rs
 src/test/ui/packed/packed-struct-vec.rs
 src/test/ui/packed/packed-tuple-struct-layout.rs
 src/test/ui/simd/array-type.rs
+src/test/ui/simd/intrinsic/float-minmax-pass.rs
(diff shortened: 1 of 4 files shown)


failing-ui-tests12.txt
```diff
@@ -8,6 +8,7 @@ src/test/ui/packed/packed-struct-size.rs
 src/test/ui/packed/packed-struct-vec.rs
 src/test/ui/packed/packed-tuple-struct-layout.rs
 src/test/ui/simd/array-type.rs
+src/test/ui/simd/intrinsic/float-minmax-pass.rs
```
sadlerap, line 11: This fails with a type error from within gcc (there's a bitcast we make that has incompatible source & destination types).  I'm not entirely sure what's causing this, but considering this doesn't fail with the patched versions of libgccjit, I'm pretty sure this is safe to ignore for now.

antoyo: I'm not exactly sure what's causing this. Perhaps it was fixed by [this commit](https://github.com/antoyo/gcc/commit/6fdab49ae779e60e1e4fd85111be1fb44fe8c225) which could be why it fails with libgccjit12.

antoyo (approved):
Thanks a lot!

PR #215, opened 2022-09-09
simd: impl extract_element for vector types

sadlerap:
This fixes some tests that needed vector element extraction.

### DIFF
--- a/src/builder.rs
+++ b/src/builder.rs
@@ -1059,8 +1059,19 @@ impl<'a, 'gcc, 'tcx> BuilderMethods<'a, 'tcx> for Builder<'a, 'gcc, 'tcx> {
         unimplemented!();
     }
 
-    fn extract_element(&mut self, _vec: RValue<'gcc>, _idx: RValue<'gcc>) -> RValue<'gcc> {
-        unimplemented!();
+    #[cfg(feature="master")]
+    fn extract_element(&mut self, vec: RValue<'gcc>, idx: RValue<'gcc>) -> RValue<'gcc> {
+        self.context.new_vector_access(None, vec, idx).to_rvalue()
+    }
+
+    #[cfg(not(feature="master"))]
+    fn extract_element(&mut self, vec: RValue<'gcc>, idx: RValue<'gcc>) -> RValue<'gcc> {
+        let vector_type = vec.get_type().unqualified().dyncast_vector().expect("Called extract_element on a non-vector type");
+        let element_type = vector_type.get_element_type();
+        let vec_num_units = vector_type.get_num_units();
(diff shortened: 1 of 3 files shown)


src/builder.rs
```diff
@@ -1059,8 +1059,20 @@ impl<'a, 'gcc, 'tcx> BuilderMethods<'a, 'tcx> for Builder<'a, 'gcc, 'tcx> {
...
+
+    // TODO(antoyo): remove this when new_vector_access stabilizes
+    #[cfg(not(feature="master"))]
+    fn extract_element(&mut self, vec: RValue<'gcc>, idx: RValue<'gcc>) -> RValue<'gcc> {
+        let vector_type = vec.get_type().unqualified().dyncast_vector().expect("Called extract_element on a non-vector type");
+        let element_type = vector_type.get_element_type();
+        let vec_num_units = vector_type.get_num_units();
+        let array_type = self.context.new_array_type(None, element_type, vec_num_units as i32);
+        let array = self.context.new_bitcast(None, vec, array_type).to_rvalue();
+        self.context.new_array_access(None, array, idx).to_rvalue()
```
antoyo, line 1075: Does this sometimes fail in libgccjit12 because of an error like "array not addressable"?
If so, perhaps we could store the value on the stack to make the test pass with libgccjit 12 as well.

sadlerap: `generic-comparison-pass.rs` failures [look to be related to floating-point comparison types not being what was expected](https://github.com/rust-lang/rustc_codegen_gcc/runs/8262086841?check_suite_focus=true#step:19:14582).  `generics.rs` passes completely in all three libgccjit cases.

src/builder.rs
```diff
@@ -1059,8 +1059,20 @@ impl<'a, 'gcc, 'tcx> BuilderMethods<'a, 'tcx> for Builder<'a, 'gcc, 'tcx> {
...
     }
 
-    fn extract_element(&mut self, _vec: RValue<'gcc>, _idx: RValue<'gcc>) -> RValue<'gcc> {
-        unimplemented!();
+    #[cfg(feature="master")]
+    fn extract_element(&mut self, vec: RValue<'gcc>, idx: RValue<'gcc>) -> RValue<'gcc> {
+        self.context.new_vector_access(None, vec, idx).to_rvalue()
+    }
+
+    // TODO(antoyo): remove this when new_vector_access stabilizes
```
antoyo, line 1067: Please remove this comment.
I don't find it necessary since I'll just look at every cfg when it's time to remove that kind of stuff.

antoyo:
Thanks a lot!

PR #219, opened 2022-09-15
simd: implement float math intrinsics

sadlerap:
Implements the intrinsics required to pass float-math-pass and libm_std_can_float ui tests.

### DIFF
--- a/src/intrinsic/simd.rs
+++ b/src/intrinsic/simd.rs
@@ -14,7 +14,6 @@ use rustc_span::{Span, Symbol, sym};
 use rustc_target::abi::Align;
 
 use crate::builder::Builder;
-use crate::intrinsic;
 
 pub fn generic_simd_intrinsic<'a, 'gcc, 'tcx>(bx: &mut Builder<'a, 'gcc, 'tcx>, name: Symbol, callee_ty: Ty<'tcx>, args: &[OperandRef<'tcx, RValue<'gcc>>], ret_ty: Ty<'tcx>, llret_ty: Type<'gcc>, span: Span) -> Result<RValue<'gcc>, ()> {
     // macros for error handling:
@@ -415,8 +414,8 @@ pub fn generic_simd_intrinsic<'a, 'gcc, 'tcx>(bx: &mut Builder<'a, 'gcc, 'tcx>, 
             if let ty::Float(f) = in_elem.kind() {
                 let elem_ty = bx.cx.type_float_from_ty(*f);
                 match f.bit_width() {
-                    32 => ("f32", elem_ty),
-                    64 => ("f64", elem_ty),
+                    32 => ("f", elem_ty),
+                    64 => ("", elem_ty),
                     _ => {
                         return_error!(
                             "unsupported element type `{}` of floating-point vector `{}`",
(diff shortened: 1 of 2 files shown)


src/intrinsic/simd.rs
```diff
@@ -432,30 +431,47 @@ pub fn generic_simd_intrinsic<'a, 'gcc, 'tcx>(bx: &mut Builder<'a, 'gcc, 'tcx>,
...
             };
-        let llvm_name = &format!("llvm.{0}.v{1}{2}", intr_name, in_len, elem_ty_str);
-        let function = intrinsic::llvm::intrinsic(llvm_name, &bx.cx);
-        let function: RValue<'gcc> = unsafe { std::mem::transmute(function) };
-        let c = bx.call(fn_ty, function, &args.iter().map(|arg| arg.immediate()).collect::<Vec<_>>(), None);
+        let builtin_name = format!("{0}{1}", intr_name, elem_ty_str);
+        let funcs = bx.cx.functions.borrow();
+        let function = funcs.get(&builtin_name).unwrap();
+
+        let mut vector_elements = vec![];
```
antoyo, line 458: Can you please add a `TODO` to use a more optimal way to do this on architectures that have an instruction for those vector operations, probably only GPUs?

src/intrinsic/simd.rs
```diff
@@ -432,30 +431,47 @@ pub fn generic_simd_intrinsic<'a, 'gcc, 'tcx>(bx: &mut Builder<'a, 'gcc, 'tcx>,
...
+                sym::simd_trunc => ("trunc", 1),
                 _ => return_error!("unrecognized intrinsic `{}`", name),
             };
-        let llvm_name = &format!("llvm.{0}.v{1}{2}", intr_name, in_len, elem_ty_str);
-        let function = intrinsic::llvm::intrinsic(llvm_name, &bx.cx);
-        let function: RValue<'gcc> = unsafe { std::mem::transmute(function) };
-        let c = bx.call(fn_ty, function, &args.iter().map(|arg| arg.immediate()).collect::<Vec<_>>(), None);
+        let builtin_name = format!("{0}{1}", intr_name, elem_ty_str);
+        let funcs = bx.cx.functions.borrow();
+        let function = funcs.get(&builtin_name).unwrap();
```
antoyo, line 456: Please use `expect()` instead of `unwrap()`.
I'll add the clippy lint for that.

src/intrinsic/simd.rs
```diff
@@ -432,30 +431,47 @@ pub fn generic_simd_intrinsic<'a, 'gcc, 'tcx>(bx: &mut Builder<'a, 'gcc, 'tcx>,
 
         let vec_ty = bx.cx.type_vector(elem_ty, in_len);
 
-        let (intr_name, fn_ty) =
+        let (intr_name, num_args) =
```
antoyo, line 434: Can you please add a comment to explain why `num_args` is useful here?
I don't find it obvious: is it because `args` contain more arguments for LLVM in some cases?

antoyo (commented):
Good job!
A few nits.

src/intrinsic/simd.rs
sadlerap: I had `unwrap()` because this lookup should never fail; we declare all of these builtins [here](https://github.com/rust-lang/rustc_codegen_gcc/blob/master/src/context.rs#L173-L187), so unwrapping should be safe to do.  That said, this should probably be a little more expressive in what failed and why if the lookup fails.

Would something like `unwrap_or_else(|_| panic!("unable to find builtin function {}", builtin_name))` be acceptable?  If we failed to lookup a builtin, it would be nice to report what builtin we missed.

src/intrinsic/simd.rs
sadlerap: I had this because I was trying to be defensive about argument handling and only pass the number of arguments each function was expecting.  On further review, `args` should always be correct coming from rustc (AFAICT validation of these intrinsics' signatures happens before we see it), so it's probably safe to refactor `num_args` out.

src/intrinsic/simd.rs
antoyo: Yes, that would be perfect.

sadlerap: Fixed in 9691257.

src/intrinsic/simd.rs
sadlerap: Fixed in 9691257.

src/intrinsic/simd.rs
sadlerap: Fixed in 9691257.

src/intrinsic/simd.rs
```diff
@@ -432,30 +431,49 @@ pub fn generic_simd_intrinsic<'a, 'gcc, 'tcx>(bx: &mut Builder<'a, 'gcc, 'tcx>,
...
+                sym::simd_fsqrt => "sqrt",
+                sym::simd_round => "round",
+                sym::simd_trunc => "trunc",
                 _ => return_error!("unrecognized intrinsic `{}`", name),
             };
-        let llvm_name = &format!("llvm.{0}.v{1}{2}", intr_name, in_len, elem_ty_str);
-        let function = intrinsic::llvm::intrinsic(llvm_name, &bx.cx);
-        let function: RValue<'gcc> = unsafe { std::mem::transmute(function) };
-        let c = bx.call(fn_ty, function, &args.iter().map(|arg| arg.immediate()).collect::<Vec<_>>(), None);
+        let builtin_name = format!("{0}{1}", intr_name, elem_ty_str);
```
antoyo, line 454: I see the original code has this as well, but can you please remove the numbers in the format string?
```suggestion
        let builtin_name = format!("{}{}", intr_name, elem_ty_str);
```

antoyo (commented):
Sorry, I missed that one.

src/intrinsic/simd.rs
sadlerap: Fixed in f7694d0.

antoyo (approved):
Thanks for your work!

There are some issues with the CI, so I'll have to wait a bit before merging this PR.

sadlerap:
Rebased.

antoyo:
Thanks!

PR #228, opened 2022-10-03
simd: enable simd_as intrinsic

sadlerap:
The method context.convert_vector, added to libgccjit for `simd_cast`, appears to give the correct behavior for `simd_as`.  Instead of special-casing `simd_as`, re-use `simd_cast`'s impl for `simd_as`.


antoyo (approved):
Thanks!

Issue #60, opened 2021-08-12
Finish implementing the conversion from register class to gcc constraint code


antoyo:
[Here](https://github.com/antoyo/rustc_codegen_gcc/blob/8b6a696e965e67ce9b3b45b8b66425dca728d81e/src/asm.rs#L416), [here](https://github.com/antoyo/rustc_codegen_gcc/blob/8b6a696e965e67ce9b3b45b8b66425dca728d81e/src/asm.rs#L467) and [here](https://github.com/antoyo/rustc_codegen_gcc/blob/8b6a696e965e67ce9b3b45b8b66425dca728d81e/src/asm.rs#L576).

The [GCC constraints doc](https://gcc.gnu.org/onlinedocs/gcc/Constraints.html) will be helpful to do this.

Commeownist:
I'll try to tackle this one. 

Btw, I think you were right that support for explicit registers like `in("r11")` will require creating "GCC register variables" which is not yet supported by gccjit yet. I guess that'll be one more libgccjit patch. Anyway, I don't  think this kind of syntax is used often;  `syscall` is the only example I can come up with. 

For now, I'm going to emit an error. The code as it's now is definitely incorrect.

bjorn3:
Explicit registers are used by compiler-builtins. Almost every input and output uses an explicit register in https://github.com/rust-lang/compiler-builtins/blob/826d9e9d41dca79c9893118fdbb600fc9a837764/src/mem/x86_64.rs

Tock uses them when switching to userspace: https://github.com/tock/tock/blob/bd4808c787434f6d493f643ea7f9abe8339048e0/arch/cortex-m0/src/lib.rs#L204-L260

The x86_64 crate uses them for ports (https://github.com/rust-osdev/x86_64/blob/a9cbf1471d38b18dcfe22bf399082a445341c3e5/src/instructions/port.rs#L14) and model specific registers (https://github.com/rust-osdev/x86_64/blob/a9cbf1471d38b18dcfe22bf399082a445341c3e5/src/registers/model_specific.rs#L134-L139)

Commeownist:
@bjorn3 I agree with all that you said, let me clarify. (I'm using x86-64 as example; everything I say applies to other architectures as well)

Here's the code part I was referring to:

https://github.com/antoyo/rustc_codegen_gcc/blob/6f50986667debbfc67776304a8ee23fe0158613f/src/asm.rs#L235-L250

[...]

PR #232, opened 2022-10-13
Add missing register class conversion for inline asm

GuillaumeGomez:
Part of https://github.com/rust-lang/rustc_codegen_gcc/issues/60.

### DIFF
--- a/src/asm.rs
+++ b/src/asm.rs
@@ -768,34 +768,41 @@ impl<'gcc, 'tcx> AsmMethods<'tcx> for CodegenCx<'gcc, 'tcx> {
 }
 
 fn modifier_to_gcc(arch: InlineAsmArch, reg: InlineAsmRegClass, modifier: Option<char>) -> Option<char> {
+    // The modifiers can be retrieved from
+    // https://gcc.gnu.org/onlinedocs/gcc/Modifiers.html#Modifiers
     match reg {
         InlineAsmRegClass::AArch64(AArch64InlineAsmRegClass::reg) => modifier,
-        InlineAsmRegClass::AArch64(AArch64InlineAsmRegClass::preg) => modifier,
         InlineAsmRegClass::AArch64(AArch64InlineAsmRegClass::vreg)
         | InlineAsmRegClass::AArch64(AArch64InlineAsmRegClass::vreg_low16) => {
-            unimplemented!()
+            if modifier == Some('v') { None } else { modifier }
         }
-        InlineAsmRegClass::Arm(ArmInlineAsmRegClass::reg)  => unimplemented!(),
+        InlineAsmRegClass::AArch64(AArch64InlineAsmRegClass::preg) => {
+            unreachable!("clobber-only")
+        }
+        InlineAsmRegClass::Arm(ArmInlineAsmRegClass::reg) => None,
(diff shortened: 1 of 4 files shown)


src/asm.rs
```diff
@@ -564,56 +576,77 @@ fn reg_to_gcc(reg: InlineAsmRegOrRegClass) -> ConstraintOrRegister {
...
             }
         },
+        // They can be retrieved from https://gcc.gnu.org/onlinedocs/gcc/Machine-Constraints.html
         InlineAsmRegOrRegClass::RegClass(reg) => match reg {
-            InlineAsmRegClass::AArch64(AArch64InlineAsmRegClass::preg) => unimplemented!(),
-            InlineAsmRegClass::AArch64(AArch64InlineAsmRegClass::reg) => unimplemented!(),
-            InlineAsmRegClass::AArch64(AArch64InlineAsmRegClass::vreg) => unimplemented!(),
-            InlineAsmRegClass::AArch64(AArch64InlineAsmRegClass::vreg_low16) => unimplemented!(),
-            InlineAsmRegClass::Arm(ArmInlineAsmRegClass::reg) => unimplemented!(),
+            InlineAsmRegClass::AArch64(AArch64InlineAsmRegClass::reg) => {
```
Amanieu, line 581: This should just be `"r"`. `L` and `K` are for integer constants, not registers.

GuillaumeGomez: Thanks, fixed!

src/base.rs
```diff
@@ -112,7 +112,9 @@ pub fn compile_codegen_unit<'tcx>(tcx: TyCtxt<'tcx>, cgu_name: Symbol, supports_
         context.add_command_line_option("-mavx");
 
         for arg in &tcx.sess.opts.cg.llvm_args {
-            context.add_command_line_option(arg);
+            if !arg.starts_with("--x86-asm-syntax=") {
+                context.add_command_line_option(arg);
+            }
```
bjorn3, line 117: This would break anyone that needs a specific assembly syntax in rustc's output, right?

antoyo: We're currently doing some tests.

This option is an LLVM option; for gcc it should be `-masm=intel`. I'm not sure what's the plan for supporting setting options for the different backends, but cg_gcc currently uses `llvm-args` to set gcc options. Since this option is to send arguments to the backend, I'm not sure it makes sense to do the mapping between the LLVM options and GCC options.

I guess we would need to discuss this somewhere to agree on the correct solution, though.

src/asm.rs
```diff
@@ -691,7 +706,10 @@ impl<'gcc, 'tcx> AsmMethods<'tcx> for CodegenCx<'gcc, 'tcx> {
             && options.contains(InlineAsmOptions::ATT_SYNTAX);
 
         // Build the template string
-        let mut template_str = String::new();
+        let mut template_str = ".pushsection .text\n".to_owned();
```
antoyo, line 709: Note to self: check why the change to these directives was done in the past to make sure this change is correct.

GuillaumeGomez: This change itself is just to make the code easier to follow, it still outputs the same thing.

antoyo (approved):
Thanks a lot!

PR #233, opened 2022-10-15
Bump Github actions

paulcacheux:
Recent Github actions runs display the following warning:
```
Node.js 12 actions are deprecated. For more information see: https://github.blog/changelog/2022-09-22-github-actions-all-actions-will-begin-running-on-node16-instead-of-node12/. Please update the following actions to use Node.js 16: actions/checkout, actions/checkout, actions/cache, actions/cache, actions/cache, actions/cache, actions-rs/cargo, actions/cache, actions/cache, actions/cache, actions/cache, actions/checkout, actions/checkout
```

This PR fixes this issue by bumping actions dependencies to use new releases based on node16.
This PR also remove cargo cache fixups that are not needed anymore.

The failing jobs (`--test-failing-rustc`) are the same as on main so not related to this PR.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -35,9 +35,9 @@ jobs:
         ]
 
     steps:
-    - uses: actions/checkout@v2
+    - uses: actions/checkout@v3
 
-    - uses: actions/checkout@v2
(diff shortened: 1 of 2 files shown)


antoyo:
Thanks!

PR #234, opened 2022-10-15
Remove not applicable rustc patch

paulcacheux:
Currently the CI displays:
```
2022-10-15T14:12:46.1306418Z HEAD is now at 7480389611f Auto merge of #100911 - tmiasko:update-stdarch, r=Amanieu
2022-10-15T14:12:46.1313753Z error: patch failed: src/tools/compiletest/src/header.rs:806
2022-10-15T14:12:46.1330234Z ##[error]error: src/tools/compiletest/src/header.rs: patch does not apply
```

This is because the rustc patch is no longer necessary.
This PR fixes this by simply removing the patch. Happy to remove the whole directory if desired (instead of the current `.gitkeep` solution).


antoyo:
Thanks for this PR.

I cannot remember why exactly that patch was needed and I'm not sure whether we'll need more patches like this in the future, so please remove the `.gitkeep` file.

PR #237, opened 2022-10-20
Fix configure command line

GuillaumeGomez:


Readme.md
```diff
@@ -26,12 +26,13 @@ $ cd gcc
 $ sudo apt install flex libmpfr-dev libgmp-dev libmpc3 libmpc-dev
 $ ./configure \
    --enable-host-shared \
-   --enable-languages=jit \
+   --enable-languages=jit,c,c++ \
```
bjorn3, line 29: The C and C++ frontends are not necessary right?

antoyo: They are for running libgccjit tests, so perhaps we should mention that.

GuillaumeGomez: I'm currently checking without the C, however the C++ one is required for tests.

PR #244, opened 2023-01-08
Implement unwinding

antoyo:


src/base.rs
```diff
@@ -147,6 +157,7 @@ pub fn compile_codegen_unit<'tcx>(tcx: TyCtxt<'tcx>, cgu_name: Symbol, supports_
         }
 
         // TODO(bjorn3): Remove once unwinding is properly implemented
+        // TODO: remove.
         context.set_allow_unreachable_blocks(true);
```
antoyo, line 161: @bjorn3: Do you remember the reason that some blocks were unreachable when you implemented a dummy unwinding support?
I have proper unwinding in this PR, but I still get unreachable blocks.

bjorn3: The reason back then was that cleanup blocks would get codegened when using `-Cpanic=unwind` but never jumped to due to unwinding being unimplemented. With real unwinding it shouldn't be necessary anymore. No clue why it is still necessary with this PR.

src/base.rs
```diff
@@ -87,6 +87,16 @@ pub fn compile_codegen_unit<'tcx>(tcx: TyCtxt<'tcx>, cgu_name: Symbol, supports_
         // Instantiate monomorphizations without filling out definitions yet...
         //let llvm_module = ModuleLlvm::new(tcx, &cgu_name.as_str());
         let context = Context::default();
+
+        context.add_command_line_option("-fexceptions");
+        context.add_driver_option("-fexceptions");
+
+        /*context.add_command_line_option("-fasynchronous-unwind-tables");
+        context.add_driver_option("-fasynchronous-unwind-tables");
```
bjorn3, line 95: Asynchronous unwind tables aren't used by the LLVM backend either. Not sure if LLVM supports them in the first place.

src/builder.rs
```diff
@@ -450,8 +451,43 @@ impl<'a, 'gcc, 'tcx> BuilderMethods<'a, 'tcx> for Builder<'a, 'gcc, 'tcx> {
         self.block.end_with_switch(None, value, default_block, &gcc_cases);
     }
 
+    #[cfg(feature="master")]
+    fn invoke(&mut self, typ: Type<'gcc>, func: RValue<'gcc>, args: &[RValue<'gcc>], then: Block<'gcc>, catch: Block<'gcc>, _funclet: Option<&Funclet>) -> RValue<'gcc> {
+        let try_block = self.current_func().new_block("try");
+
+        let current_block = self.block.clone();
+        self.block = try_block;
+        let call = self.call(typ, func, args, None); // TODO: use funclet here?
```
bjorn3, line 460: Funclets are only for implementing SEH unwinding on Windows. GCC may or may not generate them automatically behind the scenes (in which case cg_ssa should be changed to not use funclets with GCC), but LLVM requires explicitly using them instead of landing pads when using SEH. In case you didn't know, funclets are kind of functions, except that they share the stack frame with another function. In SEH a funclet is called by the unwinder for every frame that needs cleanup, unlike on Unix where the unwinder moves the IP to a landingpad and then returns, the landingpad then tail calls back into the unwinder to continue with the next frame.

src/builder.rs
```diff
@@ -450,8 +451,43 @@ impl<'a, 'gcc, 'tcx> BuilderMethods<'a, 'tcx> for Builder<'a, 'gcc, 'tcx> {
...
+            println!("Try/catch in {:?}", self.current_func());
+            self.block.add_try_catch(None, try_block, catch);
+        }
+
+        self.block.end_with_jump(None, then);
+
+        // NOTE: since jumps were added in a place rustc does not expect, the current blocks in the
+        // state need to be updated.
+        // FIXME: not sure it's actually needed.
+        self.switch_to_block(then);
```
bjorn3, line 484: cg_clif expects `invoke` to be a terminator and as such to end with no current block in the builder.

src/builder.rs
```diff
@@ -1160,23 +1196,52 @@ impl<'a, 'gcc, 'tcx> BuilderMethods<'a, 'tcx> for Builder<'a, 'gcc, 'tcx> {
...
+        self.block.add_assignment(None, value.access_field(None, field2), zero); // TODO: set the proper value here (the type of exception?).
+
+        value.to_rvalue()
     }
 
-    fn resume(&mut self, _exn: RValue<'gcc>) {
-        // TODO(bjorn3): Properly implement unwinding.
+    fn resume(&mut self, exn: RValue<'gcc>) {
+        // TODO: check if this is normal that we need to dereference the value.
+        let exn = exn.dereference(None).to_rvalue();
```
bjorn3, line 1240: exn here should be a struct forming a pair with the same values as the landing pad returned. By the way in the latest rustc I changed cleanup_landing_pad and resume to return/accept two separate values rather than a single pair value.

src/builder.rs
```diff
@@ -1160,23 +1196,52 @@ impl<'a, 'gcc, 'tcx> BuilderMethods<'a, 'tcx> for Builder<'a, 'gcc, 'tcx> {
...
+        let param = self.context.new_parameter(None, exn.get_type(), "exn");
+        // TODO(antoyo): should we call __builtin_unwind_resume instead? This might actually be the same.
```
bjorn3, line 1242: Quite possibly. As far as I knwo `_Unwind_Resume` needs to be a tail call to avoid repeatedly unwinding the same frame. `__builtin_unwind_resume` probably handles this for you.

src/base.rs
antoyo: Here's what I saw while trying to debug this.
Here's the MIR of the function:
```
// MIR for `drop_in_place` after AddMovesForPackedDrops

fn drop_in_place(_1: *mut mini_core::Box<&str>) -> () {
    let mut _0: ();                      // return place in scope 0 at /home/bouanto/Ordinateur/Programmation/Rust/Projets/rustc_codegen_gcc/example/mini_core.rs:445:1: 445:56
    let mut _2: ();                      // in scope 0 at /home/bouanto/Ordinateur/Programmation/Rust/Projets/rustc_codegen_gcc/example/mini_core.rs:445:1: 445:56
    let mut _3: ();                      // in scope 0 at /home/bouanto/Ordinateur/Programmation/Rust/Projets/rustc_codegen_gcc/example/mini_core.rs:445:1: 445:56

    bb0: {
        goto -> bb5;                     // scope 0 at /home/bouanto/Ordinateur/Programmation/Rust/Projets/rustc_codegen_gcc/example/mini_core.rs:445:1: 445:56
    }

    bb1: {
        return;                          // scope 0 at /home/bouanto/Ordinateur/Programmation/Rust/Projets/rustc_codegen_gcc/example/mini_core.rs:445:1: 445:56
    }

    bb2 (cleanup): {
        resume;                          // scope 0 at /home/bouanto/Ordinateur/Programmation/Rust/Projets/rustc_codegen_gcc/example/mini_core.rs:445:1: 445:56
    }

    bb3: {
        _2 = mini_core::box_free::<&str>(move ((*_1).0: mini_core::Unique<&str>), move ((*_1).1: mini_core::Global)) -> bb1; // scope 0 at /home/bouanto/Ordinateur/Programmation/Rust/Projets/rustc_codegen_gcc/example/mini_core.rs:445:1: 445:56
                                         // mir::Constant
                                         // + span: /home/bouanto/Ordinateur/Programmation/Rust/Projets/rustc_codegen_gcc/example/mini_core.rs:445:1: 445:56
                                         // + literal: Const { ty: unsafe fn(mini_core::Unique<&str>, ()) {mini_core::box_free::<&str>}, val: Value(<ZST>) }
    }

    bb4 (cleanup): {
        _3 = mini_core::box_free::<&str>(move ((*_1).0: mini_core::Unique<&str>), move ((*_1).1: mini_core::Global)) -> bb2; // scope 0 at /home/bouanto/Ordinateur/Programmation/Rust/Projets/rustc_codegen_gcc/example/mini_core.rs:445:1: 445:56
                                         // mir::Constant
                                         // + span: /home/bouanto/Ordinateur/Programmation/Rust/Projets/rustc_codegen_gcc/example/mini_core.rs:445:1: 445:56
                                         // + literal: Const { ty: unsafe fn(mini_core::Unique<&str>, ()) {mini_core::box_free::<&str>}, val: Value(<ZST>) }
    }

    bb5: {
        drop((*((((*_1).0: mini_core::Unique<&str>).0: mini_core::NonNull<&str>).0: *const &str))) -> [return: bb3, unwind: bb4]; // scope 0 at /home/bouanto/Ordinateur/Programmation/Rust/Projets/rustc_codegen_gcc/example/mini_core.rs:445:1: 445:56
    }
}
```
And here's the GIMPLE:
```
__attribute__((visibility ("hidden")))
void _RINvCsdmDVVNwMhzd_9mini_core13drop_in_placeINtB2_3BoxReEECsb9oj70DTblp_21mini_core_hello_world (struct struct * * param0)
{
  long long int dummyValueThatShouldNeverBeUsed;
  struct  undefined;
  long int * loadedValue96;
  struct struct undefined;
  struct struct undefined;
  int loadedValue95;
  unsigned char * loadedValue94;
  struct struct stack_var_1;
  long long int invokeResult;
  long long int dummyValueThatShouldNeverBeUsed;
  long long int dummyValueThatShouldNeverBeUsed;
  struct landing_pad landing_pad;
  struct  undefined;
  long int * loadedValue93;
  struct struct undefined;
  struct struct * * stack_var_1;
  struct struct undefined;
  struct struct undefined;
  struct struct undefined;

  try
    {
      start:
      stack_var_1.62_1 = &stack_var_1;
      MEM[(struct struct * * *)stack_var_1.62_1] = param0;
      goto bb3;
      bb3:
      loadedValue96 = MEM[(long int * *)param0];
      _RINvCsdmDVVNwMhzd_9mini_core8box_freeReECsb9oj70DTblp_21mini_core_hello_world (loadedValue96);
      dummyValueThatShouldNeverBeUsed = 0;
      goto bb1;
      bb4:        // HERE
      loadedValue93 = MEM[(long int * *)param0];
      try
        {
          _RINvCsdmDVVNwMhzd_9mini_core8box_freeReECsb9oj70DTblp_21mini_core_hello_world (loadedValue93);
          dummyValueThatShouldNeverBeUsed = 0;
          invokeResult = dummyValueThatShouldNeverBeUsed;
          goto bb2;
        }
      finally
        {
          __asm__ __volatile__("");
        }
      else
        {
          _2 = __builtin_eh_pointer (0);
          landing_pad.landing_pad_field_1 = _2;
          landing_pad.landing_pad_field_2 = 0;
          _RNvCsdmDVVNwMhzd_9mini_core15panic_no_unwind ();
          dummyValueThatShouldNeverBeUsed = 0;
          __builtin_unreachable ();
          return;
        }
      goto bb2;
      bb2:
      stack_var_1.63_3 = &stack_var_1;
      loadedValue94 = MEM[(unsigned char * *)stack_var_1.63_3];
      loadedValue95 = stack_var_1.field1_TODO;
      undefined.field0_TODO = loadedValue94;
      undefined.field1_TODO = loadedValue95;
      _Unwind_Resume (undefined);
      __builtin_unreachable ();
      return;
      bb1:
      return;
    }
  finally
    {
      undefined = {CLOBBER(eol)};
      undefined = {CLOBBER(eol)};
      undefined = {CLOBBER(eol)};
      stack_var_1 = {CLOBBER(eol)};
      landing_pad = {CLOBBER(eol)};
      undefined = {CLOBBER(eol)};
      undefined = {CLOBBER(eol)};
      stack_var_1 = {CLOBBER(eol)};
      undefined = {CLOBBER(eol)};
      undefined = {CLOBBER(eol)};
      undefined = {CLOBBER(eol)};
    }
}
```

There's the bb5 basic block that doesn't seem to be generated, which contains an unwind target to bb4 which is the basic block that is unreachable.
Any idea what could cause this?

Thanks for your help!

bjorn3: I guess that bb5 was created for the case where box_free would unwind, but given that it can't unwind, it is unreachable. SimplifyCFG likely doesn't run for shims like the drop glue.

antoyo: Oh, so it's actually normal that the block is unreachable?
The LLVM IR shows this:
```
define internal void @"_ZN4core3ptr53drop_in_place$LT$alloc..boxed..Box$LT$$RF$str$GT$$GT$17h3c013dbafa368214E"(ptr %_1) unnamed_addr #0 personality ptr @rust_eh_personality !dbg !990 {
start:
  %0 = alloca { ptr, i32 }, align 8
  %_1.dbg.spill = alloca ptr, align 8
  store ptr %_1, ptr %_1.dbg.spill, align 8
  call void @llvm.dbg.declare(metadata ptr %_1.dbg.spill, metadata !996, metadata !DIExpression()), !dbg !999
  %_4 = load ptr, ptr %_1, align 8, !dbg !999
  br label %bb3, !dbg !999

bb3:                                              ; preds = %start
  %1 = load ptr, ptr %_1, align 8, !dbg !999, !nonnull !23, !noundef !23
; call alloc::alloc::box_free
  call void @_ZN5alloc5alloc8box_free17h47848291629b205dE(ptr %1), !dbg !999
  ret void, !dbg !999

bb4:                                              ; No predecessors!
  %2 = load ptr, ptr %_1, align 8, !dbg !999, !nonnull !23, !noundef !23
; invoke alloc::alloc::box_free
  invoke void @_ZN5alloc5alloc8box_free17h47848291629b205dE(ptr %2) #19
          to label %bb2 unwind label %abort, !dbg !999

abort:                                            ; preds = %bb4
  %3 = landingpad { ptr, i32 }
          cleanup, !dbg !999
  %4 = extractvalue { ptr, i32 } %3, 0, !dbg !999
  %5 = extractvalue { ptr, i32 } %3, 1, !dbg !999
; call core::panicking::panic_cannot_unwind
  call void @_ZN4core9panicking19panic_cannot_unwind17h8d29da3a0acc193eE() #20, !dbg !999
  unreachable, !dbg !999

bb2:                                              ; preds = %bb4
  %6 = load ptr, ptr %0, align 8, !dbg !999
  %7 = getelementptr inbounds { ptr, i32 }, ptr %0, i32 0, i32 1, !dbg !999
  %8 = load i32, ptr %7, align 8, !dbg !999
  %9 = insertvalue { ptr, i32 } undef, ptr %6, 0, !dbg !999
  %10 = insertvalue { ptr, i32 } %9, i32 %8, 1, !dbg !999
  resume { ptr, i32 } %10, !dbg !999
}
```
It mentions that bb4 has no predecessors, so am I right in assuming that this mean we should keep `context.set_allow_unreachable_blocks(true)`?

bjorn3: Hang on, bb5 is actually reachable through the `goto -> bb5;` in `bb0`. It seems to actually be the unwind edge from `bb5` to `bb4` that is missing. This is likely due to the `drop` in `bb5` being a no-op. I guess it is just a missing optimization that the unwind edge isn't removed. However even with this optimization applied, it can still happen in generic contexts where such an optimization wouldn't know in advance that the type doesn't have any drop glue. Yes, I think `context.set_allow_unreachable_blocks(true)` should be kept.

src/builder.rs
antoyo: Yeah, I haven't sync with rustc for a while because I was having issues. I'll try to do it soon.

bjorn3:
🎉

PR #249, opened 2023-01-23
Fix/signed integer overflow

antoyo:


src/builder.rs
```diff
@@ -503,15 +503,35 @@ impl<'a, 'gcc, 'tcx> BuilderMethods<'a, 'tcx> for Builder<'a, 'gcc, 'tcx> {
     }
 
     fn add(&mut self, a: RValue<'gcc>, b: RValue<'gcc>) -> RValue<'gcc> {
-        self.gcc_add(a, b)
+        let typ = a.get_type();
+        if typ.is_signed(&self.cx) {
+            // NOTE: signed overflow is UB, so cast to unsigned and cast back to the correct type.
```
bjorn3, line 508: There is the `-fwrapv` option to make signed overflow wrap, right?

PR #272, opened 2023-05-22
Fix error in libgccjit12 while creating an array type

antoyo:


antoyo:
Perhaps I removed some patch to the sysroot disabling SIMD.

antoyo:
Superceeded by #357.

PR #301, opened 2023-07-06
Add support for "returns_twice" function attribute

GuillaumeGomez:
Depends on https://github.com/antoyo/gcc/pull/20 and https://github.com/antoyo/gccjit.rs/pull/19.

### DIFF
--- a/src/attributes.rs
+++ b/src/attributes.rs
@@ -118,6 +118,9 @@ pub fn from_fn_attrs<'gcc, 'tcx>(
         if codegen_fn_attrs.flags.contains(CodegenFnAttrFlags::COLD) {
             func.add_attribute(FnAttribute::Cold);
         }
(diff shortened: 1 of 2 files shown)


src/attributes.rs
```diff
@@ -118,6 +118,9 @@ pub fn from_fn_attrs<'gcc, 'tcx>(
         if codegen_fn_attrs.flags.contains(CodegenFnAttrFlags::COLD) {
             func.add_attribute(FnAttribute::Cold);
         }
+        if codegen_fn_attrs.flags.contains(CodegenFnAttrFlags::FFI_RETURNS_TWICE) {
+            to_add.push(FnAttribute::ReturnsTwice);
```
antoyo, line 122: It should be instead:
```suggestion
            func.add_attribute(FnAttribute::ReturnsTwice);
```

PR #304, opened 2023-07-12
Fix asm.rs test

bjorn3:
`add 1, %rax` in AT&T syntax gets interpreted as `add rax, qword ptr [1]`. Use `$` to make it explicit that it is a constant rather than an address.

### DIFF
--- a/tests/run/asm.rs
+++ b/tests/run/asm.rs
@@ -124,7 +124,7 @@ fn main() {
     // check const (ATT syntax)
     let mut x: u64 = 42;
     unsafe {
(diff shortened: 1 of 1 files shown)


bjorn3:
I don't know how this didn't result in a crash for cg_gcc (maybe GCC adds the `$` for you?), but both LLVM and GAS (as used by Cranelift) produce code that crashes.

antoyo:
Thanks for finding this. I fixed it in #330.

cg_gcc added `$` manually.

PR #308, opened 2023-07-21
Add instructions on how to generate GIMPLE format

GuillaumeGomez:


Readme.md
```diff
@@ -74,6 +74,9 @@ To run the tests:
 $ ./test.sh --release
 ```
 
+If you need to check what gccjit is generating (GIMPLE), then take a look at how to
```
antoyo, line 77: Can you please move that below, in the "Debugging" section?

doc/gimple.md
```diff
@@ -0,0 +1,80 @@
...
+
+And finally when you run it:
+
+```console
+LD_LIBRARY_PATH=`pwd`/gcc-build/gcc ./out
+```
+
+It should display:
+
+```console
```
antoyo, line 56: Maybe you meant:
```suggestion
```c
```
?

GuillaumeGomez: It's not really C, so are you sure you want to use the C highlighting?

PR #312, opened 2023-08-12
Add support for `noalias` function parameters

GuillaumeGomez:


src/abi.rs
```diff
@@ -120,30 +120,42 @@ impl<'gcc, 'tcx> FnAbiGccExt<'gcc, 'tcx> for FnAbi<'tcx, Ty<'tcx>> {
                 }
             };
 
+        let apply_attrs = |ty: Type<'gcc>, attrs: &ArgAttributes| {
+            if cfg!(feature = "master") && attrs.regular.contains(ArgAttribute::NoAlias) {
```
antoyo, line 124: You would need to use `#[cfg(…)]` instead to make this compile.

GuillaumeGomez:
We figured out where the bug was coming from: it's because of `ByValue` arguments. It can be reproduced with this code:

```rust
use std::mem;

#[repr(C)]
pub struct Buffer {
    data: *mut u8,
    len: usize,
    drop: extern "C" fn(Buffer),
}

impl Default for Buffer {
    #[inline]
    fn default() -> Self {
        Self::from(vec![])
    }
}

impl Buffer {
    #[inline]
    pub fn new() -> Self {
        Self::default()
    }

    #[inline]
    pub fn push(&mut self, v: u8) {
    }
}

impl From<Vec<u8>> for Buffer {
    fn from(mut v: Vec<u8>) -> Self {
        let (data, len) = (v.as_mut_ptr(), v.len());
        mem::forget(v);

        extern "C" fn drop(b: Buffer) {
        }

        Buffer { data, len, drop }
    }
}

fn main() {
    let mut buffer = Buffer::new();
}
```

Issue #315, opened 2023-08-15
Rewrite bash scripts in Rust


antoyo:
cc @GuillaumeGomez

bjorn3:
Maybe take inspiration from cg_clif's current build system?

GuillaumeGomez:
It's exactly what we're planning to do. :)

PR #319, opened 2023-08-18
Rustify prepare.sh command

GuillaumeGomez:
Part of https://github.com/rust-lang/rustc_codegen_gcc/issues/315.


y.sh
```diff
@@ -0,0 +1,7 @@
+#!/usr/bin/env bash
+
+echo "[BUILD] build system" 1>&2
+mkdir build_system/target
```
bjorn3, line 4: This will error if the target already exists. Maybe use `-p`?

build_system/main.rs
```diff
@@ -0,0 +1,53 @@
...
+    };
+
+    if let Err(e) = match command {
+        Command::Prepare => prepare::run(),
+        Command::Build => build::run(),
+    } {
+        eprintln!("Command failed to run: {e:?}");
+        process::exit(1);
+    }
+}
```
bjorn3, line 53: nit: missing trailing newline

This will result in a spurious extra line of changes in diffs when you add something after this line.

y.sh
GuillaumeGomez: Just did. :)

build_system/main.rs
GuillaumeGomez: Good catch, thanks! Not used to not have rustfmt having my back...

build_system/src/prepare.rs
```diff
@@ -0,0 +1,113 @@
...
+    run_command(&["git", "init"], Some(&dst_dir))?;
+    println!("[GIT] add (cwd): `{}`", dst_dir.display());
+    run_command(&["git", "add", "."], Some(&dst_dir))?;
+    println!("[GIT] commit (cwd): `{}`", dst_dir.display());
+
+    // This is needed on systems where nothing is configured.
+    // git really needs something here, or it will fail.
+    // Even using --author is not enough.
+    run_command(&["git", "config", "user.email", "none@example.com"], Some(&dst_dir))?;
+    run_command(&["git", "config", "user.name", "None"], Some(&dst_dir))?;
```
bjorn3, line 46: `core.autocrlf=false` and `commit.gpgSign=false` may also be necessary depending on existing configuration the user has. This is what I use to apply these to all git commands: https://github.com/bjorn3/rustc_codegen_cranelift/blob/9aa5fbf79c39867f144e5a38c53a8edd2e2069ba/build_system/utils.rs#L170-L187

build_system/src/prepare.rs
```diff
@@ -0,0 +1,113 @@
...
+}
+
+// build with cg_llvm for perf comparison
+fn build_raytracer(repo_dir: &Path) -> Result<(), String> {
+    run_command(&["cargo", "build"], Some(repo_dir))?;
+    run_command(&["mv", "taret/debug/main", "raytracer_cg_llvm"], Some(repo_dir))?;
+    Ok(())
+}
+
+fn clone_and_setup<F>(repo_url: &str, checkout_commit: &str, extra: Option<F>) -> Result<(), String>
```
bjorn3, line 71: The `GitRepo` struct of cg_clif handles using the github zip archives for much faster downloads when curl is available as well as applying patches and writing a lockfile for reproducibility: https://github.com/bjorn3/rustc_codegen_cranelift/blob/9aa5fbf79c39867f144e5a38c53a8edd2e2069ba/build_system/prepare.rs#L96 For an initial rewrite the current code is fine too.

GuillaumeGomez: Might be nice to have later on indeed.

build_system/src/prepare.rs
GuillaumeGomez: Thanks! Adding that.

PR #326, opened 2023-09-04
Add support for NonNull function attribute

GuillaumeGomez:
Based on https://github.com/antoyo/gcc/pull/28 and on https://github.com/antoyo/gccjit.rs/pull/23.

### DIFF
--- a/src/abi.rs
+++ b/src/abi.rs
@@ -96,14 +96,22 @@ impl GccType for Reg {
     }
 }
 
+pub struct FnAbiGcc<'gcc> {
+    pub return_type: Type<'gcc>,
+    pub arguments_type: Vec<Type<'gcc>>,
+    pub is_c_variadic: bool,
+    pub on_stack_param_indices: FxHashSet<usize>,
+    pub fn_attributes: Vec<FnAttribute<'gcc>>,
+}
+
 pub trait FnAbiGccExt<'gcc, 'tcx> {
     // TODO(antoyo): return a function pointer type instead?
-    fn gcc_type(&self, cx: &CodegenCx<'gcc, 'tcx>) -> (Type<'gcc>, Vec<Type<'gcc>>, bool, FxHashSet<usize>, Vec<FnAttribute<'gcc>>);
+    fn gcc_type(&self, cx: &CodegenCx<'gcc, 'tcx>) -> FnAbiGcc<'gcc>;
     fn ptr_to_gcc_type(&self, cx: &CodegenCx<'gcc, 'tcx>) -> Type<'gcc>;
 }
 
 impl<'gcc, 'tcx> FnAbiGccExt<'gcc, 'tcx> for FnAbi<'tcx, Ty<'tcx>> {
-    fn gcc_type(&self, cx: &CodegenCx<'gcc, 'tcx>) -> (Type<'gcc>, Vec<Type<'gcc>>, bool, FxHashSet<usize>, Vec<FnAttribute<'gcc>>) {
+    fn gcc_type(&self, cx: &CodegenCx<'gcc, 'tcx>) -> FnAbiGcc<'gcc> {
         let mut on_stack_param_indices = FxHashSet::default();
 
         // This capacity calculation is approximate.
@@ -111,7 +119,7 @@ impl<'gcc, 'tcx> FnAbiGccExt<'gcc, 'tcx> for FnAbi<'tcx, Ty<'tcx>> {
             self.args.len() + if let PassMode::Indirect { .. } = self.ret.mode { 1 } else { 0 }
         );
 
-        let return_ty =
+        let return_type =
             match self.ret.mode {
                 PassMode::Ignore => cx.type_void(),
                 PassMode::Direct(_) | PassMode::Pair(..) => self.ret.layout.immediate_gcc_type(cx),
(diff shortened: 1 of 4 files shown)


src/abi.rs
```diff
@@ -121,28 +121,33 @@ impl<'gcc, 'tcx> FnAbiGccExt<'gcc, 'tcx> for FnAbi<'tcx, Ty<'tcx>> {
...
         #[cfg(feature = "master")]
-        let apply_attrs = |ty: Type<'gcc>, attrs: &ArgAttributes| {
-            if cx.sess().opts.optimize != config::OptLevel::No
-                && attrs.regular.contains(rustc_target::abi::call::ArgAttribute::NoAlias)
-            {
-                ty.make_restrict()
-            } else {
-                ty
+        let mut apply_attrs = |mut ty: Type<'gcc>, attrs: &ArgAttributes, arg_index: usize| {
+            if cx.sess().opts.optimize == config::OptLevel::No {
```
antoyo, line 128: Just to make sure, no attributes are added when optimizations are disabled?

src/abi.rs
```diff
@@ -151,30 +156,41 @@ impl<'gcc, 'tcx> FnAbiGccExt<'gcc, 'tcx> for FnAbi<'tcx, Ty<'tcx>> {
                         argument_tys.push(Reg::i32().gcc_type(cx));
                     }
                     let ty = cast.gcc_type(cx);
-                    apply_attrs(ty, &cast.attrs)
+                    apply_attrs(ty, &cast.attrs, argument_tys.len())
```
antoyo, line 159: Why do you pass the length, here?

src/abi.rs
```diff
@@ -98,12 +98,12 @@ impl GccType for Reg {
...
 pub trait FnAbiGccExt<'gcc, 'tcx> {
     // TODO(antoyo): return a function pointer type instead?
-    fn gcc_type(&self, cx: &CodegenCx<'gcc, 'tcx>) -> (Type<'gcc>, Vec<Type<'gcc>>, bool, FxHashSet<usize>);
+    fn gcc_type(&self, cx: &CodegenCx<'gcc, 'tcx>) -> (Type<'gcc>, Vec<Type<'gcc>>, bool, FxHashSet<usize>, Vec<FnAttribute<'gcc>>);
     fn ptr_to_gcc_type(&self, cx: &CodegenCx<'gcc, 'tcx>) -> Type<'gcc>;
 }
 
 impl<'gcc, 'tcx> FnAbiGccExt<'gcc, 'tcx> for FnAbi<'tcx, Ty<'tcx>> {
-    fn gcc_type(&self, cx: &CodegenCx<'gcc, 'tcx>) -> (Type<'gcc>, Vec<Type<'gcc>>, bool, FxHashSet<usize>) {
+    fn gcc_type(&self, cx: &CodegenCx<'gcc, 'tcx>) -> (Type<'gcc>, Vec<Type<'gcc>>, bool, FxHashSet<usize>, Vec<FnAttribute<'gcc>>) {
```
antoyo, line 106: This tuple is getting large. Could you please replace with a struct?

src/abi.rs
```diff
@@ -151,30 +156,41 @@ impl<'gcc, 'tcx> FnAbiGccExt<'gcc, 'tcx> for FnAbi<'tcx, Ty<'tcx>> {
...
+        };
+        #[cfg(not(feature = "master"))]
+        let fn_attrs = Vec::new();
+
+        (return_ty, argument_tys, self.c_variadic, on_stack_param_indices, fn_attrs)
     }
 
     fn ptr_to_gcc_type(&self, cx: &CodegenCx<'gcc, 'tcx>) -> Type<'gcc> {
-        let (return_type, params, variadic, on_stack_param_indices) = self.gcc_type(cx);
+        // FIXME: Should we do something with `fn_attrs`?
```
antoyo, line 192: I have the feeling that we should, since I handled the `on stack` parameters below.
But I'm not sure how to double-check this.

src/abi.rs
GuillaumeGomez: I confirm.

src/abi.rs
GuillaumeGomez: Because we fill `argument_tys` with each argument. So the length is the index of the argument. I'm using the same logic in all `apply_attrs` calls.

src/abi.rs
GuillaumeGomez: Good idea!

src/declare.rs
```diff
@@ -80,9 +80,13 @@ impl<'gcc, 'tcx> CodegenCx<'gcc, 'tcx> {
     }
 
     pub fn declare_fn(&self, name: &str, fn_abi: &FnAbi<'tcx, Ty<'tcx>>) -> Function<'gcc> {
-        let (return_type, params, variadic, on_stack_param_indices) = fn_abi.gcc_type(self);
+        let (return_type, params, variadic, on_stack_param_indices, fn_attrs) = fn_abi.gcc_type(self);
         let func = declare_raw_fn(self, name, () /*fn_abi.llvm_cconv()*/, return_type, &params, variadic);
         self.on_stack_function_params.borrow_mut().insert(func, on_stack_param_indices);
+        // We need to handle `nonnull` here where we still have access to function args.
```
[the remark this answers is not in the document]
GuillaumeGomez: Literally the function arguments?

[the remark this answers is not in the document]
antoyo: Sure, but we also had access to the arguments earlier, so I'm not sure what this comment brings. Plus, I'm not sure why we would need the arguments since the attributes are already computed.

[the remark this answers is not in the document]
GuillaumeGomez: Comment might need improvement indeed. I was trying to explanation that we now have the "computed" attributes of the function and that `fn_attributes` is only "available" at this point and if we don't add them to `func`, it will be lost afterward.

src/abi.rs
```diff
@@ -151,31 +164,54 @@ impl<'gcc, 'tcx> FnAbiGccExt<'gcc, 'tcx> for FnAbi<'tcx, Ty<'tcx>> {
...
+            is_c_variadic: self.c_variadic,
+            on_stack_param_indices,
+            fn_attributes: fn_attrs,
+        }
     }
 
     fn ptr_to_gcc_type(&self, cx: &CodegenCx<'gcc, 'tcx>) -> Type<'gcc> {
-        let (return_type, params, variadic, on_stack_param_indices) = self.gcc_type(cx);
-        let pointer_type = cx.context.new_function_pointer_type(None, return_type, &params, variadic);
+        // FIXME: Should we do something with `fn_attrs`?
```
antoyo, line 206:
```suggestion
        // FIXME(antoyo): Should we do something with `fn_attrs`?
```
as per the rustc's style.

src/type_of.rs
```diff
@@ -372,7 +372,13 @@ impl<'gcc, 'tcx> LayoutTypeMethods<'tcx> for CodegenCx<'gcc, 'tcx> {
     }
 
     fn fn_decl_backend_type(&self, fn_abi: &FnAbi<'tcx, Ty<'tcx>>) -> Type<'gcc> {
-        let (return_type, param_types, variadic, _) = fn_abi.gcc_type(self);
-        self.context.new_function_pointer_type(None, return_type, &param_types, variadic)
+        // // FIXME: Should we do something with `fn_attrs`?
```
antoyo, line 375:
```suggestion
        // FIXME(antoyo): Should we do something with `fn_attrs`?
```

PR #344, opened 2023-09-25
Improve y.sh script (to potentially allow dependencies) and improve `build_system` help message

GuillaumeGomez:


y.sh
```diff
@@ -2,6 +2,7 @@
 
 set -e
 echo "[BUILD] build system" 1>&2
-mkdir -p build_system/target
-rustc build_system/src/main.rs -o build_system/target/y -Cdebuginfo=1 --edition 2021
-exec ./build_system/target/y "$@"
+cd build_system
+cargo build --release
+cd ..
```
bjorn3, line 7: Using pushd and popd would also be an option. Mostly a matter of preference for simple cases like this, but when you have a lot of nested directories or there is a symlink, pushd+popd less error prone.

GuillaumeGomez: In this case I used `cd` because I didn't want extra things to be displayed.

bjorn3: I believe `2>/dev/null` works for that, which makes it less nice than `cd ..` in this case though.

PR #353, opened 2023-10-12
Fix/int types alignment

antoyo:
Based off #352.


src/context.rs
```diff
@@ -265,15 +271,16 @@ impl<'gcc, 'tcx> CodegenCx<'gcc, 'tcx> {
     }
 
     pub fn is_native_int_type(&self, typ: Type<'gcc>) -> bool {
+        // TODO: cache those types to not query libgccjit everytime this is called.
```
antoyo, line 274: Don't forget this.

PR #354, opened 2023-10-18
optimize u128/i128 popcounts further

sadlerap:
Don't fall back on breaking apart the popcount operation if 128-bit integers are natively supported.

### DIFF
--- a/src/intrinsic/mod.rs
+++ b/src/intrinsic/mod.rs
@@ -4,7 +4,7 @@ mod simd;
 #[cfg(feature="master")]
 use std::iter;
 
-use gccjit::{BinaryOp, ComparisonOp, Function, RValue, ToRValue, Type, UnaryOp, FunctionType};
+use gccjit::{ComparisonOp, Function, RValue, ToRValue, Type, UnaryOp, FunctionType};
 use rustc_codegen_ssa::MemFlags;
 use rustc_codegen_ssa::base::wants_msvc_seh;
 use rustc_codegen_ssa::common::IntPredicate;
@@ -819,7 +819,9 @@ impl<'a, 'gcc, 'tcx> Builder<'a, 'gcc, 'tcx> {
                 value
             };
 
-        if value_type.is_u128(&self.cx) {
+        // only break apart 128-bit ints if they're not natively supported
+        // TODO(antoyo): remove this if/when native 128-bit integers land in libgccjit
+        if value_type.is_u128(&self.cx) && !self.cx.supports_128bit_integers {
             let sixty_four = self.gcc_int(value_type, 64);
             let right_shift = self.gcc_lshr(value, sixty_four);
             let high = self.gcc_int_cast(right_shift, self.cx.ulonglong_type);
@@ -842,30 +844,33 @@ impl<'a, 'gcc, 'tcx> Builder<'a, 'gcc, 'tcx> {
         let counter_type = self.int_type;
         let counter = self.current_func().new_local(None, counter_type, "popcount_counter");
         let val = self.current_func().new_local(None, value_type, "popcount_value");
-        let zero = self.context.new_rvalue_zero(counter_type);
+        let zero = self.gcc_zero(counter_type);
         self.llbb().add_assignment(None, counter, zero);
         self.llbb().add_assignment(None, val, value);
         self.br(loop_head);
 
         // check if value isn't zero
         self.switch_to_block(loop_head);
-        let zero = self.context.new_rvalue_zero(value_type);
-        let cond = self.context.new_comparison(None, ComparisonOp::NotEquals, val.to_rvalue(), zero);
+        let zero = self.gcc_zero(value_type);
+        let cond = self.gcc_icmp(IntPredicate::IntNE, val.to_rvalue(), zero);
         self.cond_br(cond, loop_body, loop_tail);
 
         // val &= val - 1;
         self.switch_to_block(loop_body);
-        let sub = val.to_rvalue() - self.context.new_rvalue_one(value_type);
-        loop_body.add_assignment_op(None, val, BinaryOp::BitwiseAnd, sub);
+        let one = self.gcc_int(value_type, 1);
+        let sub = self.gcc_sub(val.to_rvalue(), one);
+        let op = self.gcc_and(val.to_rvalue(), sub);
+        loop_body.add_assignment(None, val, op);
 
         // counter += 1
-        let one = self.context.new_rvalue_one(counter_type);
-        loop_body.add_assignment_op(None, counter, BinaryOp::Plus, one);
+        let one = self.gcc_int(counter_type, 1);
+        let op = self.gcc_add(counter.to_rvalue(), one);
+        loop_body.add_assignment(None, counter, op);
         self.br(loop_head);
 
         // end of loop
         self.switch_to_block(loop_tail);
-        self.context.new_cast(None, counter.to_rvalue(), result_type)
+        self.gcc_int_cast(counter.to_rvalue(), result_type)
     }
 
     // Algorithm from: https://blog.regehr.org/archives/1063


antoyo:
Thanks!

Do you think it would be worth trying to make it work even for non-native integers by using the operations from the `int` module as I suggested earlier?

sadlerap:
I'll give it a try - if it ends up working, I'll replace what I've posted here with whatever works.  If not, I'll let you know what fails.

sadlerap:
I think I got a bug in libgccjit.

When I compile with [this patch][1], I get the following error:

<p>

```
rustc_codegen_gcc is build for rustc 1.75.0-nightly (97c81e1b5 2023-10-07) but the default rustc version is rustc 1.75.0-nightly (49691b1f7 2023-10-16).
Using rustc 1.75.0-nightly (97c81e1b5 2023-10-07).
   Compiling minmax_test v0.1.0 (/home/andy/repos/minmax_test)
libgccjit.so: warning: command-line option ‘-x86-asm-syntax=intel’ is valid for Modula-2 but not for
error: rustc interrupted by SIGSEGV, printing backtrace

/home/andy/.rustup/toolchains/nightly-2023-10-08-x86_64-unknown-linux-gnu/lib/librustc_driver-3558b9258144fc36.so(+0x2aff526)[0x7f00466ff526]
/usr/lib/libc.so.6(+0x3e710)[0x7f00438c6710]
/home/andy/repos/gcc/build/gcc/libgccjit.so.0(+0x592255)[0x7f0038392255]
/home/andy/repos/gcc/build/gcc/libgccjit.so.0(+0x592f78)[0x7f0038392f78]
/home/andy/repos/gcc/build/gcc/libgccjit.so.0(+0x597d84)[0x7f0038397d84]

### cycle encountered after 5 frames with period 6
/home/andy/repos/gcc/build/gcc/libgccjit.so.0(+0x592780)[0x7f0038392780]
/home/andy/repos/gcc/build/gcc/libgccjit.so.0(+0x59e41d)[0x7f003839e41d]
/home/andy/repos/gcc/build/gcc/libgccjit.so.0(+0x5930a4)[0x7f00383930a4]
/home/andy/repos/gcc/build/gcc/libgccjit.so.0(+0x595f8d)[0x7f0038395f8d]
/home/andy/repos/gcc/build/gcc/libgccjit.so.0(+0x595fc7)[0x7f0038395fc7]
/home/andy/repos/gcc/build/gcc/libgccjit.so.0(+0x597e87)[0x7f0038397e87]
### recursed 41 times

/home/andy/repos/gcc/build/gcc/libgccjit.so.0(+0x592780)[0x7f0038392780]
/home/andy/repos/gcc/build/gcc/libgccjit.so.0(+0x59e41d)[0x7f003839e41d]
/home/andy/repos/gcc/build/gcc/libgccjit.so.0(+0x5930a4)[0x7f00383930a4]
/home/andy/repos/gcc/build/gcc/libgccjit.so.0(+0x595f8d)[0x7f0038395f8d]
/home/andy/repos/gcc/build/gcc/libgccjit.so.0(+0x595fc7)[0x7f0038395fc7]

note: rustc unexpectedly overflowed its stack! this is a bug
note: maximum backtrace depth reached, frames may have been lost
note: we would appreciate a report at https://github.com/rust-lang/rust
note: backtrace dumped due to SIGSEGV! resuming signal
error: could not compile `minmax_test` (lib)

Caused by:
  process didn't exit successfully: `/home/andy/.rustup/toolchains/nightly-2023-10-08-x86_64-unknown-linux-gnu/bin/rustc --crate-name minmax_test --edition=2021 src/lib.rs --error-format=json --json=diagnostic-rendered-ansi,artifacts,future-incompat --diagnostic-width=141 --crate-type lib --emit=dep-info,metadata,link -C opt-level=3 -C embed-bitcode=no --emit asm -Ccodegen-units=1 -C llvm-args=-x86-asm-syntax=intel -Cdebuginfo=2 -C metadata=19b01f69af993630 -C extra-filename=-19b01f69af993630 --out-dir /home/andy/repos/minmax_test/target/release/deps -L dependency=/home/andy/repos/minmax_test/target/release/deps -Csymbol-mangling-version=v0 -Cdebuginfo=2 -Clto=off -Zcodegen-backend=/home/andy/repos/rustc_codegen_gcc/target/release/librustc_codegen_gcc.so --sysroot /home/andy/repos/rustc_codegen_gcc/build_sysroot/sysroot` (signal: 11, SIGSEGV: invalid memory reference)

Cargo failed with exit status: 101
```

</p>
</details> 

[1]: https://github.com/sadlerap/rustc_codegen_gcc/commit/c715c88d06335c88f74b63060aa676766eb0b508

antoyo:
Indeed. Good find!
You can see in the [CI](https://github.com/sadlerap/rustc_codegen_gcc/actions/runs/6555101043/job/17803082877#step:20:4317) that it's because you use an array for a binary operation.
It's in [the call to `add_assignment_op()`](https://github.com/sadlerap/rustc_codegen_gcc/commit/c715c88d06335c88f74b63060aa676766eb0b508#diff-34f8c479ac3b1727ae8bccaf335c8f3be407f4d48dbf725e26b793ebee763dc4R849).
libgccjit doesn't check that the values in this call are numeric.
I'll fix this!
Thanks!

sadlerap:
hmm, I wonder if replacing the assignment ops with just plain assignments would work.  I'll give it a try.

antoyo:
I added the type check in [this commit](https://github.com/antoyo/gcc/commit/2eca081142f9161adda821812f74c6669935fe0a).

Yeah, that might work indeed.

sadlerap:
Sadly, while subbing out assignment ops with pure assignments works around the bug, it also results in poor codegen when 128-bit integers aren't supported and [the fallback][1] is skipped.

[1]: https://github.com/rust-lang/rustc_codegen_gcc/blob/a9f31ab4814a936c91dec8be0da8807c7210b44a/src/intrinsic/mod.rs#L822-L831

<p>


Before:

```asm
popcount_128:

        xor eax, eax
        popcnt rax, rsi
        test rsi, rsi
        cmove rax, rsi

        xor ecx, ecx
        popcnt rcx, rdi
        mov rdx, rax
        add rax, rcx

        test rdi, rdi
        cmove rax, rdx

        ret
```

After:

```asm
popcount_128:

        push r14

        mov rax, rsi
        push r13
        push r12
        push rbp
        push rbx
        sub rsp, 32

        or rax, rdi
        mov QWORD PTR 16[rsp], rdi

        mov QWORD PTR 24[rsp], rsi
        vmovdqa xmm3, XMMWORD PTR 16[rsp]
        vmovdqa XMMWORD PTR [rsp], xmm3
        je .L67
        mov rbp, rsi

        mov rbx, rdi

        xor r12d, r12d

.L66:
        mov r14, QWORD PTR [rsp]
        mov r13, QWORD PTR 8[rsp]
        xor ecx, ecx
        mov edx, 1
        add r12d, 1
        mov rdi, r14
        mov rsi, r13
        call __rust_u128_sub@PLT

        xor ecx, ecx
        mov edx, 1
        mov rdi, r14
        mov rsi, r13
        and rbx, rax
        mov QWORD PTR 16[rsp], rax
        call __rust_u128_sub@PLT

        vmovq xmm0, QWORD PTR 16[rsp]
        mov rax, rbx
        and rbp, rdx
        vpinsrq xmm0, xmm0, rdx, 1
        vpand xmm1, xmm0, XMMWORD PTR [rsp]
        or rax, rbp
        vmovdqa XMMWORD PTR [rsp], xmm1
        jne .L66
        mov eax, r12d
.L63:
        add rsp, 32
        pop rbx

        pop rbp

        pop r12

        pop r13
        pop r14

        ret
.L67:

        xor eax, eax
        jmp .L63
```

</p>
</details> 

However, when 128-bit integers are supported, we do see the improvement as expected.

<p>

Before: 

```asm
popcount_128:

        xor eax, eax
        popcnt rax, rsi
        test rsi, rsi
        cmove rax, rsi

        xor ecx, ecx
        popcnt rcx, rdi
        mov rdx, rax
        add rax, rcx

        test rdi, rdi
        cmove rax, rdx

        ret
```

After:

```asm
popcount_128:

        mov rdx, rdi
        or rdx, rsi

        je .L65

        xor ecx, ecx

        popcnt rsi, rsi

        popcnt rcx, rdi
        lea eax, [rsi+rcx]
        ret
.L65:

        xor eax, eax
        ret
```

</p>
</details> 

I think if we gate the fallback behind native support for 128-bit integers, we'll get the best of both worlds - good-ish codegen when 128-bit integers are supported, and a roughly as-good fallback when they aren't.

antoyo:
Do you know (or did you do some benchmarks to check) if having one jump is indeed faster than the 2 conditional moves?

sadlerap:
I wrote some microbenchmarks with `criterion`, but they seemed to get stuck on the anaylsis step (let it run for about 15 minutes and it still wasn't done, whereas the same code compiled with llvm completed in a few seconds).  I'll try some other options, see if I get more luck there.

sadlerap:
Alright, I think I have some solid numbers.

My benchmarks run popcount 1000 times on the same number and add the results together.  This is done to inflate iteration times a bit - with an operation as small as this, it's hard to get good numbers for how long it takes to complete.

Before:

```
test tests::bench_popcount_u128_0       ... bench:         668 ns/iter (+/- 57)
test tests::bench_popcount_u128_1       ... bench:         680 ns/iter (+/- 9)
test tests::bench_popcount_u128_10      ... bench:         683 ns/iter (+/- 12)
test tests::bench_popcount_u128_minus_1 ... bench:         705 ns/iter (+/- 16)
```

After:

```
test tests::bench_popcount_u128_0       ... bench:         580 ns/iter (+/- 5)
test tests::bench_popcount_u128_1       ... bench:         592 ns/iter (+/- 91)
test tests::bench_popcount_u128_10      ... bench:         595 ns/iter (+/- 186)
test tests::bench_popcount_u128_minus_1 ... bench:         593 ns/iter (+/- 12)
```

Source:

```rust
#![feature(test)]

pub fn popcount_128(a: u128) -> u32 {
    a.count_ones()
}

#[cfg(test)]
pub mod tests {
    use super::*;
    use test::{Bencher, black_box};

    #[bench]
    fn bench_popcount_u128_0(b: &mut Bencher) {
        let x = black_box(0u128);
        b.iter(|| (0..1000).fold(0, |old, _new| old + popcount_128(black_box(x))))
    }

    #[bench]
    fn bench_popcount_u128_1(b: &mut Bencher) {
        let x = black_box(1u128);
        b.iter(|| (0..1000).fold(0, |old, _new| old + popcount_128(black_box(x))))
    }

    #[bench]
    fn bench_popcount_u128_10(b: &mut Bencher) {
        let x = black_box(10u128);
        b.iter(|| (0..1000).fold(0, |old, _new| old + popcount_128(black_box(x))))
    }

    #[bench]
    fn bench_popcount_u128_minus_1(b: &mut Bencher) {
        let x = black_box(-1i128 as u128);
        b.iter(|| (0..1000).fold(0, |old, _new| old + popcount_128(black_box(x))))
    }

}
```

antoyo:
Thanks a lot for your work!

PR #355, opened 2023-10-18
Refactor CI

antoyo:


.github/workflows/failures.yml
```diff
@@ -0,0 +1,120 @@
...
+    steps:
+    - uses: actions/checkout@v3
+
+    - uses: actions/checkout@v3
+      with:
+        repository: llvm/llvm-project
+        path: llvm
+
+    - name: Install packages
+      # `llvm-14-tools` is needed to install the `FileCheck` binary which is used for asm tests.
```
antoyo, line 34: This package is probably not needed here.

.github/workflows/release.yml
```diff
@@ -19,8 +19,10 @@ jobs:
       fail-fast: false
       matrix:
         commands: [
-          "--test-successful-rustc --nb-parts 2 --current-part 0",
-          "--test-successful-rustc --nb-parts 2 --current-part 1",
+          "--test-successful-rustc --nb-parts 4 --current-part 0",
+          "--test-successful-rustc --nb-parts 4 --current-part 1",
+          "--test-successful-rustc --nb-parts 4 --current-part 2",
+          "--test-successful-rustc --nb-parts 4 --current-part 3",
```
antoyo, line 25: Try with more parts as this still seems much slower than the tests without LTO.

PR #374, opened 2023-10-27
Do not emit .eh_frame section if using -Cpanic=abort

antoyo:


bjorn3:
`.eh_frame` is necessary once exceptions reaching into panic=abort rust code starts to abort the program rather than being UB.

antoyo:
@bjorn3: This is necessary for Rust for Linux.
How do you suggest we handle this?

bjorn3:
It may be that this change is fine and once unwinding into `panic=abort` rust code will abort the presence of the landingpad will cause `.eh_frame` to be emitted by both cg_llvm and cg_gcc. When that happens whatever solution is chosen for cg_llvm may work for cg_gcc too.

build_sysroot/build_sysroot.sh
```diff
@@ -27,4 +27,10 @@ fi
 
 # Copy files to sysroot
 mkdir -p sysroot/lib/rustlib/$TARGET_TRIPLE/lib/
+# TODO: copy src if asked by the user.
```
antoyo, line 30: Don't forget this.

antoyo:
@bjorn3: Do you think this should instead be set/unset depending on the value of `-Cforce-unwind-tables`?

PR #377, opened 2023-11-05
Feature/comment section

antoyo:


src/lib.rs
```diff
@@ -244,17 +245,29 @@ impl CodegenBackend for GccCodegenBackend {
...
 }
 
+fn new_context<'gcc, 'tcx>(tcx: &TyCtxt<'tcx>) -> Context<'gcc> {
+    let context = Context::default();
+    if tcx.sess.target.arch == "x86" || tcx.sess.target.arch == "x86_64" {
+        context.add_command_line_option("-masm=intel");
+    }
+    let version = Version::get();
+    let version = format!("{}.{}.{}", version.major, version.minor, version.patch);
+    context.set_output_ident(&format!("rustc version {} with libgccjit {}",
```
bjorn3, line 255: This is what cg_clif uses: https://github.com/rust-lang/rustc_codegen_cranelift/blob/ef3703694ff6b19d38aff83477bb2ad38f5a58d8/src/debuginfo/mod.rs#L22-L26 This includes the full `rustc -V` string including commit and date, not just the version number (which is identical for all nightlies released over the course of 6 weeks)

PR #382, opened 2023-11-10
simd: implement missing intrinsics from simd/generic-arithmetic-pass.rs

sadlerap:
This implements/fixes the following intrinsics:

- `simd_bswap`
- `simd_frem`
- `simd_bitreverse`
- `simd_ctlz`
- `simd_cttz`
- `simd_neg`

These were the remaining intrinsics from the `simd/generic-arithmetic-pass.rs` ui test, which should now pass for a patched libgccjit.  However, [it seems not to compile against a non-patched libgccjit][1], so leave it disabled there for now.

[1]: https://github.com/sadlerap/rustc_codegen_gcc/actions/runs/6820180639/job/18548672444#step:15:4375

### DIFF
--- a/src/intrinsic/simd.rs
+++ b/src/intrinsic/simd.rs
@@ -1,3 +1,5 @@
+use std::iter::FromIterator;
+
 use gccjit::ToRValue;
 use gccjit::{BinaryOp, RValue, Type};
 #[cfg(feature = "master")]
@@ -154,6 +156,58 @@ pub fn generic_simd_intrinsic<'a, 'gcc, 'tcx>(
         // FIXME(antoyo): allow comparing vector types as equal in libgccjit.
         let arg2 = bx.context.new_bitcast(None, args[1].immediate(), arg1.get_type());
         return Ok(compare_simd_types(bx, arg1, arg2, in_elem, llret_ty, cmp_op));
+    }
+
+    if name == sym::simd_bswap {
+        let vector = args[0].immediate();
+        let ret = match in_elem.kind() {
+            ty::Int(i) if i.bit_width() == Some(8) => vector,
+            ty::Uint(i) if i.bit_width() == Some(8) => vector,
+            ty::Int(_) | ty::Uint(_) => {
+                let v_type = vector.get_type();
+                let vector_type = v_type.unqualified().dyncast_vector().expect("vector type");
+                let elem_type = vector_type.get_element_type();
+                let elem_size_bytes = elem_type.get_size();
+                let type_size_bytes = elem_size_bytes as u64 * in_len;
+
+                let shuffle_indices = Vec::from_iter(0..type_size_bytes);
+                let byte_vector_type = bx.context.new_vector_type(bx.type_u8(), type_size_bytes);
+                let byte_vector = bx.context.new_bitcast(None, args[0].immediate(), byte_vector_type);
+
+                #[cfg(not(feature = "master"))]
+                let shuffled = {
+                    let new_elements: Vec<_> = shuffle_indices.chunks_exact(elem_size_bytes as _)
+                        .flat_map(|x| x.iter().rev())
+                        .map(|&i| {
+                            let index = bx.context.new_rvalue_from_long(bx.u64_type, i as _);
+                            bx.extract_element(byte_vector, index)
+                        })
+                        .collect();
+
+                    bx.context.new_rvalue_from_vector(None, byte_vector_type, &new_elements)
+                };
+                #[cfg(feature = "master")]
+                let shuffled = {
+                    let indices: Vec<_> = shuffle_indices.chunks_exact(elem_size_bytes as _)
+                        .flat_map(|x| x.iter().rev())
+                        .map(|&i| bx.context.new_rvalue_from_int(bx.u8_type, i as _))
+                        .collect();
+
+                    let mask = bx.context.new_rvalue_from_vector(None, byte_vector_type, &indices);
+                    bx.context.new_rvalue_vector_perm(None, byte_vector, byte_vector, mask)
+                };
+                bx.context.new_bitcast(None, shuffled, v_type)
+            }
+            _ => {
+                return_error!(InvalidMonomorphization::UnsupportedOperation {
+                    span,
+                    name,
+                    in_ty,
+                    in_elem,
+                });
+            }
+        };
+        return Ok(ret);
     }
 
     if name == sym::simd_shuffle {
--- a/src/intrinsic/simd.rs
+++ b/src/intrinsic/simd.rs
@@ -333,6 +333,22 @@ pub fn generic_simd_intrinsic<'a, 'gcc, 'tcx>(
         return Ok(bx.context.new_rvalue_from_vector(None, vector_ty, &new_elements));
     }
 
+    if name == sym::simd_ctlz || name == sym::simd_cttz {
+        let vector = args[0].immediate();
+        let elements: Vec<_> = (0..in_len)
+            .map(|i| {
+                let index = bx.context.new_rvalue_from_long(bx.i32_type, i as i64);
+                let value = bx.extract_element(vector, index).to_rvalue();
+                if name == sym::simd_ctlz {
+                    bx.count_leading_zeroes(value.get_type().get_size() as u64 * 8, value)
+                } else {
+                    bx.count_trailing_zeroes(value.get_type().get_size() as u64 * 8, value)
+                }
+            })
+            .collect();
+        return Ok(bx.context.new_rvalue_from_vector(None, vector.get_type(), &elements));
+    }
+
     if name == sym::simd_shuffle {
         // Make sure this is actually an array, since typeck only checks the length-suffixed
         // version of this intrinsic.
--- a/src/intrinsic/simd.rs
+++ b/src/intrinsic/simd.rs
@@ -23,6 +23,8 @@ use rustc_target::abi::Align;
 use crate::builder::Builder;
 #[cfg(feature = "master")]
 use crate::context::CodegenCx;
+#[cfg(not(feature = "master"))]
+use crate::common::SignType;
 
 pub fn generic_simd_intrinsic<'a, 'gcc, 'tcx>(
     bx: &mut Builder<'a, 'gcc, 'tcx>,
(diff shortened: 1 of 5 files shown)


src/intrinsic/simd.rs
```diff
@@ -156,6 +160,181 @@ pub fn generic_simd_intrinsic<'a, 'gcc, 'tcx>(
...
+        if elem_size_bytes == 1 {
+            return vector;
+        }
+
+        let type_size_bytes = elem_size_bytes as u64 * in_len;
+
+        let shuffle_indices = Vec::from_iter(0..type_size_bytes);
+        let byte_vector_type = bx.context.new_vector_type(bx.type_u8(), type_size_bytes);
+
+        // FIXME: gccjit doesn't allow casting vectors from one type to another of the same
```
antoyo, line 177: Using a `bitcast` should work.

src/intrinsic/simd.rs
```diff
@@ -156,6 +160,181 @@ pub fn generic_simd_intrinsic<'a, 'gcc, 'tcx>(
...
+        let lo_nibble = bx.context.new_rvalue_from_vector(None, long_byte_vector_type, &lo_nibble_elements);
+
+        let mask = bx.context.new_rvalue_from_vector(
+            None,
+            long_byte_vector_type,
+            &vec![bx.gcc_int(bx.u8_type, 0x0f); byte_vector_type_size as _]);
+
+        let swapped = simd_bswap(bx, args[0].immediate());
+
+        // See bswap above for why we have to make a round-trip through the stack.
```
antoyo, line 250: Same.

src/intrinsic/simd.rs
sadlerap: D'oh, don't know why I didn't try a bitcast.  Fixed in 1d88c5b.

src/intrinsic/simd.rs
sadlerap: Fixed in 1d88c5b.

antoyo:
I'll do the review soon, hopefully in the next few days.
I've been quite busy lately.

sadlerap:
No worries, take all the time you need.

src/intrinsic/simd.rs
```diff
@@ -156,6 +160,163 @@ pub fn generic_simd_intrinsic<'a, 'gcc, 'tcx>(
...
+        let byte_vector_type = bx.context.new_vector_type(bx.type_u8(), type_size_bytes);
+        let byte_vector = bx.context.new_bitcast(None, args[0].immediate(), byte_vector_type);
+
+        let new_elements: Vec<_> = shuffle_indices.chunks_exact(elem_size_bytes as _)
+            .flat_map(|x| x.iter().rev())
+            .map(|&i| {
+                let index = bx.context.new_rvalue_from_long(bx.type_u64(), i as _);
+                bx.extract_element(byte_vector, index)
+            })
+            .collect();
```
antoyo, line 184: It might be more efficient to implement this by calling `self.context.new_rvalue_vector_perm()` (see `shuffle_vector`).
What do you think?

src/intrinsic/simd.rs
```diff
@@ -156,6 +160,163 @@ pub fn generic_simd_intrinsic<'a, 'gcc, 'tcx>(
...
+    #[cfg(feature = "master")]
+    if name == sym::simd_bitreverse {
+        let vector = args[0].immediate();
+        let v_type = vector.get_type();
+        let vector_type = v_type.unqualified().dyncast_vector().expect("vector type");
+        let elem_type = vector_type.get_element_type();
+        let elem_size_bytes = elem_type.get_size();
+
+        let type_size_bytes = elem_size_bytes as u64 * in_len;
+        let byte_vector_type_size = type_size_bytes.max(16);
```
antoyo, line 219: Could you please add a comment to explain this line?

src/intrinsic/simd.rs
```diff
@@ -156,6 +160,163 @@ pub fn generic_simd_intrinsic<'a, 'gcc, 'tcx>(
...
+        // same, since that's what new_rvalue_vector_perm expects.
+        let byte_vector = if byte_vector_type_size > type_size_bytes {
+            let mut byte_vector_elements = Vec::with_capacity(byte_vector_type_size as _);
+            for i in 0..type_size_bytes {
+                let idx = bx.gcc_int(bx.u32_type, i as _);
+                let val = bx.extract_element(byte_vector, idx);
+                byte_vector_elements.push(val);
+            }
+            for _ in type_size_bytes..byte_vector_type_size {
+                let val = bx.gcc_zero(bx.u8_type);
```
antoyo, line 253: It might be more efficient to that this `val` defined outside the loop.

src/intrinsic/simd.rs
```diff
@@ -156,6 +160,163 @@ pub fn generic_simd_intrinsic<'a, 'gcc, 'tcx>(
...
+        } else {
+            bx.context.new_bitcast(None, byte_vector, long_byte_vector_type)
+        };
+
+        let four = bx.gcc_int(bx.u8_type, 4);
+        let four_vec = bx.context.new_rvalue_from_vector(
+            None,
+            long_byte_vector_type,
+            &vec![four; byte_vector_type_size as _]);
+        let masked_hi = bx.gcc_lshr(byte_vector, four_vec) & mask;
```
antoyo, line 266: Would using `>>` directly work?

src/intrinsic/simd.rs
```diff
@@ -156,6 +160,163 @@ pub fn generic_simd_intrinsic<'a, 'gcc, 'tcx>(
...
+        let masked_lo = byte_vector & mask;
+
+        let hi = bx.context.new_rvalue_vector_perm(None, hi_nibble, hi_nibble, masked_lo);
+        let lo = bx.context.new_rvalue_vector_perm(None, lo_nibble, lo_nibble, masked_hi);
+
+        let shuffled = hi | lo;
+        let loaded = bx.context.new_bitcast(None, shuffled, v_type);
+        let elems: Vec<_> = (0..in_len)
+            .map(|i| {
+                let idx = bx.gcc_int(bx.u32_type, i as _);
```
antoyo, line 276: When it is known to be native types, `context.new_rvalue_from_int()` is preferred.

src/intrinsic/simd.rs
sadlerap: I'll keep the current implementation around behind a `#[cfg(not(feature = "master"))`, since I'm hoping to keep as much compatibility with non-`master` builds as possible.

See b8bba37 and 0cc4c03.

src/intrinsic/simd.rs
sadlerap: Fixed, see 0cc4c03.

src/intrinsic/simd.rs
sadlerap: Fixed, see 0cc4c03.

src/intrinsic/simd.rs
sadlerap: Do you mean this?

```suggestion
        let masked_hi = (byte_vector >> four_vec) & mask;
```

If so, I've added this in 0cc4c03, and it seems to work.

src/intrinsic/simd.rs
sadlerap: Fixed in 0cc4c03.

src/intrinsic/simd.rs
```diff
@@ -156,6 +160,180 @@ pub fn generic_simd_intrinsic<'a, 'gcc, 'tcx>(
...
+        let masked_hi = (byte_vector >> four_vec) & mask;
+        let masked_lo = byte_vector & mask;
+
+        let hi = bx.context.new_rvalue_vector_perm(None, hi_nibble, hi_nibble, masked_lo);
+        let lo = bx.context.new_rvalue_vector_perm(None, lo_nibble, lo_nibble, masked_hi);
+
+        let shuffled = hi | lo;
+        let cast_ty = bx.context.new_vector_type(elem_type, byte_vector_type_size / (elem_size_bytes as u64));
+        let loaded = bx.context.new_bitcast(None, shuffled, cast_ty);
+        let elems: Vec<_> = (0..in_len)
```
antoyo, line 291: Do you extract the individual elements instead of casting because you want to truncate the vector?
Please add a comment to explain.

src/intrinsic/simd.rs
```diff
@@ -156,6 +160,180 @@ pub fn generic_simd_intrinsic<'a, 'gcc, 'tcx>(
...
+    // component-wise bitreverses if they're not available.
+    #[cfg(not(feature = "master"))]
+    if name == sym::simd_bitreverse {
+        let vector = args[0].immediate();
+        let vector_ty = vector.get_type();
+        let vector_type = vector_ty.unqualified().dyncast_vector().expect("vector type");
+        let num_elements = vector_type.get_num_units();
+
+        let elem_type = vector_type.get_element_type();
+        let elem_size_bytes: u32 = elem_type.get_size();
```
antoyo, line 309: Is there any reason to have a type annotation here?

src/intrinsic/simd.rs
```diff
@@ -156,6 +160,180 @@ pub fn generic_simd_intrinsic<'a, 'gcc, 'tcx>(
...
+            })
+            .collect();
+        return Ok(bx.context.new_rvalue_from_vector(None, vector_ty, &new_elements));
+    }
+
+    if name == sym::simd_ctlz || name == sym::simd_cttz {
+        let vector = args[0].immediate();
+        let elements: Vec<_> = (0..in_len)
+            .map(|i| {
+                let index = bx.gcc_int(bx.i32_type, i as i64);
```
antoyo, line 325: Since `i32` is always native, I would use `new_rvalue_from_long()`.

antoyo (commented):
A few nitpicks.

src/intrinsic/simd.rs
sadlerap: Fixed in b993b22.

src/intrinsic/simd.rs
sadlerap: Nope; that shouldn't be there.  Fixed in 4c9fe11.

src/intrinsic/simd.rs
sadlerap: Yeah, the intention here is to truncate the vector.  Is there a more idiomatic way to do this?

sadlerap:
Apologies on the delay on getting back to your review - last week's holidays made things busy for me, and I couldn't get back to this until now.

src/intrinsic/simd.rs
sadlerap: I've added some more documentation for how this particular algorithm is supposed to work in 8d42a82.

antoyo:

PR #384, opened 2023-11-16
Rustify test.sh

GuillaumeGomez:


GuillaumeGomez:
Note for myself: show the stdout/stderr of a command if it failed.

build_system/src/build.rs
```diff
@@ -138,72 +115,101 @@ fn build_sysroot(
...
+    let mut rustflags = env.get("RUSTFLAGS").cloned().unwrap_or_default();
+    if sysroot_panic_abort {
         rustflags.push_str(" -Cpanic=abort -Zpanic-abort-tests");
     }
-    env.insert(
-        "RUSTFLAGS".to_string(),
-        format!("{} -Zmir-opt-level=3", rustflags),
-    );
-    let channel = if args.sysroot_release_channel {
+    rustflags.push_str(" -Z force-unstable-if-unmarked");
```
antoyo, line 127: Where does this come from?

build_system/src/build.rs
```diff
@@ -138,72 +115,101 @@ fn build_sysroot(
...
 }
 
-fn build_codegen(args: &BuildArg) -> Result<(), String> {
+pub fn build_sysroot(
+    env: &HashMap<String, String>,
+    sysroot_panic_abort: bool,
+    sysroot_release_channel: bool,
+    config: &ConfigInfo,
+) -> Result<(), String> {
+    build_sysroot_inner(
```
antoyo, line 203: What's the reason for having `build_sysroot_inner` if the only call is here?

build_system/src/config.rs
```diff
@@ -1,149 +1,226 @@
...
+            "--sysroot-panic-abort" => self.sysroot_panic_abort = true,
+            _ => return Ok(false),
         }
+        Ok(true)
     }
 
-    if set_target_triple && !set_target {
-        target = target_triple.clone();
-    }
+    pub fn setup(
```
antoyo, line 47: This is mostly the former `set_config`, right?

build_system/src/build.rs
```diff
@@ -237,18 +244,19 @@ fn build_codegen(args: &BuildArg) -> Result<(), String> {
 
     println!("[BUILD] sysroot");
     build_sysroot(
-        &mut env,
-        args,
-        &config,
+        &env,
+        args.config_info.sysroot_panic_abort,
```
antoyo, line 248: Why not pass `args` directly like before?

build_system/src/config.rs
```diff
@@ -1,149 +1,226 @@
...
-                if let Some(arg) = args.next() {
-                    target = arg;
-                    set_target = true;
-                } else {
-                    return Err(
-                        "Expected a value after `--target`, found nothing".to_string()
-                    );
+            "--out-dir" => match args.next() {
+                Some(arg) if !arg.is_empty() => {
+                    // env.insert("CARGO_TARGET_DIR".to_string(), arg.to_string());
```
antoyo, line 35: Why is this commented?

build_system/src/config.rs
```diff
@@ -1,149 +1,226 @@
...
+        }
+        if self.target_triple.is_empty() {
+            self.target_triple = self.host_triple.clone();
+        }
+
+        let mut linker = None;
+
+        if self.host_triple != self.target_triple {
+            if self.target_triple == "m68k-unknown-linux-gnu" {
+                linker = Some("-Clinker=m68k-unknown-linux-gnu-gcc".to_string());
```
antoyo, line 94: Please do like in config.rs:
```
format!("-Clinker={}-gcc", target_triple)
```
and remove this whole `if` expression.

build_system/src/utils.rs
```diff
@@ -29,22 +29,39 @@ fn check_exit_status(
...
+            .collect::<Vec<_>>()
+            .join(" "),
+        cwd.map(|cwd| format!(" (running in folder `{}`)", cwd.display()))
+            .unwrap_or_default(),
+        exit_status.code()
+    );
+    let input = input.iter().map(|i| i.as_ref()).collect::<Vec<&OsStr>>();
+    eprintln!("Command `{:?}` failed", input);
+    if let Some(output) = output {
+        unsafe {
```
antoyo, line 51: I would rather not use unsafe since this is not performance-sensitive.

build_system/src/utils.rs
```diff
@@ -238,3 +294,56 @@ where
...
+
+pub fn split_args(args: &str) -> Vec<String> {
+    let mut out = Vec::new();
+    let mut start = 0;
+    let mut iter = args.char_indices().peekable();
+
+    while iter.peek().is_some() {
+        while let Some((pos, c)) = iter.next() {
+            if c == ' ' {
+                if pos != 0 {
```
antoyo, line 306: Could we just trim `args` to avoid this condition?

build_system/src/utils.rs
```diff
@@ -238,3 +294,56 @@ where
...
+                    if c == end {
+                        found_end = true;
+                        break;
+                    } else if c == '\\' {
+                        // We skip the escaped character.
+                        iter.next();
+                    }
+                }
+                if !found_end {
+                    out.push(args[start..].to_string());
```
antoyo, line 335: Shouldn't we error-out instead?

antoyo (commented):
Here's the review of everything except one file ;) .

build_system/src/test.rs
```diff
@@ -1,15 +1,1176 @@
...
+        &"--target",
+        &args.config_info.target_triple,
+    ]);
+    if !args.no_default_features {
+        command.extend_from_slice(&[&"--cfg", &"feature=\"master\""]);
+    }
+    run_command_with_env(&command, None, Some(env))?;
+
+    // FIXME: doesn't work on m68k.
+    if args.config_info.host_triple != args.config_info.target_triple {
```
antoyo, line 403: Not sure how you made this pass the CI, because it should be the following:
```suggestion
    if args.config_info.host_triple == args.config_info.target_triple {
```

build_system/src/test.rs
```diff
@@ -1,15 +1,1176 @@
...
+    command.extend_from_slice(&[
+        &"example/subslice-patterns-const-eval.rs",
+        &"--crate-type",
+        &"bin",
+        &"--target",
+        &args.config_info.target_triple,
+    ]);
+    run_command_with_env(&command, None, Some(env))?;
+    run_command_in_vm(
+        &[&Path::new(&args.config_info.cargo_target_dir).join("subslice-patterns-const-eval")],
```
antoyo, line 490: To avoid duplicating this `Path::new(…)`, please put it in a variable.

build_system/src/test.rs
```diff
@@ -1,15 +1,1176 @@
...
+    }
+    command.extend_from_slice(&[
+        &"example/mod_bench.rs",
+        &"--crate-type",
+        &"bin",
+        &"--target",
+        &args.config_info.target_triple,
+    ]);
+    run_command_with_env(&command, None, Some(env))?;
+    // FIXME: the compiled binary is not run. Is it normal?
```
antoyo, line 529: Yeah, I never ran any benchmark in the tests.

build_system/src/test.rs
```diff
@@ -1,15 +1,1176 @@
...
+// COMPILE_MOD_BENCH_LLVM_2="rustc example/mod_bench.rs --crate-type bin -Copt-level=2 -o $cargo_target_dir/mod_bench_llvm_2 -Cpanic=abort"
+// COMPILE_MOD_BENCH_LLVM_3="rustc example/mod_bench.rs --crate-type bin -Copt-level=3 -o $cargo_target_dir/mod_bench_llvm_3 -Cpanic=abort"
+//
+// Use 100 runs, because a single compilations doesn't take more than ~150ms, so it isn't very slow
+// hyperfine --runs ${COMPILE_RUNS:-100} "$COMPILE_MOD_BENCH_INLINE" "$COMPILE_MOD_BENCH_LLVM_0" "$COMPILE_MOD_BENCH_LLVM_1" "$COMPILE_MOD_BENCH_LLVM_2" "$COMPILE_MOD_BENCH_LLVM_3"
+// echo "[BENCH RUN] mod_bench"
+// hyperfine --runs ${RUN_RUNS:-10} $cargo_target_dir/mod_bench{,_inline} $cargo_target_dir/mod_bench_llvm_*
+
+fn extended_rand_tests(env: &Env, args: &TestArg) -> Result<(), String> {
+    if args.no_default_features {
```
antoyo, line 737: Could you make a method `is_gcc_master_branch()` that just returns this field to make it more clear that we don't run this test on non-master branch?

build_system/src/test.rs
```diff
@@ -1,15 +1,1176 @@
...
+        let path_str = file.display().to_string().replace("\\", "/");
+        if !path_str.ends_with(".rs") {
+            return Ok(());
+        } else if should_not_remove_test(&path_str) {
+            return Ok(());
+        } else if should_remove_test(file, &path_str) {
+            return std::fs::remove_file(file)
+                .map_err(|error| format!("Failed to remove `{}`: {:?}", file.display(), error));
+        }
+        let file_content = std::fs::read_to_string(file)
```
antoyo, line 953: This seems pretty inefficient. I was about to ask if the CI is noticeably slower with this, but the [tests are not even ran](https://github.com/rust-lang/rustc_codegen_gcc/actions/runs/6985421387/job/19009663991?pr=384#step:15:1531), so please fix this and then check the performance.

build_system/src/test.rs
```diff
@@ -1,15 +1,1176 @@
...
+                std::fs::remove_dir_all(dir).map_err(|error| {
+                    format!("Failed to remove folder `{}`: {:?}", dir.display(), error)
+                })?;
+            }
+            Ok(())
+        },
+        |_| Ok(()),
+    )?;
+
+    fn dir_handling(dir: &Path) -> Result<(), String> {
```
antoyo, line 940: Could you please add comments to explain what these functions do?

build_system/src/test.rs
```diff
@@ -1,15 +1,1176 @@
...
+                .map_err(|error| format!("Failed to remove `{}`: {:?}", file.display(), error))?;
+        }
+        Ok(())
+    }
+
+    let rust_path = Path::new("rust");
+
+    walk_dir(rust_path.join("tests/ui"), dir_handling, file_handling)?;
+    let file = rust_path.join("tests/ui/consts/const_cmp_type_id.rs");
+    std::fs::remove_file(&file)
```
antoyo, line 966: It might be worth creating a function doing the `remove_file()` and `map_err()`.

build_system/src/test.rs
```diff
@@ -1,15 +1,1176 @@
...
+            // Tests generating errors.
+            "consts/issue-94675.rs",
+            // this test is oom-killed in the CI.
+            "mir/mir_heavy/issue-miri-1910.rs",
+        ]
+        .iter()
+        .any(|to_ignore| path_str.ends_with(to_ignore))
+}
+
+fn test_rustc_inner<F>(env: &Env, args: &TestArg, callback: F) -> Result<(), String>
```
antoyo, line 906: Please give a more meaningful name to `callback`.

build_system/src/test.rs
```diff
@@ -1,15 +1,1176 @@
...
+        .map_err(|error| format!("Failed to remove `{}`: {:?}", file.display(), error))?;
+
+    if !callback()? {
+        // FIXME: create a function "display_if_not_quiet" or something along the line.
+        println!("Keeping all UI tests");
+    }
+
+    let nb_parts = args.nb_parts.unwrap_or(0);
+    if nb_parts > 0 {
+        let current_part = args.current_part.unwrap_or(0);
```
antoyo, line 979: It might be more useful to do `expect()` instead of `unwrap_or(0)` so we would want to know if we forget to specify the part.

build_system/src/test.rs
```diff
@@ -1,15 +1,1176 @@
...
+        files.sort();
+        // We increment the number of tests by one because if this is an odd number, we would skip
+        // one test.
+        let count = files.len() / nb_parts + 1;
+        let start = nb_parts * count;
+        let end = start + count;
+        for (pos, path) in files.iter().enumerate() {
+            if pos >= start && pos <= end {
+                continue;
+            }
```
antoyo, line 1018: I cannot understand this part. Perhaps this code is wrong and that's why the tests are not ran in the CI?

build_system/src/test.rs
```diff
@@ -1,15 +1,1176 @@
...
+}
+
+// TODO(GuillaumeGomez): when rewriting in Rust, refactor with the code in tests/lang_tests_common.rs if possible.
+fn run_command_in_vm(
+    command: &[&dyn AsRef<OsStr>],
+    env: &Env,
+    args: &TestArg,
+) -> Result<(), String> {
+    if !args.config_info.run_in_vm {
+        run_command_with_env(command, None, Some(env))?;
```
antoyo, line 330: This is very confusing: please rename the function to let it known that it doesn't necessarily run in the vm.

build_system/src/test.rs
```diff
@@ -1,15 +1,1176 @@
...
+    let mut command: Vec<&dyn AsRef<OsStr>> = Vec::new();
+    for arg in args.config_info.rustc_command.iter() {
+        command.push(arg);
+    }
+    command.extend_from_slice(&[
+        &"example/subslice-patterns-const-eval.rs",
+        &"--crate-type",
+        &"bin",
+        &"--target",
+        &args.config_info.target_triple,
```
antoyo, line 486: Shouldn't we add `TEST_FLAGS` here?

build_system/src/test.rs
```diff
@@ -1,15 +1,1176 @@
...
+    let mut command: Vec<&dyn AsRef<OsStr>> = Vec::new();
+    for arg in args.config_info.rustc_command.iter() {
+        command.push(arg);
+    }
+    command.extend_from_slice(&[
+        &"example/track-caller-attribute.rs",
+        &"--crate-type",
+        &"bin",
+        &"--target",
+        &args.config_info.target_triple,
```
antoyo, line 506: Same for `TEST_FLAGS`.

build_system/src/test.rs
```diff
@@ -1,15 +1,1176 @@
...
+        None,
+        Some(env),
+    );
+    run_command_with_output_and_env(&[&"git", &"fetch"], rust_dir, Some(env))?;
+    let rustc_commit = match rustc_version_info(env.get("RUSTC").map(|s| s.as_str()))?.commit_hash {
+        Some(commit_hash) => commit_hash,
+        None => return Err("Couldn't retrieve rustc commit hash".to_string()),
+    };
+    run_command_with_output_and_env(&[&"git", &"checkout", &rustc_commit], rust_dir, Some(env))?;
+    env.insert("RUSTFLAGS".to_string(), String::new());
```
antoyo, line 550: Do you think this is still needed, or we did that in the bash version to remove the value from this env var?

antoyo (commented):
That's all for now.

build_system/src/build.rs
GuillaumeGomez: I have no idea.

build_system/src/build.rs
```diff
@@ -50,25 +43,11 @@ impl BuildArg {
...
-                "--target-triple" => {
-                    if args.next().is_some() {
-                        // Handled in config.rs.
-                    } else {
-                        return Err(
-                            "Expected a value after `--target-triple`, found nothing".to_string()
-                        );
-                    }
-                }
-                "--target" => {
```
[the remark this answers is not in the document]
antoyo: That's not tested in the CI, currently.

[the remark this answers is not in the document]
antoyo: The case is when testing for a target unsupported by rustc.

build_system/src/build.rs
GuillaumeGomez: I did some back and forth changes so now we can indeed put back `args` like before.

build_system/src/config.rs
GuillaumeGomez: It can indeed be removed.

build_system/src/config.rs
GuillaumeGomez: Yes.

build_system/src/test.rs
```diff
@@ -1,15 +1,1176 @@
...
+        let mut args = std::env::args().skip(2);
+        let runners = get_runners();
+
+        while let Some(arg) = args.next() {
+            match arg.as_str() {
+                "--release" => {
+                    test_arg.channel = Channel::Release;
+                    test_arg.config_info.sysroot_release_channel = true;
+                }
+                "--release-sysroot" => test_arg.config_info.sysroot_release_channel = true,
```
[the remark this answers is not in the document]
GuillaumeGomez: Yes it is, good catch!

build_system/src/test.rs
```diff
@@ -1,15 +1,1176 @@
...
+        }
+        Ok(Some(test_arg))
+    }
+}
+
+fn build_if_no_backend(env: &Env, args: &TestArg) -> Result<(), String> {
+    if args.use_backend {
+        return Ok(());
+    }
+    let mut command: Vec<&dyn AsRef<OsStr>> = vec![&"cargo", &"rustc"];
```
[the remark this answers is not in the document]
GuillaumeGomez: Without it:

```
cargo check
    Checking y v0.1.0 (/home/imperio/rust/rustc_codegen_gcc/build_system)
error[E0308]: mismatched types
    --> src/test.rs:225:26
     |
225  |             command.push(flag);
     |                     ---- ^^^^ expected `&&str`, found `&String`
     |                     |
     |                     arguments to this method are incorrect
     |
     = note: expected reference `&&str`
                found reference `&String`
note: method defined here
    --> /home/imperio/.rustup/toolchains/nightly-2023-11-17-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/alloc/src/vec/mod.rs:1880:12
     |
1880 |     pub fn push(&mut self, value: T) {
     |            ^^^^

error[E0308]: mismatched types
   --> src/test.rs:227:41
    |
227 |         run_command_with_output_and_env(&command, None, Some(&env))
    |         ------------------------------- ^^^^^^^^ expected `&[&dyn AsRef<OsStr>]`, found `&Vec<&&str>`
    |         |
    |         arguments to this function are incorrect
    |
    = note: expected reference `&[&dyn AsRef<OsStr>]`
               found reference `&Vec<&&str>`
note: function defined here
   --> src/utils.rs:110:8
    |
110 | pub fn run_command_with_output_and_env(
    |        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
111 |     input: &[&dyn AsRef<OsStr>],
    |     ---------------------------

error[E0308]: mismatched types
    --> src/test.rs:230:26
     |
230  |             command.push(flag);
     |                     ---- ^^^^ expected `&&str`, found `&String`
     |                     |
     |                     arguments to this method are incorrect
     |
     = note: expected reference `&&str`
                found reference `&String`
note: method defined here
    --> /home/imperio/.rustup/toolchains/nightly-2023-11-17-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/alloc/src/vec/mod.rs:1880:12
     |
1880 |     pub fn push(&mut self, value: T) {
     |            ^^^^

error[E0308]: mismatched types
   --> src/test.rs:232:41
    |
232 |         run_command_with_output_and_env(&command, None, Some(&env))
    |         ------------------------------- ^^^^^^^^ expected `&[&dyn AsRef<OsStr>]`, found `&Vec<&&str>`
    |         |
    |         arguments to this function are incorrect
    |
    = note: expected reference `&[&dyn AsRef<OsStr>]`
               found reference `&Vec<&&str>`
note: function defined here
   --> src/utils.rs:110:8
    |
110 | pub fn run_command_with_output_and_env(
    |        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
111 |     input: &[&dyn AsRef<OsStr>],
    |     ---------------------------

For more information about this error, try `rustc --explain E0308`.
error: could not compile `y` (bin "y") due to 4 previous errors
```

build_system/src/test.rs
```diff
@@ -1,15 +1,1176 @@
...
+        &"lib",
+        &"--target",
+        &args.config_info.target_triple,
+    ]);
+    run_command_with_output_and_env(&command, None, Some(&env))?;
+
+    // FIXME: create a function "display_if_not_quiet" or something along the line.
+    println!("[AOT] mini_core_hello_world");
+    command.clear();
+    for arg in args.config_info.rustc_command.iter() {
```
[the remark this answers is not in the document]
GuillaumeGomez: Good idea.

build_system/src/test.rs
GuillaumeGomez: `run_command_in_vm` you mean?

build_system/src/test.rs
GuillaumeGomez: Oh wow, indeed.

build_system/src/test.rs
GuillaumeGomez: Yes indeed, completely missed it.

build_system/src/test.rs
GuillaumeGomez: I'll add a FIXME to check later on.

build_system/src/test.rs
```diff
@@ -1,15 +1,1176 @@
...
+        Ok(())
+    })
+}
+
+fn run_cargo_command_with_callback<F>(
+    command: &[&dyn AsRef<OsStr>],
+    cwd: Option<&Path>,
+    env: &Env,
+    args: &TestArg,
+    callback: F,
```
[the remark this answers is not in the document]
GuillaumeGomez: I don't see how it can look better considering we set env variables here as well as building the command vec.

build_system/src/test.rs
antoyo: Yes.

build_system/src/test.rs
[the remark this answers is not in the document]
antoyo: Yes, that would be done in `run_cargo_command_inner()`, so no duplication as well without the callback argument.

build_system/src/test.rs
GuillaumeGomez: Indeed!

build_system/src/utils.rs
GuillaumeGomez: It's not because of that I did it, it's because I'm not sure that the output will be valid utf8 content.

build_system/src/utils.rs
GuillaumeGomez: I was hesitant about that but I guess it'd be better indeed. better to catch invalid arguments early on rather than when debugging.

build_system/src/test.rs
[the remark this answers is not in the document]
GuillaumeGomez: I don't understand at all what you want me to do here. ^^'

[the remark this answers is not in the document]
antoyo: Make this function instead, so you won't need a callback:
```rust
fn run_cargo_command_inner<F>(
     command: &[&dyn AsRef<OsStr>],
     cwd: Option<&Path>,
     env: &mut Env,
     args: &TestArg,
) -> Result<Vec<&dyn AsRef<OsStr>>, String>
 {
     let toolchain = get_toolchain()?;
     let toolchain_arg = format!("+{}", toolchain);
     let rustc_version = String::from_utf8(
         run_command_with_env(&[&args.config_info.rustc_command[0], &"-V"], cwd, Some(env))?.stdout,
     )
     .map_err(|error| format!("Failed to retrieve rustc version: {:?}", error))?;
     let rustc_toolchain_version = String::from_utf8(
         run_command_with_env(
             &[&args.config_info.rustc_command[0], &toolchain_arg, &"-V"],
             cwd,
             Some(env),
         )?
         .stdout,
     )
     .map_err(|error| format!("Failed to retrieve rustc +toolchain version: {:?}", error))?;

     if rustc_version != rustc_toolchain_version {
         eprintln!(
             "rustc_codegen_gcc is built for `{}` but the default rustc version is `{}`.",
             rustc_toolchain_version, rustc_version,
         );
         eprintln!("Using `{}`.", rustc_toolchain_version);
     }
     let mut env = env.clone();
     let rustflags = env.get("RUSTFLAGS").cloned().unwrap_or_default();
     env.insert("RUSTDOCFLAGS".to_string(), rustflags);
     let mut cargo_command: Vec<&dyn AsRef<OsStr>> = vec![&"cargo", &toolchain_arg];
     cargo_command.extend_from_slice(&command);
     cargo_command
 }
```

and then, you can implement the above function like this:
```rust
 fn run_cargo_command(
     command: &[&dyn AsRef<OsStr>],
     cwd: Option<&Path>,
     env: &Env,
     args: &TestArg,
 ) -> Result<(), String> {
     let mut env = env.clone();
     let cargo_command = run_cargo_command_inner(command, cwd, &mut env, args)?;
     run_command_with_output_and_env(cargo_command, cwd, Some(env))?;
     Ok(())
 }
```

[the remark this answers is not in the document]
GuillaumeGomez: But then it doesn't work because the `Vec` we return contains local references in the function. Fun times. :)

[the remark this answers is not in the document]
antoyo: Ok, then.

build_system/src/test.rs
GuillaumeGomez: I rewrote it. Much simpler now indeed.

build_system/src/build.rs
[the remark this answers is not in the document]
antoyo: Please add this back.

build_system/src/test.rs
antoyo: Please adjust this comment.

build_system/src/test.rs
antoyo: Seems like this doesn't really affect the performance.

build_system/src/utils.rs
```diff
@@ -238,3 +299,68 @@ where
...
     Ok(())
 }
+
+pub fn split_args(args: &str) -> Result<Vec<String>, String> {
+    let mut out = Vec::new();
+    let mut start = 0;
+    let args = args.trim();
+    let mut iter = args.char_indices().peekable();
+
+    while iter.peek().is_some() {
```
antoyo, line 309: Do you really need this outer loop? If so, can you add a comment to explain why?

GuillaumeGomez: No it appears to be useless. Good catch!

build_system/src/config.rs
```diff
@@ -1,149 +1,264 @@
...
+            // a bit differently.
+            self.cg_backend_path = "gcc".to_string();
+
+            match env.get("RUSTC_SYSROOT") {
+                Some(rustc_sysroot) if !rustc_sysroot.is_empty() => {
+                    rustflags.extend_from_slice(&["--sysroot".to_string(), rustc_sysroot.clone()]);
+                }
+                _ => {}
+            }
+            rustflags.push("-Cpanic=abort".to_string());
```
antoyo, line 165: I find this strange that we specify `-Cpanic=abort` not in the usual way, e.g. by not specifying `--features master`.
If this line is necessary, can you please add a comment to explain why.

build_system/src/config.rs
```diff
@@ -1,149 +1,264 @@
...
+            "Linux" => "so",
+            "Darwin" => "dylib",
+            os => return Err(format!("unsupported OS `{}`", os)),
+        }
+        .to_string();
+        let rustc = match env.get("RUSTC") {
+            Some(r) if !r.is_empty() => r.to_string(),
+            _ => "rustc".to_string(),
+        };
+        self.host_triple = rustc_version_info(Some(&rustc))?.host.unwrap_or_default();
```
antoyo, line 114: I believe we should error out and not use the empty string as this is going to cause issues later if it is empty.

build_system/src/config.rs
```diff
@@ -1,149 +1,264 @@
...
+        // display metadata load errors
+        env.insert("RUSTC_LOG".to_string(), "warn".to_string());
+
+        let sysroot = current_dir.join(&format!(
+            "build_sysroot/sysroot/lib/rustlib/{}/lib",
+            self.target_triple,
+        ));
+        let ld_library_path = format!(
+            "{target}:{sysroot}:{gcc_path}",
+            target = current_dir.join("target/out").display(),
```
antoyo, line 219: I believe we can set a different directory for this. Is this an issue?

build_system/src/config.rs
GuillaumeGomez: Ok, updating then.

build_system/src/config.rs
GuillaumeGomez: We can but it's rare. I'll add a fixme comment though for potential future improvements.

build_system/src/config.rs
GuillaumeGomez: It's the equivalent of `config.sh` line 56. So you tell me. ^^'

antoyo: You wrote this code to add the tests in the rust CI, so I don't know.

build_system/src/utils.rs
```diff
@@ -29,22 +30,43 @@ fn check_exit_status(
...
+    let input = input.iter().map(|i| i.as_ref()).collect::<Vec<&OsStr>>();
+    eprintln!("Command `{:?}` failed", input);
+    if let Some(output) = output {
+        let stdout = String::from_utf8_lossy(&output.stdout);
+        if !stdout.is_empty() {
+            error.push_str("\n==== STDOUT ====\n");
+            match stdout {
+                Cow::Owned(s) => error.push_str(&s),
+                Cow::Borrowed(s) => error.push_str(s),
+            }
```
antoyo, line 58: Seems like this would work:
```suggestion
            error.push_str(&*stdout);
```

build_system/src/utils.rs
```diff
@@ -29,22 +30,43 @@ fn check_exit_status(
...
+        }
+        let stderr = String::from_utf8_lossy(&output.stderr);
+        if !stderr.is_empty() {
+            error.push_str("\n==== STDERR ====\n");
+            match stderr {
+                Cow::Owned(s) => error.push_str(&s),
+                Cow::Borrowed(s) => error.push_str(s),
+            }
```
antoyo, line 66: Same.

antoyo (commented):
A few more stuff. Only one file that I still need to review.

build_system/src/utils.rs
GuillaumeGomez: Wow, that's great, thanks!

antoyo:
Some issues with the tests:

* `--mini-tests`, `--std-tests`:
   * Before: it showed the output of the executables ran.
   * After: it is not shown.
   * Please make it so that we see the output.
* `--extended-rand-tests`, `--extended-regex-example-tests`, `--extended-regex-tests`: The CI task for libgccjit 12 should be removed since it is skipped (I know it ran before this PR, but it's easy to remove in gcc12.yml).

antoyo (commented):
I'm about half-way through the last file to review.

build_system/src/test.rs
```diff
@@ -1,15 +1,1132 @@
...
+    );
+    run_command(&[&"git", &"checkout", &"--", &"tests/"], rust_dir)?;
+    run_command_with_output_and_env(&[&"git", &"fetch"], rust_dir, Some(env))?;
+    let rustc_commit = match rustc_version_info(env.get("RUSTC").map(|s| s.as_str()))?.commit_hash {
+        Some(commit_hash) => commit_hash,
+        None => return Err("Couldn't retrieve rustc commit hash".to_string()),
+    };
+    run_command_with_output_and_env(&[&"git", &"checkout", &rustc_commit], rust_dir, Some(env))?;
+    // FIXME: Is it really needed to empty `RUSTFLAGS` here?
+    // env.insert("RUSTFLAGS".to_string(), String::new());
```
antoyo, line 508: Leftover?

build_system/src/test.rs
```diff
@@ -1,15 +1,1132 @@
...
+            &"./x.py",
+            &"test",
+            &"--run",
+            &"always",
+            &"--stage",
+            &"0",
+            &"tests/assembly/asm",
+            &"--rustc-args",
+            &format!(
+                r#"-Zpanic-abort-tests -Csymbol-mangling-version=v0 \
```
antoyo, line 589: Won't putting all those flags in the same string make the shell runner consider it as only one argument?

build_system/src/test.rs
```diff
@@ -1,15 +1,1132 @@
...
+        &[
+            &"./x.py",
+            &"test",
+            &"--run",
+            &"always",
+            &"--stage",
+            &"0",
+            &"tests/ui",
+            &"--rustc-args",
+            &rustc_args,
```
antoyo, line 1002: Won't putting all those flags in the same string make the shell runner consider it as only one argument?

build_system/src/test.rs
```diff
@@ -1,15 +1,1132 @@
...
+        let end = current_part * count + count;
+        // We remove the files we don't want to test.
+        for path in files
+            .iter()
+            .enumerate()
+            .filter(|(pos, _)| *pos < start || *pos >= end)
+            .map(|(_, path)| path)
+        {
+            remove_file(&rust_path.join(path))?;
+        }
```
antoyo, line 978: This would probably be simple with [`skip()`](https://doc.rust-lang.org/stable/std/iter/trait.Iterator.html#method.skip) and [`take()`](https://doc.rust-lang.org/stable/std/iter/trait.Iterator.html#method.take).
This could be in a follow-up PR, though.

build_system/src/test.rs
```diff
@@ -1,15 +1,1132 @@
...
-        args.push(extra_arg);
+use std::collections::{BTreeSet, HashMap};
+use std::ffi::OsStr;
+use std::fs::{remove_dir_all, File};
+use std::io::{BufRead, BufReader};
+use std::path::{Path, PathBuf};
+use std::str::FromStr;
+
+type Env = HashMap<String, String>;
+type Runner = &'static dyn Fn(&Env, &TestArg) -> Result<(), String>;
```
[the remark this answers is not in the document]
GuillaumeGomez: The first one works just fine. Thanks for the improvement!

build_system/src/test.rs
```diff
@@ -1,15 +1,1132 @@
...
+    --features [arg]       : Add a new feature [arg]
+    --use-system-gcc       : Use system installed libgccjit
+    --build-only           : Only build rustc_codegen_gcc then exits
+    --use-backend          : Useful only for rustc testsuite
+    --nb-parts             : Used to split rustc_tests (for CI needs)
+    --current-part         : Used with `--nb-parts`, allows you to specify which parts to test"#
+    );
+    ConfigInfo::show_usage();
+    for (option, (doc, _)) in get_runners() {
+        let needed_spaces = 23_usize.saturating_sub(option.len());
```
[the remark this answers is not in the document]
GuillaumeGomez: I'll add a FIXME comment for the moment.

build_system/src/test.rs
GuillaumeGomez: Yes.

build_system/src/test.rs
GuillaumeGomez: I opened https://github.com/rust-lang/rustc_codegen_gcc/issues/395.

build_system/src/test.rs
```diff
@@ -207,9 +207,6 @@ impl TestArg {
 }
 
 fn build_if_no_backend(env: &Env, args: &TestArg) -> Result<(), String> {
-    if args.use_backend {
```
antoyo: I believe it should have been replaced by:
```rust
if args.backend.is_some()
```
but you know this code better than me.

build_system/src/test.rs
```diff
@@ -13,50 +13,50 @@ use std::path::{Path, PathBuf};
...
+type Runner = fn(&Env, &TestArg) -> Result<(), String>;
 type Runners = HashMap<&'static str, (&'static str, Runner)>;
 
 fn get_runners() -> Runners {
     let mut runners = HashMap::new();
 
     runners.insert(
         "--test-rustc",
-        ("Run all rustc tests", &test_rustc as Runner),
+        ("Run all rustc tests", test_rustc as Runner),
```
antoyo, line 24: The cast `as Runner` is probably not necessary anymore.

build_system/src/test.rs
```diff
@@ -1,15 +1,1130 @@
...
+// COMPILE_MOD_BENCH_LLVM_3="rustc example/mod_bench.rs --crate-type bin -Copt-level=3 -o $cargo_target_dir/mod_bench_llvm_3 -Cpanic=abort"
+//
+// Use 100 runs, because a single compilations doesn't take more than ~150ms, so it isn't very slow
+// hyperfine --runs ${COMPILE_RUNS:-100} "$COMPILE_MOD_BENCH_INLINE" "$COMPILE_MOD_BENCH_LLVM_0" "$COMPILE_MOD_BENCH_LLVM_1" "$COMPILE_MOD_BENCH_LLVM_2" "$COMPILE_MOD_BENCH_LLVM_3"
+// echo "[BENCH RUN] mod_bench"
+// hyperfine --runs ${RUN_RUNS:-10} $cargo_target_dir/mod_bench{,_inline} $cargo_target_dir/mod_bench_llvm_*
+
+fn extended_rand_tests(env: &Env, args: &TestArg) -> Result<(), String> {
+    if !args.is_using_gcc_master_branch() {
+        println!("Using GCC master branch. Skipping `extended_rand_tests`.");
```
antoyo, line 682:
```suggestion
        println!("Not using GCC master branch. Skipping `extended_rand_tests`.");
```
and same for the others.

antoyo (commented):
Hopefully, we can merge after this.

build_system/src/test.rs
GuillaumeGomez: No, this one is still needed.

GuillaumeGomez:
```
rror[E0308]: mismatched types
  --> src/test.rs:28:40
   |
28 |         ("Run successful rustc tests", test_successful_rustc),
   |                                        ^^^^^^^^^^^^^^^^^^^^^ expected fn item, found a different fn item
   |
   = note: expected fn item `for<'a, 'b> fn(&'a HashMap<String, String>, &'b TestArg) -> Result<_, _> {test_rustc}`
              found fn item `for<'a, 'b> fn(&'a HashMap<String, String>, &'b TestArg) -> Result<_, _> {test_successful_rustc}`

error[E0308]: mismatched types
  --> src/test.rs:32:37
   |
32 |         ("Run failing rustc tests", test_failing_rustc),
   |                                     ^^^^^^^^^^^^^^^^^^ expected fn item, found a different fn item
   |
   = note: expected fn item `for<'a, 'b> fn(&'a HashMap<String, String>, &'b TestArg) -> Result<_, _> {test_rustc}`
              found fn item `for<'a, 'b> fn(&'a HashMap<String, String>, &'b TestArg) -> Result<_, _> {test_failing_rustc}`
...
```

antoyo (approved):
:partying_face: Thanks a lot!

PR #396, opened 2023-12-20
Rustify `clean_all.sh`

GuillaumeGomez:


build_system/src/clean.rs
```diff
@@ -0,0 +1,56 @@
+use crate::utils::remove_file;
+
+use std::fs::remove_dir_all;
+
+fn check_args() -> Result<Option<()>, String> {
+    for arg in std::env::args().skip(2) {
```
antoyo, line 6: I always like a comment to mention that this skips the binary name and the command `clean`.

build_system/src/clean.rs
```diff
@@ -0,0 +1,56 @@
...
+        match arg.as_str() {
+            "--help" => {
+                usage();
+                return Ok(None);
+            }
+            a => return Err(format!("Unknown argument `{a}`")),
```
antoyo, line 12: I prefer the old syntax, but I guess if this uses clippy, this might throw a warning:
```suggestion
            arg => return Err(format!("Unknown argument `{}`", arg)),
```
Hopefully, GitHub will highlight those inline arguments in the future.

build_system/src/main.rs
```diff
@@ -22,6 +23,7 @@ fn usage() {
         "\
 Available commands for build_system:
 
+    clean    : Run clean command
```
antoyo, line 26: I think I'll keep the name `clean-all` as I could easily see new commands added like `clean-projects`.

GuillaumeGomez: Wouldn't you prefer instead having `clean all` and `clean project`? Less top-level commands like that. And therefore I'd keep the current state until we add a new one.

antoyo: `clean all` is good, yeah.

GuillaumeGomez: So I add `all`?

antoyo: Yes.

GuillaumeGomez:
Updated!

antoyo (approved):

PR #399, opened 2023-12-20
If the rustc commit cannot be retrieve, just checkout the repository

GuillaumeGomez:


antoyo:
Please make sure we can run the tests locally with a custom toolchain.

PR #405, opened 2023-12-21
Rustify `cargo.sh`

GuillaumeGomez:


build_system/src/cargo.rs
```diff
@@ -0,0 +1,112 @@
...
+    get_toolchain, run_command_with_output_and_env, rustc_toolchain_version_info,
+    rustc_version_info,
+};
+
+use std::collections::HashMap;
+use std::ffi::OsStr;
+use std::path::PathBuf;
+
+fn args() -> Result<Option<Vec<String>>, String> {
+    // We skip the binary and the "clean" option.
```
antoyo, line 12:
```suggestion
    // We skip the binary and the "cargo" option.
```

build_system/src/cargo.rs
```diff
@@ -0,0 +1,112 @@
...
+            .unwrap_or(false)
+        {
+            parent_dir.pop();
+        } else {
+            return Err(format!(
+                "Build script not executed from `build_system/target/release/y` (in path {})",
+                current_exe.display(),
+            ));
+        }
+    }
```
antoyo, line 69: I'm not sure if you really need to change directory here.
I'm not sure I understand the original code, but it looks to me that it compared the toolchain set for the project with the toolchain needed by cg_gcc.

antoyo:
I tried cross-compiling a Rust program with this new cargo tool and it doesn't work.

build_system/src/cargo.rs
GuillaumeGomez: Yes we do because we need to read at least the `gcc-path` file and for that to work, we can either be in the right folder or give the extra path as argument. I preferred the current way (changing folder).

GuillaumeGomez:
Forgot to say but updated. ^^'

antoyo:
If a program is ran via `./y.sh cargo run` and exits with an error, we get this:
```
Command `["cargo", "+nightly-2023-11-17", "run"]` failed
Command failed to run: Command `cargo +nightly-2023-11-17 run` exited with status Some(1)
```
In this case, I would prefer not to show this to keep the same behavior as `cargo`.

Issue #381, opened 2023-11-10
CI: move away from actions-rs


sadlerap:
As of October 13th, 2023, most of the actions in `actions-rs` have been archived.  Even before then, some actions haven't seen an update in quite a long time (for instance, [`cargo`](https://github.com/actions-rs/cargo) hasn't received any updates since July 2021).

This might be a problem soon, since [github is deprecating support for actions running on node12](https://github.blog/changelog/2023-06-13-github-actions-all-actions-will-run-on-node16-instead-of-node12-by-default/) and forcing them to use node16 runners.  Since these are unmaintained, there's a possibility these actions could break without any possibility of fixes landing upstream.

It might be a good idea to find an alternative, though I'm not entirely sure what alternatives to these actions are available.  Does the `cargo` that comes built into the ubuntu image runners support picking up the `rust-toolchain` file as a version override?  If so, that might be an option.

madsmtm:
There is https://github.com/dtolnay/rust-toolchain, which seems to be what a lot of people are gravitating towards

PR #412, opened 2024-01-12
Switch from actions-rs to preinstalled rustup

tgross35:
actions-rs is deprecated. Switch to using the preinstalled rustup to install the toolchain, and https://github.com/Swatinem/rust-cache to configure cacheing.

https://github.com/dtolnay/rust-toolchain cannot be used because it does not respect `rust-toolchain`.

I believe there isn't any need to run `cargo build` separately anymore since there are no longer error annotations.

Fixes https://github.com/rust-lang/rustc_codegen_gcc/issues/381

### DIFF
--- a/.github/workflows/m68k.yml
+++ b/.github/workflows/m68k.yml
@@ -36,12 +36,19 @@ jobs:
         ]
 
     steps:
+    - uses: actions/checkout@v3
+
+    # `rustup show` installs from rust-toolchain.toml
+    - name: Setup rust toolchain
+      run: rustup show
+
+    - name: Setup rust cache
+      uses: Swatinem/rust-cache@v2
+
     - name: Install packages
       run: |
         sudo apt-get update
         sudo apt-get install qemu qemu-user-static
-
-    - uses: actions/checkout@v3
 
     - name: Download GCC artifact
       uses: dawidd6/action-download-artifact@v2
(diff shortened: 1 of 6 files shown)


.github/workflows/m68k.yml
```diff
@@ -36,13 +36,20 @@ jobs:
...
+
+    - name: Setup rust cache
+      uses: Swatinem/rust-cache@v2
+
     - name: Install packages
       run: |
         sudo apt-get update
         sudo apt-get install qemu qemu-user-static
 
-    - uses: actions/checkout@v3
```
antoyo, line 44: Was it intentional to remove this?

tgross35: I moved it to the top of the block to be consistent with the other files. Is there a reason it was lower?

antoyo: Ah, sorry, I missed this.
Move it should be fine.

tgross35:
@antoyo does matrix items libgccjit_version or commands ([link](https://github.com/rust-lang/rustc_codegen_gcc/pull/412/files#diff-b803fcb7f17ed9235f1e5cb1fcd2f5d3b2838429d4368ae4c57ce4436577f03fR21-R35)) affect what is built? If so, maybe they should be part of the cache key

antoyo:
> @antoyo does matrix items libgccjit_version or commands ([link](https://github.com/rust-lang/rustc_codegen_gcc/pull/412/files#diff-b803fcb7f17ed9235f1e5cb1fcd2f5d3b2838429d4368ae4c57ce4436577f03fR21-R35)) affect what is built? If so, maybe they should be part of the cache key

No, rustc_codegen_gcc itself is built with LLVM, so the crates used by it are not affected by those flags.

antoyo:
It seems like you forgot to remove `actions-rs` in the [gcc12 CI](https://github.com/rust-lang/rustc_codegen_gcc/pull/412/files#diff-8cbc9596828db027668803e348fc382630a5c6b7bf04e63851af2dea94559151R82).

tgross35:
Good catch, removed that

antoyo:
Thanks a lot!

PR #416, opened 2024-01-20
Improve iterator for files suppression

### LINKED ISSUE #395: Improve iterator for files suppression
In `build_system/test.rs` we currently have:

```rust
        for path in files
            .iter()
            .enumerate()
            .filter(|(pos, _)| *pos < start || *pos >= end)
            .map(|(_, path)| path)
        {
            remove_file(&rust_path.join(path))?;
        }
```

It would probably be simpler to use `skip` and `take`.

Liewyec:
fixes #395

### DIFF
--- a/build_system/src/test.rs
+++ b/build_system/src/test.rs
@@ -979,12 +979,7 @@ where
         let start = current_part * count;
         let end = current_part * count + count;
         // We remove the files we don't want to test.
-        for path in files
-            .iter()
-            .enumerate()
-            .filter(|(pos, _)| *pos < start || *pos >= end)
-            .map(|(_, path)| path)
-        {
+        for path in files.iter().skip(start).take(count) {
             remove_file(&rust_path.join(path))?;
         }
     }


build_system/src/test.rs
```diff
@@ -982,7 +982,8 @@ where
         for path in files
             .iter()
             .enumerate()
-            .filter(|(pos, _)| *pos < start || *pos >= end)
+            .take(start)
+            .skip(count)
             .map(|(_, path)| path)
```
Speedy37, line 987:
```rust
files.iter().take(start).skip(count)
```

you don't need enumerate and map anymore.

Liewyec: I asked in the issue, but let repeat it here: I do not think this can be done with skip and take, because the filter if I understand it correctly yields items from beginning of the files, skips some in the middle and then possibly yields additional items from end. And I do not know how to do this with take and skip.

antoyo: I believe the original code might be wrong. You probably want `.skip(start).take(count)`, though.

The code is supposed to take contiguous elements: that's why I think it might be wrong. Please use `.skip(start).take(count)` and I'll check if this is OK.

Liewyec: I updated the code.

build_system/src/test.rs
```diff
@@ -979,12 +979,7 @@ where
         let start = current_part * count;
         let end = current_part * count + count;
         // We remove the files we don't want to test.
-        for path in files
-            .iter()
-            .enumerate()
-            .filter(|(pos, _)| *pos < start || *pos >= end)
-            .map(|(_, path)| path)
-        {
+        for path in files.iter().skip(start).take(count) {
```
GuillaumeGomez, line 982: Nice, thanks for the improvement!

GuillaumeGomez (approved):

build_system/src/test.rs
tempdragon: The original code may be likely to be corresponding to 
```Rust
for path in files.iter().take(start).chain(files.iter().skip(end)) {
```
Which is weird.

antoyo:
Thanks for your contribution!

PR #420, opened 2024-01-29
Implement prefetch data intrinsics

### LINKED ISSUE #414: Minimal implimentation of prefetch
https://github.com/rust-lang/rustc_codegen_gcc/blob/1a8e0c39fac364211a92ace4f872f6091518bd50/src/intrinsic/mod.rs#L170-L175

Currently, all the prefetch operations lead to `unimplemented!()` macro. However, prefetches are just a suggestion and can be interpreted as a noop, like assume is.

NCGThompson:
Fixes #414 

It should probably have tests that should fail before the changes. What kind of test should I add?

I edited the builtin list in context.rs, and that might create a conflict when syncing with the rust repository.

### DIFF
--- a/src/context.rs
+++ b/src/context.rs
@@ -209,7 +209,7 @@ impl<'gcc, 'tcx> CodegenCx<'gcc, 'tcx> {
             "__builtin_usub_overflow", "sqrtf", "sqrt", "__builtin_powif", "__builtin_powi", "sinf", "sin", "cosf", "cos",
             "powf", "pow", "expf", "exp", "exp2f", "exp2", "logf", "log", "log10f", "log10", "log2f", "log2", "fmaf",
             "fma", "fabsf", "fabs", "fminf", "fmin", "fmaxf", "fmax", "copysignf", "copysign", "floorf", "floor", "ceilf",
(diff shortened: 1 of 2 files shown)


antoyo:
> It should probably have tests that should fail before the changes. What kind of test should I add?

Do you know what kind of tests there is for this in the LLVM codegen?

NCGThompson:
The codegen tests are probably the most thorough, but they are LLVM specific. However, according to the FileCheck documentation, it doesn't do much that can't be done with grep.

At a minimum though, I think we should check that the compiler can encounter the function without panicking.

antoyo:
> At a minimum though, I think we should check that the compiler can encounter the function without panicking.

I would be OK with such a test. You can add it [in this directory](https://github.com/rust-lang/rustc_codegen_gcc/tree/master/tests/run).

antoyo:
Is this still a draft or should I review this PR?

antoyo:
Oh, sorry, I didn't notice you marked this PR as ready for review (please ping me if I don't answer for some time: I can easily miss some things).

There are some conflicts to fix.

PR #427, opened 2024-02-09
Switch to `config.toml` instead of `gcc-path`

GuillaumeGomez:
This is the first step to support the `download` command.


build_system/src/config.rs
```diff
@@ -19,6 +23,72 @@ impl Channel {
...
+pub struct ConfigFile {
+    gcc_path: Option<String>,
+    download_gccjit: Option<bool>,
+}
+
+impl ConfigFile {
+    pub const CONFIG_FILE: &'static str = "config.toml";
+
+    pub fn new() -> Result<Self, String> {
+        let content = fs::read_to_string(Self::CONFIG_FILE).map_err(|_| {
```
bjorn3, line 44: This requires `config.toml` to be at the root of the cg_gcc source, right? When building as part of rustc that will not work. The source directory should be assumed to be immutable, thus preventing that file to be created by either the user or the rust build system. You could add a way to configure where it will look for `config.toml`.

PR #430, opened 2024-02-12
Add `download` config

GuillaumeGomez:
First commit comes from https://github.com/rust-lang/rustc_codegen_gcc/pull/429.


build_system/src/config.rs
```diff
@@ -159,30 +187,151 @@ impl ConfigInfo {
...
+                        &"[Net.ServicePointManager]::SecurityProtocol = [Net.SecurityProtocolType]::Tls12;",
+                        &format!(
+                            "(New-Object System.Net.WebClient).DownloadFile('{}', '{}')",
+                            url,
+                            tempfile_name,
+                        ).as_str(),
+                    ],
+                    Some(&output_dir),
+                );
             }
```
antoyo, line 284: Please put this in a separate function (or two separate functions with `#[cfg(windows)]` if you don't want to run `curl` on Windows) since this is platform-specific.

build_system/src/config.rs
```diff
@@ -159,30 +187,151 @@ impl ConfigInfo {
...
+                format!(
+                    "Failed to rename `{}` into `{}`: {:?}",
+                    tempfile.display(),
+                    libgccjit_so.display(),
+                    err,
+                )
+            })?;
+
+            println!("Downloaded libgccjit.so version {} successfully!", commit);
+            create_symlink(
```
antoyo, line 299: Please add a comment explaining why we need this symlink.

build_system/src/main.rs
```diff
@@ -29,6 +30,7 @@ Available commands for build_system:
     prepare  : Run prepare command
     build    : Run build command
     test     : Run test command
+    info:    : Run info command
```
antoyo, line 33: At this point, I'm starting to believe we should give better explanation of what each command does here.
But this can be done in a later PR.

build_system/src/config.rs
```diff
@@ -159,30 +187,148 @@ impl ConfigInfo {
         command
     }
 
-    pub fn setup_gcc_path(&mut self, override_gcc_path: Option<&str>) -> Result<(), String> {
-        let ConfigFile { gcc_path, .. } = ConfigFile::new(self.config_file.as_deref())?;
+    fn download_gccjit_if_needed(&mut self) -> Result<(), String> {
+        let output_dir = Path::new(
+            std::env::var("CARGO_TARGET_DIR")
```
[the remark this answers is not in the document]
GuillaumeGomez: Indeed. I plan to clean all this up in a future PR to have less top-level folders.

build_system/src/config.rs
```diff
@@ -159,30 +187,151 @@ impl ConfigInfo {
...
-                path.to_string()
+        let commit_hash_file = self.compute_path("libgccjit.version");
+        let content = fs::read_to_string(&commit_hash_file).map_err(|_| {
+            format!(
+                "Failed to read `{}`. Take a look at `Readme.md` to see how to set up the project",
+                commit_hash_file.display(),
+            )
+        })?;
+        let commit = content.trim();
+        if commit.contains('/') || commit.contains('\\') {
```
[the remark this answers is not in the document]
GuillaumeGomez: To be sure it's not a path. I think the rest is "ok" to ignore, it'll just fail.

build_system/src/config.rs
GuillaumeGomez: :+1:

build_system/src/config.rs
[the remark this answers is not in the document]
antoyo: Please add a comment to explain this.

PR #432, opened 2024-02-13
Use shallow clone in test.rs to reduce cloning overhead

tempdragon:
GitHub may not necessarily be accessible in some conditions/locations. Switching to a shallow clone can significantly reduce the time for cloning rust from GitHub.

### DIFF
--- a/build_system/src/test.rs
+++ b/build_system/src/test.rs
@@ -499,17 +499,29 @@ fn setup_rustc(env: &mut Env, args: &TestArg) -> Result<PathBuf, String> {
             &"clone",
             &"https://github.com/rust-lang/rust.git",
             &rust_dir_path,
+            &"--depth",
+            &"1",
         ],
         None,
         Some(env),
     );
     let rust_dir: Option<&Path> = Some(&rust_dir_path);
     run_command(&[&"git", &"checkout", &"--", &"tests/"], rust_dir)?;
-    run_command_with_output_and_env(&[&"git", &"fetch"], rust_dir, Some(env))?;
     let rustc_commit = match rustc_version_info(env.get("RUSTC").map(|s| s.as_str()))?.commit_hash {
         Some(commit_hash) => commit_hash,
         None => return Err("Couldn't retrieve rustc commit hash".to_string()),
     };
+    run_command_with_output_and_env(
+        &[
+            &"git",
+            &"fetch",
+            &"https://github.com/rust-lang/rust.git",
+            &rustc_commit.as_str(),
+            &"--depth=1",
+        ],
+        rust_dir,
+        Some(env),
+    )?;
     if rustc_commit != "unknown" {
         run_command_with_output_and_env(
             &[&"git", &"checkout", &rustc_commit],
--- a/build_system/src/test.rs
+++ b/build_system/src/test.rs
@@ -511,17 +511,30 @@ fn setup_rustc(env: &mut Env, args: &TestArg) -> Result<PathBuf, String> {
         Some(commit_hash) => commit_hash,
         None => return Err("Couldn't retrieve rustc commit hash".to_string()),
     };
-    run_command_with_output_and_env(
-        &[
-            &"git",
-            &"fetch",
-            &"https://github.com/rust-lang/rust.git",
-            &rustc_commit.as_str(),
-            &"--depth=1",
-        ],
-        rust_dir,
-        Some(env),
-    )?;
+    let has_commit = {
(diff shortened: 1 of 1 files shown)


build_system/src/test.rs
```diff
@@ -494,16 +497,32 @@ fn setup_rustc(env: &mut Env, args: &TestArg) -> Result<(), String> {
     let rust_dir = Some(Path::new("rust"));
     // If the repository was already cloned, command will fail, so doesn't matter.
     let _ = run_command_with_output_and_env(
-        &[&"git", &"clone", &"https://github.com/rust-lang/rust.git"],
+        &[
+            &"git",
+            &"clone",
+            &"https://github.com/rust-lang/rust.git",
+            &"--depth",
+            &"1",
```
bjorn3, line 505: `--filter=blob:0` is another option which may behave better.

bjorn3: I mean in the place of `--depth 1` to make future `git fetch` cheaper. Removing the `git fetch` breaks when the rust repo has already been cloned previously as it may not have the right commit.

tempdragon: Sorry, I didn't really understand the usage of this option(nor have I ever used this) and thought it could add up.

tempdragon: Thanks for advice.

tempdragon: Somehow, the option didn't work as I expected after the demanded replacement.
```
$ git clone https://github.com/rust-lang/rust.git --filter=blob:none

Cloning into 'rust'...
remote: Enumerating objects: 1598065, done.
remote: Counting objects: 100% (1056/1056), done.
remote: Compressing objects: 100% (693/693), done.
Receiving objects:   1% (18171/1598065), 6.85 MiB | 24.00 KiB/s
```
~600MiB
Is it that I misunderstood your advice?

bjorn3: Looks like it is way less effective on the rust repo compared to a shallow clone than I expected. (I just tired it and I'm getting 450MB locally) Even `--filter=tree:0` uses 110MB compared to the 30MB for `--depth 1`. Given the huge difference, I did say that using `--depth 1` is fine.

tempdragon: The option didn't seem to work as expected. I have probably *misunderstood* your advice.
Please point out my mistake if so. Sorry for wasting your time as a green hand.

Just in case you really mean direct replacement, the following experiments may be of relevance.
With only the `--filter=blob:none`(0 seem to be a wrong choice), the objects remain extremely large:
```
$ git clone https://github.com/rust-lang/rust.git --filter=blob:none
Cloning into 'rust'...
remote: Enumerating objects: 1598065, done.
remote: Counting objects: 100% (1056/1056), done.
remote: Compressing objects: 100% (693/693), done.
Receiving objects:   2% (32526/1598065), 12.48 MiB | 57.00 KiB/s
```
~600MiB

while the larger seem to be smaller.
```
$ git clone https://github.com/rust-lang/rust.git --filter=blob:none --depth=1
Cloning into 'rust'...
remote: Enumerating objects: 3287, done.
remote: Counting objects: 100% (3287/3287), done.
remote: Compressing objects: 100% (2661/2661), done.
remote: Total 3287 (delta 7), reused 2314 (delta 7), pack-reused 0
Receiving objects: 100% (3287/3287), 1.45 MiB | 5.32 MiB/s, done.
Resolving deltas: 100% (7/7), done.
remote: Enumerating objects: 44743, done.
remote: Counting objects: 100% (44743/44743), done.
remote: Compressing objects: 100% (37881/37881), done.
Receiving objects: 100% (44743/44743), 28.86 MiB | 1.65 MiB/s, done.
```
~30MiB

which are both lower than the average size.

As for the real effect of the two options added together, we can see the original download size to be slightly SMALLER than the previous choice:

```
$ git clone https://github.com/rust-lang/rust.git --depth=1
Cloning into 'rust'...
remote: Enumerating objects: 48030, done.
remote: Counting objects: 100% (48030/48030), done.
remote: Compressing objects: 100% (40550/40550), done.
remote: Total 48030 (delta 6420), reused 36610 (delta 5967), pack-reused 0
Receiving objects: 100% (48030/48030), 30.22 MiB | 186.00 KiB/s, done.
Resolving deltas: 100% (6420/6420), done.
```

GuillaumeGomez (approved):

GuillaumeGomez:
I remember having issues when doing shallow clones. However I can't figure which ones. Let's get this in for now and if issues arise, we can always revert. Thanks!

PR #444, opened 2024-02-17
Improve docs

GuillaumeGomez:


doc/debugging-libgccjit.md
```diff
@@ -0,0 +1,78 @@
...
+
+```
+set substitute-path /usr/src/debug/gcc /path/to/gcc-repo/gcc
+```
+
+TODO(antoyo): but that's not what I remember I was doing.
+
+# How to debug GCC LTO
+
+Run do the command with `-v -save-temps` and then extract the `lto1` line from the output and run that under the debugger.
```
antoyo, line 78: This is duplicated in the file above.

doc/tips.md
```diff
@@ -0,0 +1,74 @@
...
+`rustc` needs to be built without `jemalloc` so that `mem-trace` can overload `malloc` since `jemalloc` is linked statically, so a `LD_PRELOAD`-ed library won't a chance to intercept the calls to `malloc`.
+
+### How to generate GIMPLE
+
+If you need to check what gccjit is generating (GIMPLE), then take a look at how to
+generate it in [gimple.md](./doc/gimple.md).
+
+### How to build a cross-compiling libgccjit
+
+To fix this, set the variables to `gcc-build/build/gcc`.
```
antoyo, line 49: This should be at the end of doc/errors.md.

doc/debugging-libgccjit.md
GuillaumeGomez: Good catch!

doc/tips.md
GuillaumeGomez: Wow indeed.

PR #445, opened 2024-02-17
Revert "Use shallow clone in test.rs to reduce cloning overhead"

GuillaumeGomez:
Reverts rust-lang/rustc_codegen_gcc#432

After further discussions with @antoyo, we came to the conclusion that rather than using extra complex commands, downloading the whole git history was taking more disk space but was simpler to maintain as a whole. Like I said, I'm not a big fan of `--depth` because of issues I encountered with it in the past too so at least it's always that off my mind.

Thanks in  any case @tempdragon!

### DIFF
--- a/build_system/src/test.rs
+++ b/build_system/src/test.rs
@@ -499,41 +499,16 @@ fn setup_rustc(env: &mut Env, args: &TestArg) -> Result<PathBuf, String> {
             &"clone",
             &"https://github.com/rust-lang/rust.git",
             &rust_dir_path,
-            &"--depth",
-            &"1",
         ],
         None,
         Some(env),
     );
     let rust_dir: Option<&Path> = Some(&rust_dir_path);
     run_command(&[&"git", &"checkout", &"--", &"tests/"], rust_dir)?;
+    run_command_with_output_and_env(&[&"git", &"fetch"], rust_dir, Some(env))?;
     let rustc_commit = match rustc_version_info(env.get("RUSTC").map(|s| s.as_str()))?.commit_hash {
         Some(commit_hash) => commit_hash,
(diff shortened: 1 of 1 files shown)


antoyo (approved):

antoyo:
@tempdragon: I guess you could do the git clone manually with `--depth 1` in the correct directory or do it once where you have good internet connection.

tempdragon:
Well, I found some better ways to handle the internet, so it is no longer a problem for me to access GitHub. However, this may still cause potential problems for 
I hope to offer something like an option in the script, for example one that can be manually specified (e.g. separate the downloading/cloning logic instead of the hard-coded download command) to give people potentially using a poor network connection a better choice, while using the clone as a default.
I do respect the conclusion despite (though I still wish to see better reasons or evidence of what happened exactly in the 
Could you please consider publishing the discussions so that I may potentially learn from the information provided?
Thank you all the same. @antoyo @GuillaumeGomez

tempdragon:
Well, the point is not gcc, but Rust. In the script it is possible to use a custom build of gcc, but doing a manual shallow clone and then fetching and checking out to the desired commit manually will still cause the script to re-clone the whole repo if you do not change the script. In a poor web condition, it is not possible to receive more than 10MiB of data from github every time you clone. That is the cause of original my commit.

antoyo:
Oh, sorry, I was confused.

You should be able to do a shallow clone the Rust repo manually: the script is supposed to do a `git fetch` and then a `git checkout`; if this doesn't work, let me know and we'll find a solution.

Or maybe the `git fetch` is the operation that is too slow and you would like to make it faster?
If so, would having a command to give you the git commit hash of the Rust repo be sufficient for your needs (so you could do the `git clone` manually for the right commit)?
If not, we could talk about having a command to clone the repo or something.

PR #446, opened 2024-02-17
fix tests/ui/simd/issue-89193.rs and mark as passing

sadlerap:
Work around an issue where usize and isize can sometimes (but not always) get canonicalized to their corresponding integer type.  This causes shuffle_vector to panic, since the types of the vectors it got passed aren't the same.

Also insert a cast on the mask element, since we might get passed a signed integer of any size, not just i32.  For now, we always cast to i32.

### DIFF
--- a/src/intrinsic/simd.rs
+++ b/src/intrinsic/simd.rs
@@ -710,7 +710,7 @@ pub fn generic_simd_intrinsic<'a, 'gcc, 'tcx>(
         };
         let elem_type = vector_type.dyncast_vector().expect("vector type").get_element_type();
 
-        let mut values = vec![];
+        let mut values = Vec::with_capacity(in_len as usize);
         for i in 0..in_len {
             let index = bx.context.new_rvalue_from_long(bx.i32_type, i as i64);
             let int = bx.context.new_vector_access(None, pointers, index).to_rvalue();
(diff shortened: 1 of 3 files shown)


antoyo:
Would changing [this line](https://github.com/antoyo/gcc/blob/ad4ffde712df49c958e325ebdd658c13f7865e30/gcc/jit/jit-recording.h#L992) to the following fix the issue:
```cpp
      && get_element_type ()->is_same_type_as (other_vec_type->get_element_type ())
```
?

I'm hoping we won't need the bitcast if we fix the type comparison in libgccjit.

sadlerap:
IMO we still need some kind of type checking within libgccjit.  Element types definitely need to be "close enough" for a vector shuffle to make any kind of sense semantically, and I think relaxing type restrictions might not be the right step.  I think asserting elements have the same size and alignment _might_ be acceptable, but I don't want to take out too much here.


Some other notes:

I noticed that [this function][type_uint_from_ty] returns the signed variants of types, when it might make sense to return the unsigned variant.  I tried this change, but ran into even more canonicalization issues (gcc thinks a vector of 4 u64s is distinct from a vector of 4 `size_t`s, which IMO is a correct assertion).  I suspect this is related to the issue.

I also tried constructing the vector from `default`, but I got assertions that `default`'s type wasn't a vector type?  Very strange.  To reproduce, add this line and run tests:
<p>

```diff
diff --git a/src/intrinsic/simd.rs b/src/intrinsic/simd.rs
index 33d659f..108bdb0 100644
--- a/src/intrinsic/simd.rs
+++ b/src/intrinsic/simd.rs
@@ -708,7 +708,7 @@ pub fn generic_simd_intrinsic<'a, 'gcc, 'tcx>(
         } else {
             vector_ty(bx, underlying_ty, in_len)
         };
-        let elem_type = vector_type.dyncast_vector().expect("vector type").get_element_type();
+        let elem_type = default.get_type().dyncast_vector().expect("vector type").get_element_type();
 
         let mut values = Vec::with_capacity(in_len as usize);
         for i in 0..in_len {

```

</p>
</details> 

[type_uint_from_ty]: https://github.com/rust-lang/rustc_codegen_gcc/blob/master/src/type_of.rs#L40-L50

src/intrinsic/simd.rs
```diff
@@ -723,19 +723,26 @@ pub fn generic_simd_intrinsic<'a, 'gcc, 'tcx>(
...
             let value = index + masked;
             mask_values.push(value);
         }
         let mask_type = bx.context.new_struct_type(None, "mask_type", &mask_types);
         let mask = bx.context.new_struct_constructor(None, mask_type.as_type(), None, &mask_values);
 
+        // FIXME(antoyo): We sometimes need to bitcast here, since usize/isize sometimes (but not
+        // always) get canonicalized to their corresponding integer type (i.e. uint64_t/int64_t on
+        // 64-bit platforms).  This causes the shuffle_vector call below to panic, since the types
+        // of the two vectors aren't the same.  This is a workaround for now.
```
antoyo, line 743: Is this `FIXME` still needed?

sadlerap: Nope, removed in 4ec4209.

sadlerap:
As of 4ec4209, we now use `default`'s type as the output type, and we extract the element type from there.

antoyo:
Thanks for your contribution!

Issue #428, opened 2024-02-12
Adding `rustc_private = true` to Global Cargo.toml to Allow Parsing in Some lsp-based IDE/editor Backends


tempdragon:
I noticed that on Emacs, the `rustc_codegen_gcc`  along with some other rustc-related libraries/projects can be correctly parsed with the following lines added to `Cargo.toml`,

```toml
[package.metadata.rust-analyzer]
rustc_private = true
```

according to [rust-analyzer manual](https://rust-analyzer.github.io/manual.html#configuration) (Search for "rustc_private" if you can't see it directly.).

Otherwise, the rust-analyzer will fail to parse with `unresolved external crate` for rustc-related crates.

Since there is probably other methods to allow the correct analysis in your IDE without this line, could you please give me some hints about how to have them parse the gcc backend correctly if you choose not to add these lines?

Thanks in advance.

antoyo:
Adding these lines doesn't fix the issue for me (I use neovim and nvm-lspconfig).

Maybe you could add this setting directly in your editor-specific LSP config until we find a proper fix that also works in neovim?

tempdragon:
In my case, it also requires setting the "Lsp Rust Analyzer Rustc Source" to `discover`. 
You can see here for the original answer to this question.
[Rust Forum](https://users.rust-lang.org/t/rust-analyzer-fails-to-index-due-to-unresolved-external-crate-in-a-rustc-private-project/105909/4)

sadlerap:
I got rust-analyzer working in neovim a while back.  I needed to do two things:
1. Set the lines above in `Cargo.toml`.
2. Set the lsp option `rust-analyzer.rustc.source = "discover"` (you can also set it to an explicit path).

For `nvim-lspconfig` in particular, you need to have the following:

```lua
require('lspconfig').rust_analyzer.setup{
  settings = {
    ['rust-analyzer'] = {
      -- add this section
      rustc = {
        source = "discover";
      },
    },
  },
}
```

[...]

PR #447, opened 2024-02-18
feat(Cargo.toml): Set `rustc_private` to `true` to allow lsp parsing

tempdragon:
Based on the comments in #428, it may be a good idea to add this option to `Cargo.toml` .
Hopefully I didn't misunderstand the suggestions from @antoyo & @sadlerap .
Thanks in advance.


antoyo:
Thanks for your contribution!

PR #453, opened 2024-02-21
Correctly ahdnle `--use-system-gcc` and `--backend` options

GuillaumeGomez:
Originally implemented in https://github.com/rust-lang/rust/pull/121390.


GuillaumeGomez:
It was done on rust side and synced afterwards, so no need to keep this open.

PR #455, opened 2024-02-24
feat(debuginfo): Add support for debuginfo, without scope support

tempdragon:
This PR depends on antoyo/gccjit.rs#28.


tempdragon:
TODO:
1. Clean the unnecessary locations in builder.rs & int.rs
2. Add demangling support
3. Add debug scope support
4. Add vtable support
5. Clean up builder.rs locations

antoyo:
Thank you so much! I can't wait to try this.

I'm perfectly fine with merging this PR as is, but I was wondering if you knew why rustc doesn't handle the debuginfo in the `builder.rs` module directly like is done here.
The idea I thought was maybe for performance reasons, since you can disable emitting debug info and might not want to create the objects for locations and names.
Do you have any ideas?

tempdragon:
# Short Answer
Of course I know it. In fact, it was my initial idea of applying the debuginfo.
Currently, it won't create extra overhead but a few loads of the None from `self.loc`, if the `self.sess().opts.debuginfo` is false.
# Warning
This looks miserable now, don't expect too much, and this is only a proof of concept/"draft" at this time.

tempdragon:
However, I do mean to have it tested using the CI since there are always suprises.

tempdragon:
To set the debuginfo afterwards, we first have to read the mapping from the debuginfo to the object from a separate function, this has a similar overhead to doing it in builder.rs.
Since location creation is disabled through scope generation in both cg_llvm and cg_gcc, no extra overhead is needed.

src/builder.rs
```diff
@@ -539,7 +543,13 @@ impl<'a, 'gcc, 'tcx> BuilderMethods<'a, 'tcx> for Builder<'a, 'gcc, 'tcx> {
     }
 
     fn fmul(&mut self, a: RValue<'gcc>, b: RValue<'gcc>) -> RValue<'gcc> {
-        a * b
+        let i=a * b;
```
antoyo, line 546: We might as well switch to `Context::new_binary_op()` to set the location directly.

src/base.rs
```diff
@@ -184,8 +184,8 @@ pub fn compile_codegen_unit(tcx: TyCtxt<'_>, cgu_name: Symbol, target_info: Lock
             // wrapper here
             maybe_create_entry_wrapper::<Builder<'_, '_, '_>>(&cx);
 
-            // Finalize debuginfo
-            if cx.sess().opts.debuginfo != DebugInfo::None {
+            // FINALIZE debuginfo
```
antoyo, line 187: Was there any reason to capitalize this word? If not, please revert this change.

src/base.rs
```diff
@@ -184,8 +184,8 @@ pub fn compile_codegen_unit(tcx: TyCtxt<'_>, cgu_name: Symbol, target_info: Lock
...
+            if cx.sess().opts.debuginfo != DebugInfo::None  {
```
antoyo, line 188: Please revert this change.

src/builder.rs
```diff
@@ -78,6 +70,7 @@ impl<'a, 'gcc, 'tcx> Builder<'a, 'gcc, 'tcx> {
             cx,
             block,
             stack_var_count: Cell::new(0),
+            loc:None
```
antoyo, line 73:
```suggestion
            loc: None
```

antoyo (commented):
Sending what I reviewed for now.

src/debuginfo.rs
```diff
@@ -51,25 +180,45 @@ impl<'gcc, 'tcx> DebugInfoMethods<'tcx> for CodegenCx<'gcc, 'tcx> {
...
         _file: &SourceFile,
     ) -> Self::DIScope {
-        unimplemented!();
+        //unimplemented!();
     }
 
     fn debuginfo_finalize(&self) {
-        // TODO(antoyo)
+        // TODO(antoyo): Get the debug flag/predicate to allow optional generation of debuginfo.
+        self.context.set_debug_info(true)
```
antoyo, line 221: It would be nice if you could handle this in this PR.

Please also remove [the call to `set_debug_info()` in `base.rs`](https://github.com/rust-lang/rustc_codegen_gcc/blob/master/src/base.rs#L155).

src/debuginfo.rs
```diff
@@ -89,15 +238,47 @@ impl<'gcc, 'tcx> DebugInfoMethods<'tcx> for CodegenCx<'gcc, 'tcx> {
         _fn_abi: &FnAbi<'tcx, Ty<'tcx>>,
         _maybe_definition_llfn: Option<RValue<'gcc>>,
     ) -> Self::DIScope {
-        unimplemented!();
+        //unimplemented!();
```
antoyo, line 241:
```suggestion
        // TODO(antoyo): implement.
```

src/debuginfo.rs
```diff
@@ -89,15 +238,47 @@ impl<'gcc, 'tcx> DebugInfoMethods<'tcx> for CodegenCx<'gcc, 'tcx> {
...
 
     fn dbg_loc(
         &self,
         _scope: Self::DIScope,
         _inlined_at: Option<Self::DILocation>,
-        _span: Span,
+        span: Span,
     ) -> Self::DILocation {
-        unimplemented!();
+        //unimplemented!();
```
antoyo, line 250:
```suggestion
```

src/debuginfo.rs
```diff
@@ -89,15 +238,47 @@ impl<'gcc, 'tcx> DebugInfoMethods<'tcx> for CodegenCx<'gcc, 'tcx> {
...
+        let pos = span.lo();
+        let (file, line, col) = self.lookup_debug_loc(pos);
+        let loc = match &file.name {
+            rustc_span::FileName::Real(name) => match name {
+                rustc_span::RealFileName::LocalPath(name) => {
+                    if let Some(name) = name.to_str() {
+                        self.context
+                            .new_location(name, line as i32, col as i32)
+                    }else{
```
antoyo, line 259:
```suggestion
                    } else{
```

src/debuginfo.rs
```diff
@@ -89,15 +238,47 @@ impl<'gcc, 'tcx> DebugInfoMethods<'tcx> for CodegenCx<'gcc, 'tcx> {
...
+                    if let Some(name) = name.to_str(){
+                        self.context.new_location(
+                            name,
+                            line as i32,
+                            col as i32,
+                        )
+                    } else {
+                        Location::null()
+                    }
+                }else{
```
antoyo, line 276:
```suggestion
                } else{
```

src/debuginfo.rs
```diff
@@ -89,15 +238,47 @@ impl<'gcc, 'tcx> DebugInfoMethods<'tcx> for CodegenCx<'gcc, 'tcx> {
...
     fn dbg_loc(
         &self,
         _scope: Self::DIScope,
         _inlined_at: Option<Self::DILocation>,
-        _span: Span,
+        span: Span,
     ) -> Self::DILocation {
-        unimplemented!();
+        //unimplemented!();
+        let pos = span.lo();
```
antoyo, line 251: Did you write this whole function or was it copied from the LLVM codegen? If it was copied, could you please tell me where it comes from?

src/lib.rs
```diff
@@ -174,7 +175,7 @@ impl CodegenBackend for GccCodegenBackend {
         crate::DEFAULT_LOCALE_RESOURCE
     }
 
-    fn init(&self, sess: &Session) {
+    fn init(&self, sess: &Session) {
```
antoyo, line 178:
```suggestion
    fn init(&self, sess: &Session) {
```

src/debuginfo.rs
tempdragon: The idea was from LLVM, but many data structures are not present in GCC. cg_llvm used a one(2?)-liner which is a simple call to the I guess this is how it should work and it turned out to be of some effect.

src/debuginfo.rs
```diff
@@ -1,41 +1,170 @@
...
+
+                // Use 1-based indexing.
+                let line = (line + 1) as u32;
+                let col = (file.relative_position(pos) - line_pos).to_u32() + 1;
+                (file,
+                 line,
+                 if ! self.sess().target.is_like_msvc {
+                     col } else {
+                     UNKNOWN_COLUMN_NUMBER
+                 }
```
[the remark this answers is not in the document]
tempdragon: Well, the code is actually not completely portable. Since the scope part is not currently supported(and I haven't found the corresponding structure & interface for a scope in GCC),  I have to currently separate the code from the original LLVM code. I never made it to a correct scope implementation. This will be fixed when the scope support is ready, or maybe never, if it is proved unnecessary.

src/debuginfo.rs
antoyo: Oh lol: I forgot I did that.

Is the `TODO` still necessary then?

tempdragon: I guess not, at least on my side.

> Please also remove [the call to `set_debug_info()` in `base.rs`](https://github.com/rust-lang/rustc_codegen_gcc/blob/master/src/base.rs#L155).

Done in eaeb54448e66edff57118bf9fe2ab3f13b98c04e.

antoyo: In that case, please remove the `TODO`.

src/debuginfo.rs
```diff
@@ -1,41 +1,170 @@
...
+        inlined_at,
+        file_start_pos: loc.0.start_pos,
+        file_end_pos: loc.0.end_position(),
+    };
+    instantiated.insert(scope);
+}
+
+/// Copied from LLVM backend
+impl<'gcc, 'tcx> CodegenCx<'gcc, 'tcx> {
+    pub fn lookup_debug_loc(&self, pos: BytePos) -> (Lrc<SourceFile>, u32, u32) {
```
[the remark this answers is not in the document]
tempdragon: Done in ef158f295ecd37413a11b9f01d203f7b2440cbe6

src/debuginfo.rs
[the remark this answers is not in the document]
tempdragon: In ef158f295ecd37413a11b9f01d203f7b2440cbe6, I added the original code but commented it as it is currently not in use for an empty implementation of `DIScope`.

src/debuginfo.rs
tempdragon: This was done in ef158f295ecd37413a11b9f01d203f7b2440cbe6(Sorry for forgetting to split it).

src/builder.rs
```diff
@@ -405,6 +398,15 @@ impl<'gcc, 'tcx> BackendTypes for Builder<'_, 'gcc, 'tcx> {
     type DIVariable = <CodegenCx<'gcc, 'tcx> as BackendTypes>::DIVariable;
 }
 
+pub fn set_rval_location<'a, 'gcc, 'tcx>(bx: &mut Builder<'a,'gcc,'tcx>, r:RValue<'gcc>) -> RValue<'gcc> {
+    if bx.loc.is_some(){
+        #[cfg(feature = "master")]
+        r.set_location(bx.loc.unwrap());
+    }
+    r
+
```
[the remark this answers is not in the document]
antoyo: I don't see the variable renaming: did you forget to push the change?

src/builder.rs
```diff
@@ -653,73 +655,75 @@ impl<'a, 'gcc, 'tcx> BuilderMethods<'a, 'tcx> for Builder<'a, 'gcc, 'tcx> {
...
 
     fn unchecked_umul(&mut self, a: RValue<'gcc>, b: RValue<'gcc>) -> RValue<'gcc> {
-        self.gcc_mul(a, b)
+        set_rval_location(self,self.gcc_mul(a, b))
     }
 
     fn fadd_fast(&mut self, lhs: RValue<'gcc>, rhs: RValue<'gcc>) -> RValue<'gcc> {
         // NOTE: it seems like we cannot enable fast-mode for a single operation in GCC.
-        lhs + rhs
+        set_rval_location(self,lhs + rhs)
```
antoyo, line 704: A space is missing after the `self,` for all calls to `set_rval_location`.

src/builder.rs
```diff
@@ -1002,32 +1006,32 @@ impl<'a, 'gcc, 'tcx> BuilderMethods<'a, 'tcx> for Builder<'a, 'gcc, 'tcx> {
...
 
     fn fptrunc(&mut self, value: RValue<'gcc>, dest_ty: Type<'gcc>) -> RValue<'gcc> {
         // TODO(antoyo): make sure it truncates.
-        self.context.new_cast(None, value, dest_ty)
+        set_rval_location(self,self.context.new_cast(self.loc, value, dest_ty))
     }
 
     fn fpext(&mut self, value: RValue<'gcc>, dest_ty: Type<'gcc>) -> RValue<'gcc> {
-        self.context.new_cast(None, value, dest_ty)
+        set_rval_location(self,self.context.new_cast(self.loc, value, dest_ty))
```
antoyo, line 1034: A space is missing after `self,` in all calls to `set_rval_location`.

src/debuginfo.rs
```diff
@@ -1,41 +1,216 @@
...
+                debug_context.scopes[scope] = parent_scope;
+                instantiated.insert(scope);
+                return;
+            }
+    }
+
+    let loc = cx.lookup_debug_loc(scope_data.span.lo());
+
+    /*
+    // FIXME(?): Uncommented when the scope is supported.
```
antoyo, line 137: I believe upstream will not want commented code like this, so please remove. Sorry if you added it because of my comments.

src/builder.rs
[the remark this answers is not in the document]
tempdragon: Wierd? Sorry. I guess I accidentally folded this.
Fixed in 8c975d9

src/debuginfo.rs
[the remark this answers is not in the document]
antoyo: I'm not sure I understand the difference compared with the code from cg_llvm. It just seems to me that the condition `is_like_msvc` was moved from after the `match` to inside it.

And that the return type was changed from a struct to a tuple containing the same information. Could you please explain to me how this is different?

antoyo (commented):
Some more nits.

src/debuginfo.rs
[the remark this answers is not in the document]
tempdragon: Short answer: It's not that different. I will switch back to the original implementation.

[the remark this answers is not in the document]
tempdragon: I originally wrote this stuff by hand and end up noticing the function being reusable. So I switched to their implementation. I choose a tuple in place of a struct because it seems to me that it is easier to be done not defining a data structure for something that is used only a few times internally.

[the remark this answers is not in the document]
antoyo: Yeah, I plan to attempt to deduplicate some code between cg_gcc and cg_llvm in the future. Not sure if this specific function can be deduplicated, but I guess a couple a other functions are.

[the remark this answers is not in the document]
tempdragon: Maybe this should be a good idea to do it upstream, but it also depends on the attitude of developers of other backends such as those of cranelift.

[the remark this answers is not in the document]
antoyo: Yes, it's the plan. Cranelift doesn't use the SSA traits as far as I know.

src/builder.rs
tempdragon: Should have been fixed in7c3565e569bb5017baad3a2a708dcdf7c1822d82.

src/debuginfo.rs
tempdragon: Done in 9cc0a4204d1d1dba72b79d009e5f526505ea2d3c.

src/builder.rs
tempdragon: Should have been fixed in7c3565e569bb5017baad3a2a708dcdf7c1822d82.

PR #468, opened 2024-03-10
Sync Rust 2024-03-10

GuillaumeGomez:


build_system/src/config.rs
```diff
@@ -377,39 +386,25 @@ impl ConfigInfo {
...
+            .join(channel)
+            .join(&format!("librustc_codegen_gcc.{}", self.dylib_ext))
+            .display()
+            .to_string();
+        self.sysroot_path = current_dir
+            .join("build_sysroot/sysroot")
+            .display()
+            .to_string();
+        if let Some(backend) = &self.backend {
+            rustflags.push(format!("-Zcodegen-backend={}", backend));
```
antoyo, line 401: You don't need sysroot here?

rust-toolchain
```diff
@@ -1,3 +1,3 @@
 [toolchain]
-channel = "nightly-2024-03-05"
-components = ["rust-src", "rustc-dev", "llvm-tools-preview"]
+channel = "nightly-2024-03-10"
+components = ["rust-src", "rustc-dev", "llvm-tools-preview", "rustfmt"]
```
antoyo, line 3: Why add `rustfmt` here?

GuillaumeGomez: Because we need it to run `cargo fmt` (which is enforced in the CI).

build_system/src/config.rs
GuillaumeGomez: Apparently not. This option is only used in the Rust CI (`--backend`),

rust-toolchain
antoyo: Yeah, but the people who only want to compile the project might not want it.
So, I would remove this unless you see a reason to keep it.

build_system/src/config.rs
antoyo: Ah makes sense since you compile the sysroot the usual way I guess.
Could you please add a comment to mention this?

GuillaumeGomez: Sure!

rust-toolchain
GuillaumeGomez: Hum ok, removing it.

PR #470, opened 2024-03-11
Remove `master` feature

GuillaumeGomez:
Since we now download our own GCC version, there is no need to keep supporting the GCC 12 version anymore. It also simplifies the code quite a lot. :)


antoyo:
We should keep this at least for another 6 months/1 year in case we still need it since it would be a pain to add back if removed.

PR #474, opened 2024-03-17
feat(doc): Add a doc for beginners

tempdragon:
1. Info on How to Enble Enable `rust-analyzer` Support
2. Info on How to Compile External Project with `cg_gcc`

### DIFF
--- a/doc/for-beginners.md
+++ b/doc/for-beginners.md
@@ -1,0 +1,80 @@
+# Enable full `rust-analyzer` support of `cg_gcc` for your editor
+
+## Steps to Follow
+1. (Already done in by project.) Add the following lines to `Cargo.toml`
+```toml
+[package.metadata.rust-analyzer]
+rustc_private = true
+```
+2. Install rust-src with rustup: `rustup component add rust-src`.
+   NOTE: Make sure you've switched to the corresponding toolchain as is used by cg_gcc
+3. Set the lsp option `rust-analyzer.rustc.source = "discover"` (you can also set it to an explicit path), depending on your editor.
+    1. Neovim(nvim-lspconfig):
+    ```
+    require('lspconfig').rust_analyzer.setup{
+      settings = {
+        ['rust-analyzer'] = {
+          -- add this section
+          rustc = {
+            source = "discover",
+          },
+        },
+      },
+    }
+    ```
+   2. Emacs: customize `Lsp Rust Analyzer Rustc Source` to the Path `discover`
+
+## FAQ
+
+### Why is this needed?
+
+Without these steps, `rust-analyzer` will NOT import `rustc` related packages to its scope and all rustc related clauses will fail to be analyzed in lsp, and the analyzer will notify you of the missing `rustc_*` packages.
+
+### Why is this NOT needed for `rustc_codegen_llvm` in your `rust-src`?
+In the `Cargo.toml` of the `rustc_codegen_llvm`, you may notice the following lines:
+```toml
+rustc_ast = { path = "../rustc_ast" }
+rustc_attr = { path = "../rustc_attr" }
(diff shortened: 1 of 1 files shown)


tempdragon:
Maybe it is possible to map the `doc` directory to the wiki(and its corresponding repo) and allow workflow to automatically update it.
It should also be possible to use the wiki as a `submodule` but potential misinformation or versions that are not carefully reviewed may cause problems

doc/for-beginners.md
```diff
@@ -0,0 +1,80 @@
...
+[package.metadata.rust-analyzer]
+rustc_private = true
+```
+
+# Compiling External Project with `rustc_codegen_gcc`
+
+1.  Run the program with (in the root dir of `rustc_codegen_gcc` )(where `--manifest-path` is the path to the `Cargo.toml` of the project to run `cargo` against):
+
+```shell
+    ./y.sh cargo --manifest-path="..."
```
GuillaumeGomez, line 70: No need for the indent.

doc/for-beginners.md
```diff
@@ -0,0 +1,80 @@
...
+```
+
+or you may as well add this alias in your `.bash_aliases`:
+
+
+```shell
+    alias cargcc="/path/to/rustc_codegen_gcc/y.sh cargo"
```
GuillaumeGomez, line 77: No need for the indent.

doc/for-beginners.md
```diff
@@ -0,0 +1,80 @@
...
+
+# Compiling External Project with `rustc_codegen_gcc`
+
+1.  Run the program with (in the root dir of `rustc_codegen_gcc` )(where `--manifest-path` is the path to the `Cargo.toml` of the project to run `cargo` against):
+
+```shell
+    ./y.sh cargo --manifest-path="..."
+```
+
+or you may as well add this alias in your `.bash_aliases`:
```
GuillaumeGomez, line 73: I'd drop this whole part. I never used `.bash_aliases`, only `.bashrc` for example. Having something so specific seems a bit too much imo.

doc/for-beginners.md
```diff
@@ -0,0 +1,80 @@
...
+## Steps to Follow
+1. (Already done in by project.) Add the following lines to `Cargo.toml`
+```toml
+[package.metadata.rust-analyzer]
+rustc_private = true
+```
+2. Install rust-src with rustup: `rustup component add rust-src`.
+   NOTE: Make sure you've switched to the corresponding toolchain as is used by cg_gcc
+3. Set the lsp option `rust-analyzer.rustc.source = "discover"` (you can also set it to an explicit path), depending on your editor.
+    1. Neovim(nvim-lspconfig):
```
GuillaumeGomez, line 12: Please drop this part, there are a lot of IDEs, adding docs for some of them is not useful.

tempdragon: I wrote this part for  [this comment.](https://github.com/rust-lang/rustc_codegen_gcc/issues/428#issuecomment-1949630003)

tempdragon: In my point of view, it may be trivial for us that these steps are necessary, but probably harder for others to figure it out. Adding this may potentially add to contributors(since we only have no more than 100 developers).

doc/for-beginners.md
antoyo: Since I find it very useful to have an alias, I'd keep this part.
I didn't know about `.bash_aliases` either: I'll let the author rename that to `.bashrc` if they find that appropriate.

doc/for-beginners.md
antoyo: I'm OK with keeping this part since it's useful and people can propably understand what they need to do for their own IDE from these instructions.

antoyo (commented):
Sorry for the long delay in reviewing. I forgot about this PR.
(Feel free to ping me what that happens.)

PR #475, opened 2024-03-18
Fix configure path in the README.md instructions

tgross35:
Using `gcc/configure` gives me a "no rule to make target"  error, the same as https://gcc.gnu.org/bugzilla/show_bug.cgi?id=25543. Using `/configure` seems to work correctly, hence this change.

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -36,7 +36,7 @@ $ git clone https://github.com/antoyo/gcc
 $ sudo apt install flex libmpfr-dev libgmp-dev libmpc3 libmpc-dev
 $ mkdir gcc-build gcc-install
 $ cd gcc-build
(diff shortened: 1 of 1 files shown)


GuillaumeGomez (approved):

antoyo:
This change seems wrong to me since GCC should be built out-of-tree.

Can you show me your directory structure please?

tgross35:
On second look, I misread the instructions and created `gcc-build` and `gcc-install` within the `gcc` clone, while it seems like they were supposed to be adjacent. So `../gcc/configure` was picking up `configure` in this directory https://github.com/rust-lang/gcc/tree/78dc50f0e50e6cd1433149520bd512a4e0eaa1bc/gcc

No need for this then.

Issue #11, opened 2021-05-29
fail during ./prepare.sh


dkm:
```
$ ./prepare.sh && ./test.sh --release
#!/bin/bash --verbose
set -e

rustup component add rust-src rustc-dev llvm-tools-preview
info: syncing channel updates for 'nightly-2021-05-18-x86_64-unknown-linux-gnu'
info: latest update on 2021-05-18, rust version 1.54.0-nightly (3e99439f4 2021-05-17)
warning: Force-skipping unavailable component 'miri-x86_64-unknown-linux-gnu'
info: downloading component 'cargo'
info: downloading component 'clippy'
info: downloading component 'llvm-tools-preview'
 21.4 MiB /  21.4 MiB (100 %)   7.4 MiB/s in  3s ETA:  0s
info: downloading component 'rls'
  8.2 MiB /   8.2 MiB (100 %)   1.8 MiB/s in  2s ETA:  0s
info: downloading component 'rust-analysis'
  2.7 MiB /   2.7 MiB (100 %)   1.8 MiB/s in  1s ETA:  0s
info: downloading component 'rust-analyzer-preview'
  6.7 MiB /   6.7 MiB (100 %)   1.8 MiB/s in  2s ETA:  0s
info: downloading component 'rust-docs'
 16.6 MiB /  16.6 MiB (100 %)   6.2 MiB/s in  2s ETA:  0s
info: downloading component 'rust-src'
info: downloading component 'rust-std'
 21.8 MiB /  21.8 MiB (100 %)   7.5 MiB/s in  3s ETA:  0s
info: downloading component 'rustc'
 48.8 MiB /  48.8 MiB (100 %)   7.2 MiB/s in  7s ETA:  0s
info: downloading component 'rustc-dev'
 96.8 MiB /  96.8 MiB (100 %)   6.7 MiB/s in 15s ETA:  0s
info: downloading component 'rustc-docs'
 12.0 MiB /  12.0 MiB (100 %)   4.3 MiB/s in  3s ETA:  0s
info: downloading component 'rustfmt'
info: installing component 'cargo'
info: installing component 'clippy'
info: installing component 'llvm-tools-preview'
 21.4 MiB /  21.4 MiB (100 %)  13.2 MiB/s in  1s ETA:  0s
info: installing component 'rls'
info: installing component 'rust-analysis'
info: installing component 'rust-analyzer-preview'
info: installing component 'rust-docs'
[...]

PR #484, opened 2024-04-03
Sync rust 25-03-2024

GuillaumeGomez:
Other take to #11.

### DIFF
--- a/src/builder.rs
+++ b/src/builder.rs
@@ -630,8 +630,9 @@ impl<'a, 'gcc, 'tcx> BuilderMethods<'a, 'tcx> for Builder<'a, 'gcc, 'tcx> {
         then: Block<'gcc>,
         catch: Block<'gcc>,
         _funclet: Option<&Funclet>,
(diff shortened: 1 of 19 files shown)


src/builder.rs
```diff
@@ -2422,3 +2424,19 @@ impl ToGccOrdering for AtomicOrdering {
...
+#[cfg(feature = "master")]
+fn get_maybe_pointer_size(value: RValue<'_>) -> u32 {
+    value.get_type().get_size()
+}
+
+#[cfg(not(feature = "master"))]
+fn get_maybe_pointer_size(value: RValue<'_>) -> u32 {
+    let type_ = value.get_type();
+    if type_.get_pointee().is_some() {
+        std::mem::size_of::<*const ()>() as _
```
antoyo, line 2438: This is not going to work from cross-compilation.
It would be better to ask the pointer size of the target through some rustc API.

GuillaumeGomez:
Just realized it's not the right target. ^^'

Issue #491, opened 2024-04-17
Cannot pass CG_RUSTFLAGS when building sysroot


darcagn:
I've successfully gotten libc/std building targetting SuperH (SH4/Dreamcast). I need to build sysroot with `-matomic-model=soft-imask` for it to work properly. 

I can insert `rustflags.push_str(" -Cllvm-args=-matomic-model=soft-imask");` at this line:
https://github.com/rust-lang/rustc_codegen_gcc/blob/5584f5f1a22aceb3d686cfa1165c6d160c994586/build_system/src/build.rs#L150

Or insert `context.add_command_line_option("-matomic-model=soft-imask");` at this line:
https://github.com/rust-lang/rustc_codegen_gcc/blob/5584f5f1a22aceb3d686cfa1165c6d160c994586/src/base.rs#L118

And I seem to get the desired effect. But I'd prefer not to patch things if I can do it without. 

If I do `CG_RUSTFLAGS=-Cllvm-args=-matomic-model=soft-imask y.sh build` it does not seem to honor it when building the sysroot. To test this I can put gibberish flags like `-blahblah` and get no error when the same would give me errors when building a normal project with `y.sh cargo build`. Am I doing something wrong here or is this something not supported yet?

GuillaumeGomez:
Normally we do use `CG_RUSTFLAGS` as you can see [here](https://github.com/rust-lang/rustc_codegen_gcc/blob/5584f5f1a22aceb3d686cfa1165c6d160c994586/build_system/src/config.rs#L381-L383). This function is called to build the command line which will be used for all commands, so if it's not setting variables correctly, maybe the code is wrong.

antoyo:
In this case, the environment is first set to the empty one [here](https://github.com/rust-lang/rustc_codegen_gcc/blob/5584f5f1a22aceb3d686cfa1165c6d160c994586/build_system/src/build.rs#L177), so that condition is never true.

antoyo:
Fixed by #493.

PR #493, opened 2024-04-17
Fix passing custom CG_RUSTFLAGS when building sysroot

darcagn:
Fixes issue #491

### DIFF
--- a/build_system/src/build.rs
+++ b/build_system/src/build.rs
@@ -153,6 +153,11 @@ pub fn build_sysroot(env: &HashMap<String, String>, config: &ConfigInfo) -> Resu
         "debug"
     };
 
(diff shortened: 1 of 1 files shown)


build_system/src/build.rs
```diff
@@ -153,6 +153,11 @@ pub fn build_sysroot(env: &HashMap<String, String>, config: &ConfigInfo) -> Resu
         "debug"
     };
 
+    if let Ok(cg_rustflags) = std::env::var("CG_RUSTFLAGS") {
```
GuillaumeGomez, line 156: Ah so the problem is that we don't use `config.rustflags` here. I see. :+1:

GuillaumeGomez (approved):
Looks good to me, thanks!

antoyo (approved):

antoyo:
Thanks for your contribution!

PR #494, opened 2024-04-17
Modify build_system's prepare stage to allow for custom sysroot source path

darcagn:
This change allows users to specify a custom directory from which to copy the sysroot source code instead of being forced to use the `rust-src` provided with the `rustup` toolchain. The path is passed to `--sysroot-source` during the `prepare` stage, like so:

```
./y.sh prepare --sysroot-source /path/to/custom/source
```

This removes a pain point for using rustc_codegen_gcc while creating and testing support for new platforms with architectures without LLVM support.

### DIFF
--- a/build_system/src/prepare.rs
+++ b/build_system/src/prepare.rs
@@ -5,29 +5,41 @@ use crate::utils::{
 };
 
 use std::fs;
-use std::path::Path;
+use std::path::{Path, PathBuf};
 
 fn prepare_libcore(
     sysroot_path: &Path,
     libgccjit12_patches: bool,
     cross_compile: bool,
+    sysroot_source: Option<String>,
 ) -> Result<(), String> {
-    let rustc_path = match get_rustc_path() {
-        Some(path) => path,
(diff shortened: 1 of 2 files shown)


build_system/src/prepare.rs
```diff
@@ -5,29 +5,39 @@ use crate::utils::{
...
+    let rustlib_dir: PathBuf;
 
-    let parent = match rustc_path.parent() {
-        Some(path) => path,
-        None => return Err(format!("No parent for `{}`", rustc_path.display())),
-    };
+    if let Some(path) = sysroot_source {
+        rustlib_dir = path.into();
+        if !rustlib_dir.is_dir() {
+            return Err("Custom sysroot path not found".to_string());
```
antoyo, line 21: Please also print the path.

antoyo (approved):
One small thing to fix.

antoyo:
@darcagn: I would say: choose the approach that is the most convenient to you since you're the main user of this new flag.

antoyo:
Thanks for your contribution!

PR #509, opened 2024-04-30
Stop swallowing signals in build_system when running sub-commands

### LINKED ISSUE #496: y.sh cargo run swallows the signal emitted by the running process
For instance, if the process segfaults, `./y.sh cargo run` will not emit a SIGSEGV signal while it should do that.

GuillaumeGomez:
Fixes #496.

### DIFF
--- a/build_system/src/utils.rs
+++ b/build_system/src/utils.rs
@@ -1,9 +1,41 @@
 use std::collections::HashMap;
+#[cfg(unix)]
+use std::ffi::c_int;
 use std::ffi::OsStr;
 use std::fmt::Debug;
 use std::fs;
+#[cfg(unix)]
+use std::os::unix::process::ExitStatusExt;
 use std::path::{Path, PathBuf};
 use std::process::{Command, ExitStatus, Output};
+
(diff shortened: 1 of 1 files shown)


build_system/src/utils.rs
```diff
@@ -2,9 +2,30 @@ use std::collections::HashMap;
...
+) -> Result<ExitStatus, String> {
+    let status = get_command_inner(input, cwd, env)
+        .spawn()
+        .map_err(|e| command_error(input, &cwd, e))?
+        .wait()
+        .map_err(|e| command_error(input, &cwd, e))?;
+    #[cfg(unix)]
+    {
+        if let Some(signal) = status.signal() {
+            return Err(command_error(input, &cwd, format!("Process received signal {}", signal)));
```
antoyo, line 23: I would prefer that you follow the same behavior as `cargo`, which is to emit the signal of the process so that the shell sees it.

GuillaumeGomez: Fine by me.

antoyo (approved):

PR #510, opened 2024-05-02
Add `fmt` command

GuillaumeGomez:


build_system/src/fmt.rs
```diff
@@ -0,0 +1,35 @@
...
+                show_usage();
+                return Ok(());
+            }
+            "--check" => check = true,
+            _ => return Err(format!("Unknown option {}", arg)),
+        }
+    }
+
+    let cmd: &[&dyn AsRef<OsStr>] =
+        if check { &[&"cargo", &"fmt", &"--check"] } else { &[&"cargo", &"fmt"] };
```
antoyo, line 31: Shouldn't we just forward the arguments to `cargo fmt` directly instead of handling this ourselves?

build_system/src/fmt.rs
```diff
@@ -0,0 +1,35 @@
+use crate::utils::run_command_with_output;
+use std::ffi::OsStr;
+use std::path::Path;
+
+fn show_usage() {
+    println!(
+        r#"
+`test` command help:
```
antoyo, line 8: You meant?
```suggestion
`fmt` command help:
```

build_system/src/fmt.rs
GuillaumeGomez: Well, we only want `--check` and `--help`, so better handle the available options ourselves imo.

build_system/src/fmt.rs
GuillaumeGomez: Woups.

GuillaumeGomez:
Updated!

antoyo (approved):

PR #519, opened 2024-05-21
simd: implement pointer provenance intrinsics

sadlerap:
This adds support for the simd variants of the pointer provenance intrinsics, which are `simd_cast_ptr`, `simd_expose_addr`, and `simd_from_exposed_addr`.

The preconditions for each intrinsic are adapted from `rustc_codegen_llvm` to preserve compatibility.  Each of these intrinsics are implemented as calling the non-simd variant of each intrinsic on each lane.

This is enough to enable the UI test `ui/simd/intrinsic/ptr-cast.rs` to pass.


GuillaumeGomez:
Thanks! Let's wait for @antoyo to take a look as well. :)

antoyo:
Thank you for your contribution!

Issue #495, opened 2024-04-19
Run the error UI tests in a new CI job


antoyo:
Since some of them has an ICE: https://github.com/rust-lang/rustc_codegen_gcc/issues/296

We currently remove the tests that emits an error in the frontend [here](https://github.com/rust-lang/rustc_codegen_gcc/blob/master/build_system/src/test.rs#L837-L842).
I would like to have a new test that will run those tests (and only these tests with errors) in `failures.yml`.

mubarak23:
@antoyo can you further explain this, i will like to handle it

lordshashank:
@antoyo what do you want to be done here, currently failing rust test suits are ran through ```--test-failing-rustc```  [here](https://github.com/antoyo/rustc_codegen_gcc/blob/d7c8e0fb434144b441e48f79eb2d728b1eab298a/.github/workflows/failures.yml#L99), do you want ui tests to have separate run command like `--test-failing-ui-tests`, and run it under separate name in failures.yaml or something else?

antoyo:
I edit the description to explain more.

lordshashank:
Got it, can work this out now, thanks

antoyo:
I assigned the issue to you.

PR #524, opened 2024-05-26
ui pattern failure tests

lordshashank:
- resolves https://github.com/rust-lang/rustc_codegen_gcc/issues/495
- implements function for specific ui pattern failure tests [here ](https://github.com/rust-lang/rustc_codegen_gcc/blob/master/build_system/src/test.rs#L837-L842)
- Run this test as ci job in failures.yml
- Refactored various function names as per functionality.
- Added comments of parms of `test_rustc_inner`


lordshashank:
@antoyo after ideating for hours on how to proceed and looking for various options like implementing completely new function for this, including simple bool param in test_rustc_inner for this, making global bool variable to include/exclude these tests, etc.
I thought best option would be to include dynamic boxed closure as parameter to test_rustc_inner function, this would enable us to implement separate tests/ci jobs for other checks pattern in different test suites if we implement in future. 
Let me know if you have any better way and if this didn't worked correctly, thanks.

antoyo:
To make this useful, we would need the test to fail when there's an ICE (internal compiler error).

We can see in the logs that some tests have an ICE:
```
2024-05-26T23:54:45.8957411Z thread 'rustc' panicked at compiler/rustc_trait_selection/src/traits/select/confirmation.rs:795:50:
2024-05-26T23:54:45.8957573Z future has no bound vars
2024-05-26T23:54:45.8957837Z note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
2024-05-26T23:54:45.8957843Z 
2024-05-26T23:54:45.8958035Z error: the compiler unexpectedly panicked. this is a bug.
```

Maybe we could grep for `the compiler unexpectedly panicked` and fail the test in this case.
We would then need a file for to ignore the tests that are known to produce an ICE.

It would also be interesting to check if this is a good strategy. Perhaps some tests also produce an ICE with the LLVM codegen. Perhaps there is an annotation for the tests that produce an ICE even with LLVM? Perhaps it is [`known-bug`](https://rustc-dev-guide.rust-lang.org/tests/ui.html#known-bugs)?

lordshashank:
@antoyo where are these logs from? From workflow run I can see that around 474 ui-pattern-failure-test passed. Do you mean that some of them have an ICE and we should fail them manually?

antoyo:
To see those logs, I went to the [failure workflow](https://github.com/rust-lang/rustc_codegen_gcc/actions/runs/9247012603/job/25435212736?pr=524) and clicked on "View raw logs" (because the logs are too long to be shown on the previous page).

antoyo:
Yeah, we should fail the CI when there's an ICE, but please consider the notes I mentioned above.

lordshashank:
@antoyo what do you mean by 
> we would need the test to fail when there's an ICE
all the cases for ICE (with "the compiler unexpectedly panicked") that I see in logs are for the ui tests that are already failing.

antoyo:
Indeed, but the overall CI task do not fail because we eat the exit code.
I would want the CI task to fail when an ICE is detected (except if it is expected: see notes above).

lordshashank:
@antoyo I checked few of them manually, could get "known-bug" in the one you shared above only. Could you tell me how can I grep for `the compiler unexpectedly panicked` to get the tests that are giving this as all this is being run through `exec_command` function, not sure how to get logs to grep this.

antoyo:
You can see the logs by going to [this page](https://github.com/rust-lang/rustc_codegen_gcc/actions/runs/9247012603/job/25435212736?pr=524), then click on the gear at the top right and either select "View raw logs" or "Download log archive".
You need to do this since the logs are too long to be shown in the page.

lordshashank:
Ah! I already did that, is there only manual way to get the tests with ICE from finding the logs, I was thinking of some programmatic way. (there are many, approx 100 such cases in logs)

antoyo:
You mean, finding the name of those tests?

lordshashank:
yes, cause then only i would be able to create file to omit them or check for "known-bug" right?

antoyo:
Yes.

I do not know any programmatic way to do this.

@GuillaumeGomez mentioned to me the results are in JSON at some point, but we don't know if we could get the JSON.

GuillaumeGomez:
To give more details: there is the tester which runs ui test with the `--output-format json`, then parses this output and checks if it's expected or not. I don't think this tester itself has an option to output json though.

lordshashank:
Ok, so had to do a lot here.
First I looked for the ICE failures using script on logs downloaded, checked for `known bug` found just one case you shared. Was reluctant to check for any other error pattern as didn't seem universal here, and could have led to removal of some successfull test as once happened before.
Next, while getting `the compiler unexpectedly panicked` found one case in `run-make` tests as well [here](https://github.com/rust-lang/rustc_codegen_gcc/blob/eccab8ba39e9c5a3dfcf146105909039b64b0577/tests/failing-run-make-tests.txt#L8). But I guess this was expected here. 
Then, had do callback for tests, improved some function naming there (let me know if thats an issue).
For CI failure, had to grep from logs, found no other way, let me know if you have one.
Also, @antoyo I mistakingly raised PR to this branch rather than yours, let me know if I have close this and raise there.
Huh 😅
Will look why the runs failed next.
Thanks.

lordshashank:
@antoyo ICE tests should get removed through the callback [here](https://github.com/lordshashank/rustc_codegen_gcc/blob/94d6b2338576e0108848c8708d128ab319dd5ef2/build_system/src/test.rs#L1093), I tried to log them [here](https://github.com/lordshashank/rustc_codegen_gcc/blob/94d6b2338576e0108848c8708d128ab319dd5ef2/build_system/src/test.rs#L912) and the failing tests were not there. Still the CI run took those tests leading to failing of CI. Could you help me know what am I missing here?

antoyo:
From what I can see, the command `./y.sh test --release --clean --build-sysroot --test-failing-ui-pattern-tests` will not run the tests from the file `failing-ice-tests.txt`. If we still see failures of tests like `tests/ui/sepcomp/sepcomp-statics.rs`, they happen before we run this command.

In fact, we see them in the command `./y.sh test --release --clean --build-sysroot --test-failing-rustc` because the file `tests/failing-ui-tests.txt` lists those same tests `tests/ui/sepcomp/sepcomp-statics.rs`.

(You can search for "build system" in the output to see the command that is ran about 14 lines above "build system".)

> I tried to log them [here](https://github.com/lordshashank/rustc_codegen_gcc/blob/94d6b2338576e0108848c8708d128ab319dd5ef2/build_system/src/test.rs#L912) and the failing tests were not there

For the record, this is normal that you don't see them here because they were removed by the call to `prepare_files_callback_remove`.

All of that to say, the CI fails because some other tests (for instance `tests/ui/extern/issue-64655-allow-unwind-when-calling-panic-directly.rs` gives an ICE), but these tests are not removed (currenty because the callback `should_run_test_callback` returns true), so I believe you need to add these tests to the file `failing-ice-tests.txt`.

lordshashank:
ok, so it was that. Actually I looked at logs and saw same files that I omitted in failing-ice-tests.txt , thus thought they were not getting omitted, but actually they were from another ci task, sorry.
I guess CI has passed but I got a mail of it failing, not sure why. @antoyo let me know if anything else is needed here.

build_system/src/test.rs
```diff
@@ -1030,18 +1073,30 @@ fn test_successful_rustc(env: &Env, args: &TestArg) -> Result<(), String> {
...
     )
 }
 
-fn prepare_files_callback_failing<'a>(
+fn test_failing_ui_pattern_tests(env: &Env, args: &TestArg) -> Result<(), String> {
+    test_rustc_inner(
+        env,
+        args,
+        prepare_files_callback_remove("tests/failing-ice-tests.txt", "ui"),
+        Some(Box::new(|path| should_remove_test(path).unwrap_or(false))),
```
antoyo, line 1094: Can you please explain why this is necessary?

lordshashank: >   Some(Box::new(|path| should_remove_test(path).unwrap_or(false))),

I guess you are talking of this, as stated on first comment
>@antoyo after ideating for hours on how to proceed and looking for various options like implementing completely new function for this, including simple bool param in test_rustc_inner for this, making global bool variable to include/exclude these tests, etc.
I thought best option would be to include dynamic boxed closure callback function as parameter to test_rustc_inner function, this would enable us to implement separate tests/ci jobs for other checks pattern in different test suites if we implement in future.

basically as we currently wanted tests from `should_remove_test()` function to run (i.e. those followed the error patterns), in future we may have different other test suites with other function like `should_remove_test()`, then this callback would help in generalization.
Another way could be to just make another callback function that runs those tests, or if you have something better in mind let me know.

antoyo: I'm still not sure I understand.
We have the file `failing-ice-tests.txt` that tells which files to remove, so why do we need a callback in addition to this file?

.github/workflows/failures.yml
```diff
@@ -98,3 +98,13 @@ jobs:
...
+
+    - name: Run failing ui pattern tests for ICE
+      id: ui-tests
+      run: |
+        ${{ matrix.libgccjit_version.env_extra }} ./y.sh test --release --clean --build-sysroot --test-failing-ui-pattern-tests ${{ matrix.libgccjit_version.extra }} | tee output_log_ui
+        if grep -q "the compiler unexpectedly panicked" output_log_ui; then
+          echo "Error: 'the compiler unexpectedly panicked' found in output logs. CI Error!!"
+          exit 1
+        fi
+        rg --text "test result" output_log_ui >> $GITHUB_STEP_SUMMARY
```
antoyo, line 110: Please remove this line: I don't think it's necessary to see the results of this new step in the summary.

.github/workflows/failures.yml
```diff
@@ -98,3 +98,13 @@ jobs:
       run: |
         ${{ matrix.libgccjit_version.env_extra }} ./y.sh test --release --clean --build-sysroot --test-failing-rustc ${{ matrix.libgccjit_version.extra }} | tee output_log
         rg --text "test result" output_log >> $GITHUB_STEP_SUMMARY
+
+    - name: Run failing ui pattern tests for ICE
+      id: ui-tests
+      run: |
+        ${{ matrix.libgccjit_version.env_extra }} ./y.sh test --release --clean --build-sysroot --test-failing-ui-pattern-tests ${{ matrix.libgccjit_version.extra }} | tee output_log_ui
```
antoyo, line 105: Removing `--clean --build-sysroot` will make the CI slightly faster.

build_system/src/test.rs
```diff
@@ -876,54 +881,90 @@ where
...
-                    "test-attrs",
-                ]
-                .iter()
-                .any(|name| *name == dir_name)
-                {
-                    std::fs::remove_dir_all(dir).map_err(|error| {
-                        format!("Failed to remove folder `{}`: {:?}", dir.display(), error)
-                    })?;
+        if let Some(callback) = should_run_test_callback {
+            fn walk_dir<F, G>(
```
antoyo, line 885: Why is this new function necessary? Can't you reuse the existing `walk_dir`?

antoyo (commented):
There are a few changes I'd like to see.

Thanks for your work!

build_system/src/test.rs
lordshashank: Normal walk_dir doesn't allow walking through subdirectories I guess, thus had to implement this.

build_system/src/test.rs
lordshashank: @antoyo could you elaborate on what do you mean by split the function? Do you want a separate function for `test_failing_ui_pattern_tests` having the funcitonality of `should_run_test_callback` in it?

build_system/src/test.rs
antoyo: Would it work to add a parameter `recursive` to the existing function instead?

build_system/src/test.rs
antoyo: Actually, let's not split it since it seems complicated and use a bool param (or an enum if you think it's clearer). Then, add a comment to the new part (failing ui pattern tests) explaining that this remove the tests without an error pattern.

lordshashank: Actually I also implemented it with bool initially but I feel this way though a little complicated would help in generalization in future with more test suits being integrated and more functions like `should_remove_tests` being involved. Let me implement the changes with proper comments, then let me see if you still want changes here

build_system/src/test.rs
lordshashank: Ah! not sure how? I am aware of one walkdir crate that does it. Could you tell me if you if you know a way here. Moreover, custom functions are defined [here](https://github.com/lordshashank/rustc_codegen_gcc/blob/ad82635689a67aa0834020ac2f166a4c4ad6bb2e/build_system/src/test.rs#L957-L962) as well, though some customization was needed. Let me know if you have a way here.

lordshashank:
@antoyo added few comments, let me know if you still need change in parameter or anything else.

build_system/src/test.rs
antoyo: Yeah, I prefer with the bool/enum: we can generalize later if there's really a need.

lordshashank:
@antoyo added the bool function, let me know if anything else is needed now. Also updated description.

build_system/src/test.rs
```diff
@@ -856,10 +860,19 @@ fn should_remove_test(file_path: &Path) -> Result<bool, String> {
     Ok(false)
 }
 
+// # Parameters
+//
+// * `env`: An environment variable that provides context for the function.
+// * `args`: The arguments passed to the test. This could include things like the flags, config etc.
+// * `prepare_files_callback`: A callback function that prepares the files needed for the test. Its used to remove/retain tests giving Error to run various rust test suits.
```
antoyo, line 867: Please adjust this since it is now a bool.

build_system/src/test.rs
```diff
@@ -876,54 +889,92 @@ where
...
-                    "borrowck",
-                    "test-attrs",
-                ]
-                .iter()
-                .any(|name| *name == dir_name)
-                {
-                    std::fs::remove_dir_all(dir).map_err(|error| {
-                        format!("Failed to remove folder `{}`: {:?}", dir.display(), error)
-                    })?;
+        // uses contains_ui_error_patterns function being sent as callback to run only only error pattern tests
```
antoyo, line 892: Please update the comment since it is now a bool. Also add something like this in the comment:
```
After we removed the error tests that are known to panic with rustc_codegen_gcc, we now remove the passing tests since this runs the error tests.
```

build_system/src/test.rs
lordshashank: Hey its adjusted on below line, this is the files callback

build_system/src/test.rs
lordshashank: done

build_system/src/test.rs
antoyo: Oh right. Sorry for the mistake.

build_system/src/test.rs
antoyo: I believe it should be enough that to wrap the recursive call to `walk_dir()` with an `if recursive { […] }`.

lordshashank:
@antoyo ig we are good to go now.

build_system/src/test.rs
```diff
@@ -876,54 +889,92 @@ where
...
-                    "borrowck",
-                    "test-attrs",
-                ]
-                .iter()
-                .any(|name| *name == dir_name)
-                {
-                    std::fs::remove_dir_all(dir).map_err(|error| {
-                        format!("Failed to remove folder `{}`: {:?}", dir.display(), error)
-                    })?;
+        // After we removed the error tests that are known to panic with rustc_codegen_gcc, we now remove the passing tests since this runs the error tests.
```
antoyo, line 892: Please move this comment one line below so that it's in the if: it will be clearer.

antoyo (commented):
With these changes, it will be good to go.
Thanks for the work!

build_system/src/test.rs
lordshashank: @antoyo I think I am not able to understand what you mean here, is it something like this
```
fn walk_dir<F, G>(
    dir_path: PathBuf,
    dir_callback: F,
    file_callback: G,
    recursive: bool,
) -> Result<(), String>
where
    F: Fn(&Path) -> Result<(), String> + Copy,
    G: Fn(&Path) -> Result<(), String> + Copy,
{
    if dir_path.is_dir() {
        for entry in std::fs::read_dir(dir_path).unwrap() {
            let entry = entry;
            let path = entry.unwrap().path();
            if path.is_dir() {
                dir_callback(&path)?;
                if recursive {
                    walk_dir(path, dir_callback, file_callback, recursive)?; // Recursive call
                }
            } else if path.is_file() {
                file_callback(&path)?;
            }
        }
    }
    Ok(())
}

walk_dir(
    rust_path.join("tests/ui"),
    |_dir| Ok(()),
    |file_path| {
        if contains_ui_error_patterns(file_path)? {
            Ok(())
        } else {
            remove_file(file_path).map_err(|e| e.to_string())
        }
    },
    true, // Set recursive to true
)?;
```

antoyo: Yes, but please change the `walk_dir` function in the file `build_system/src/utils.rs` since it correctly handles the errors.

lordshashank: @antoyo did that but am getting recursion limit error 
```
error: reached the recursion limit while instantiating `walk_dir::<PathBuf, &mut &mut &mut &mut ..., ...>`
   --> src/utils.rs:362:17
    |
362 |                 walk_dir(entry_path, &mut dir_cb, &mut file_cb, recursive)?; // Recursive call
    |                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    |
note: `walk_dir` defined here
   --> src/utils.rs:346:1
    |
    ```
    would have to use `Box<dyn FnMut>` to avoid this. what to do?

Comment (antoyo) [reply build_system/src/test.rs:895]: Can you please post the complete code of the function?
Also, does it work if you do:
```
walk_dir(entry_path, dir_cb, file_cb, recursive)?; // Recursive call
```

Comment (lordshashank) [reply build_system/src/test.rs:895]: here is the function
```

pub fn walk_dir<P, D, F>(dir: P, mut dir_cb: D, mut file_cb: F, recursive: bool) -> Result<(), String>
where
    P: AsRef<Path>,
    D: FnMut(&Path) -> Result<(), String>,
    F: FnMut(&Path) -> Result<(), String>,
{
    let dir = dir.as_ref();
    for entry in fs::read_dir(dir)
        .map_err(|error| format!("Failed to read dir `{}`: {:?}", dir.display(), error))?
    {
        let entry = entry
            .map_err(|error| format!("Failed to read entry in `{}`: {:?}", dir.display(), error))?;
        let entry_path = entry.path();
        if entry_path.is_dir() {
            dir_cb(&entry_path)?;
            if recursive {
                walk_dir(entry_path, &mut dir_cb, &mut file_cb, recursive)?; // Recursive call
            }
        } else {
            file_cb(&entry_path)?;
        }
    }
    Ok(())
}
```
Also @antoyo, removing `&mut` would lead to borrowing of moved value, hence used that

Comment (antoyo) [reply build_system/src/test.rs:895]: I guess this would work:
```
pub fn walk_dir<P, D, F>(dir: P, dir_cb: &mut D, file_cb: &mut F, recursive: bool) -> Result<(), String>
where
    P: AsRef<Path>,
    D: FnMut(&Path) -> Result<(), String>,
    F: FnMut(&Path) -> Result<(), String>,
{
    let dir = dir.as_ref();
    for entry in fs::read_dir(dir)
        .map_err(|error| format!("Failed to read dir `{}`: {:?}", dir.display(), error))?
    {
        let entry = entry
            .map_err(|error| format!("Failed to read entry in `{}`: {:?}", dir.display(), error))?;
        let entry_path = entry.path();
        if entry_path.is_dir() {
            dir_cb(&entry_path)?;
            if recursive {
                walk_dir(entry_path, dir_cb, file_cb, recursive)?; // Recursive call
            }
        } else {
            file_cb(&entry_path)?;
        }
    }
    Ok(())
}
```
This way, you will need to add `&mut` at the call-site for the callback arguments.

lordshashank:
LFG 🚀

PR #528, opened 2024-05-31
prevent libgccjit.so download on every other system except Linux x86-64

g4titanx:


build_system/src/build.rs
```diff
@@ -234,3 +234,21 @@ pub fn run() -> Result<(), String> {
     build_codegen(&mut args)?;
     Ok(())
 }
+
+fn main() {
+    // Check target OS and architecture
+    let target_os = std::env::var("CARGO_CFG_TARGET_OS").unwrap();
```
GuillaumeGomez, line 240: Two things:
 1. This is not a build script, so adding a `main` function here is likely to prevent compilation.
 2. We don't want to prevent compilation from everyone not on linux, we want to prevent **download**.

g4titanx: oh! my bad. that would indeed prevent compilation. `fn main` in this code shouldn't prevent compilation, and this is a way I thought of to fix this: 

`
use std::env;
use std::fs;
use std::path::Path;
use std::process::Command;

fn main() {
    // Check target platform
    let target_os = env::var("CARGO_CFG_TARGET_OS").unwrap_or_default();
    let target_arch = env::var("CARGO_CFG_TARGET_ARCH").unwrap_or_default();

    if target_os == "linux" && target_arch == "x86_64" {
        println!("Target is Linux x86_64. Proceeding with download.");

        // Download libgccjit.so on Linux x86_64
        if let Err(e) = download_libgccjit() {
            eprintln!("Failed to download libgccjit.so: {}", e);
            std::process::exit(1);
        }
    } else {
        println!("Skipping libgccjit.so download on non-Linux x86_64 platform.");
    }
}

fn download_libgccjit() -> Result<(), String> {
    // Example logic to download libgccjit.so directly to the required location
    // You might use a command like `curl` with `-o` option to specify the output location
    let output = Command::new("curl")
        .arg("-o")
        .arg("/usr/local/lib/libgccjit.so") // Example destination path
        .arg("http://example.com/path/to/libgccjit.so")
        .output()
        .map_err(|e| format!("Failed to execute curl: {}", e))?;

    if !output.status.success() {
        return Err(format!(
            "Failed to download libgccjit.so: {}",
            String::from_utf8_lossy(&output.stderr)
        ));
    }

    Ok(())
}
`

g4titanx: oh! my bad, i got that wrong, I thought of fixing the issue in this way:

```rust
// assuming the code logic to download libgccjit.so is here. you could walk me thru that tho
fn download_libgccjit() -> Result<(), String> {
    // existing program logic to download libgccjit.so
    Ok(())
}

pub fn run() -> Result<(), String> {
    let os = env::consts::OS;
    let arch = env::consts::ARCH;

    // checks if the OS is Linux and the architecture is x86_64
    if os!= "linux" || arch!= "x86_64" {
        // prints the message but continue with the rest of the compilation
        eprintln!("Compile libgccjit yourself and set option XXXX.");
    }

    let mut args = match BuildArg::new()? {
        Some(args) => args,
        None => return Ok(()),
    };
    args.config_info.setup_gcc_path()?;
    
    // conditionally proceed with the download
    if os == "linux" && arch == "x86_64" {
        if download_libgccjit()?.is_err() {
            return Err("Failed to download libgccjit.so.".to_string());
        }
    }

    build_codegen(&mut args)?;
    Ok(())
}
```

GuillaumeGomez: This way seems good indeed. :+1:

GuillaumeGomez: We have a git commit hash and we get the release for this hash. We only generate a release for linux for now so that's why we want to display to people that they can't use `download` if not on linux x64.

g4titanx: okay. this would be ideal then:
```rust
pub fn run() -> Result<(), String> {
    let os = env::consts::OS;
    let arch = env::consts::ARCH;

    let mut args = match BuildArg::new()? {
        Some(args) => args,
        None => return Ok(()),
    };
    args.config_info.setup_gcc_path()?;
    
    // check if the OS is Linux and the architecture is x86_64
    if os == "linux" && arch == "x86_64" {
        // since the release is already only for Linux, we can safely proceed with the download or release fetching
        // this is a placeholder for the actual logic that uses the release hash
        println!("Using release hash: {}", RELEASE_HASH);
    } else {
        println!("Compiling without downloading due to non-Linux platform.");
    }

    build_codegen(&mut args)?;
    Ok(())
}
```
but instead of RELEASE_HASH what would be there?

Skgland: > yeah, I could implement that in build_system/src/build.rs but what's the existing program logic to download libgccjit.so?

I think it is [download_gccjit_if_needed in /build_system/src/config.rs](https://github.com/rust-lang/rustc_codegen_gcc/blob/master/build_system/src/config.rs#L214),
which appears to only be used in [setup_gcc_path](https://github.com/rust-lang/rustc_codegen_gcc/blob/master/build_system/src/config.rs#L272) further down in the same file.

g4titanx: oh great! i didnt realize that

Skgland: Also rather than changing `run` here I think the check would make the most sense just before we do the actual download so in [`download_gccjit_if_needed` just before `download_gccjit` is called](https://github.com/rust-lang/rustc_codegen_gcc/blob/master/build_system/src/config.rs#L229) or at the start of [`download_gccjit`](https://github.com/rust-lang/rustc_codegen_gcc/blob/master/build_system/src/config.rs#L471).
That makes forgetting to check harder and we aren't warning when we wouldn't have downloaded.

g4titanx: yeah, that's a better fix, i just checked. 
you can attach it to the issue

antoyo:
I guess we should close this since #529 was merged.

PR #529, opened 2024-05-31
prevent libgccjit.so download on unsupported os/arch

### LINKED ISSUE #527: Do not allow downloading libgccjit.so on other systems than Linux x86-64 in the build system

Skgland:
Alternative to #528, I find it easier to show what I meant rather than tell.

Fixing up the download URL while at it. 

fixes #527

### DIFF
--- a/build_system/src/config.rs
+++ b/build_system/src/config.rs
@@ -474,6 +474,19 @@ fn download_gccjit(
     tempfile_name: String,
     with_progress_bar: bool,
 ) -> Result<(), String> {
+    if std::env::consts::OS != "linux" || std::env::consts::ARCH != "x86_64" {
+        eprintln!(
+            "\
+        pre-compiled libgccjit.so not available for this os or architecture\n\
+        please compile it yourself and in the config.toml set\n\
+        `download-gccjit = false` and `gcc-path` to the appropriate directory\
+        "
+        );
+        return Err(String::from(
+            "no appropriate pre-compiled libgccjit.so available for download",
+        ));
+    }
+
     // Try curl. If that fails and we are on windows, fallback to PowerShell.
     let mut ret = run_command_with_output(
         &[
--- a/build_system/src/config.rs
+++ b/build_system/src/config.rs
@@ -473,9 +473,9 @@ fn download_gccjit(
     } else {
         eprintln!(
             "\
-        pre-compiled libgccjit.so not available for this os or architecture\n\
-        please compile it yourself and in the config.toml set\n\
-        `download-gccjit = false` and `gcc-path` to the appropriate directory\
+        Pre-compiled libgccjit.so not available for this os or architecture.\n\
+        Please compile it yourself and update the `config.toml` file\n\
+        to `download-gccjit = false` and set `gcc-path` to the appropriate directory.\
         "
         );
         return Err(String::from(
--- a/build_system/src/config.rs
+++ b/build_system/src/config.rs
@@ -473,10 +473,9 @@ fn download_gccjit(
     } else {
         eprintln!(
             "\
-        Pre-compiled libgccjit.so not available for this os or architecture.\n\
-        Please compile it yourself and update the `config.toml` file\n\
-        to `download-gccjit = false` and set `gcc-path` to the appropriate directory.\
-        "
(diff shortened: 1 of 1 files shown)


build_system/src/config.rs
```diff
@@ -469,11 +463,28 @@ impl ConfigInfo {
...
     with_progress_bar: bool,
 ) -> Result<(), String> {
+    let url = if std::env::consts::OS == "linux" || std::env::consts::ARCH == "x86_64" {
+        format!("https://github.com/rust-lang/gcc/releases/download/master-{}/libgccjit.so", commit)
+    } else {
+        eprintln!(
+            r"\
+        pre-compiled libgccjit.so not available for this os or architecture\n\
+        please compile it yourself and in the config.toml set\n\
+        `download-gccjit = false` and `gcc-path` to the appropriate directory\
```
GuillaumeGomez, line 478:
```suggestion
        Pre-compiled libgccjit.so not available for this os or architecture.\n\
        Please compile it yourself and update the `config.toml` file\n\
        to `download-gccjit = false` and set `gcc-path` to the appropriate directory.\
```

build_system/src/config.rs
```diff
@@ -469,11 +463,28 @@ impl ConfigInfo {
...
+    commit: &str,
     output_dir: &Path,
     tempfile_name: String,
     with_progress_bar: bool,
 ) -> Result<(), String> {
+    let url = if std::env::consts::OS == "linux" || std::env::consts::ARCH == "x86_64" {
+        format!("https://github.com/rust-lang/gcc/releases/download/master-{}/libgccjit.so", commit)
+    } else {
+        eprintln!(
+            r"\
```
GuillaumeGomez, line 475: Is the `r` useful here?

build_system/src/config.rs
```diff
@@ -469,11 +463,28 @@ impl ConfigInfo {
...
 
 fn download_gccjit(
-    url: String,
+    commit: &str,
     output_dir: &Path,
     tempfile_name: String,
     with_progress_bar: bool,
 ) -> Result<(), String> {
+    let url = if std::env::consts::OS == "linux" || std::env::consts::ARCH == "x86_64" {
+        format!("https://github.com/rust-lang/gcc/releases/download/master-{}/libgccjit.so", commit)
```
GuillaumeGomez, line 472: Thanks! :)

Skgland: Oh, looking at this I just noticed that I messed up the condition it should be and not or.

GuillaumeGomez: Oh indeed! Completely missed it as well. ^^'

build_system/src/config.rs
Skgland: Quite the opposite of useful actually as a raw string is not what I want here.🤦‍♂️

Skgland:
Ok, applied your suggestion regarding the message text change, fixed the conditional to be `&&` instead of `||` and changed the raw-string that shouldn't be raw to a normal string.
Assuming CI agrees, this is good to go from my point.

build_system/src/config.rs
```diff
@@ -469,11 +463,28 @@ impl ConfigInfo {
...
 ) -> Result<(), String> {
+    let url = if std::env::consts::OS == "linux" && std::env::consts::ARCH == "x86_64" {
+        format!("https://github.com/rust-lang/gcc/releases/download/master-{}/libgccjit.so", commit)
+    } else {
+        eprintln!(
+            "\
+        Pre-compiled libgccjit.so not available for this os or architecture.\n\
+        Please compile it yourself and update the `config.toml` file\n\
+        to `download-gccjit = false` and set `gcc-path` to the appropriate directory.\
+        "
```
GuillaumeGomez, line 479: Just thinking but at this point we could simply do:

```suggestion
Pre-compiled libgccjit.so not available for this os or architecture.
Please compile it yourself and update the `config.toml` file
to `download-gccjit = false` and set `gcc-path` to the appropriate directory."
```

Skgland: Sure, if that is preferred, done

GuillaumeGomez (approved):

PR #533, opened 2024-06-19
Refresh README documentation

### LINKED ISSUE #501: Instructions should use shell syntax for variables
This is regarding this excerpt from the README, primarily:

> `$ LIBRARY_PATH="[gcc-path value]" LD_LIBRARY_PATH="[gcc-path value]" rustc +$(cat $CG_GCCJIT_DIR/rust-toolchain | grep 'channel' | cut -d '=' -f 2 | sed 's/"//g' | sed 's/ //g') -Cpanic=abort -Zcodegen-backend=$CG_GCCJIT_DIR/target/release/librustc_codegen_gcc.so --sysroot $CG_GCCJIT_DIR/build_sysroot/sysroot my_crate.rs`

This has a problem. Namely: if we ignore the `$`, this is *legal shell syntax*, but the command will probably fail? Variables get set to legal strings that are useless values, i.e. `[gcc-path value]` does not mean anything in terms of paths. The linker will ignore these, but I imagine you meant to suggest actual commands or paths... in other words, substitutions. Like variables?

I am aware of the `[option]` convention in man pages, but that is for _optional_ values, and these commands will simply fail if these are not set. It is also a hoary convention lost to the mists of time and not actually documented anywhere, so anyone unfamiliar with this particular notation will have no idea.

If you want to force them to define the variable (and thus, likely force them to break if they try to simply copy and paste the code), you can do this for these vars:

`LIBRARY_PATH="${YOUR_GCC_PATH_VALUE:?define this to your path}"`

This command will fail (setting `$?` to `1`, it doesn't really "exit") with an error like:
> bash: YOUR_GCC_PATH_VALUE: define this to your path

This is, despite being somewhat obscure syntax, portable shell. All POSIX shells are required to understand this, and non-POSIX shells I am aware of will at least parse-error on it, even if they don't replicate this behavior.

darcagn:
In the 6 months or so I've been using this project to target SuperH and integrate it with the toolchain targeting the Sega Dreamcast ([KallistiOS](https://github.com/KallistiOS/KallistiOS)), I've had a few issues and pain points with the documentation that I'd like to address. I've also spent a lot of time troubleshooting, figuring things out with cross-compilation, listening to feedback from interested users, etc. so I'd like to pass on what I've learned by fixing up and refreshing the documentation here. Some of the changes are slightly opinionated or adjust the original steps so feel free to ask questions or for clarification, but nothing should be too bad. 😄 

Thanks for all the hard work you all have done on rustc_codegen_gcc.

This should also address and fix #501

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -1,155 +1,135 @@
-# WIP libgccjit codegen backend for rust
+# GCC codegen backend for Rust (WIP)
 
 [![Chat on IRC](https://img.shields.io/badge/irc.libera.chat-%23rustc__codegen__gcc-blue.svg)](https://web.libera.chat/#rustc_codegen_gcc)
 [![Chat on Matrix](https://img.shields.io/badge/matrix.org-%23rustc__codegen__gcc-blue.svg)](https://matrix.to/#/#rustc_codegen_gcc:matrix.org)
 
-This is a GCC codegen for rustc, which means it can be loaded by the existing rustc frontend, but benefits from GCC: more architectures are supported and GCC's optimizations are used.
+This GCC codegen enables the official rustc compiler frontend to leverage GCC as a compiler backend. This is accomplished through interfacing with GCC's [libgccjit](https://gcc.gnu.org/wiki/JIT) library. While libgccjit is primarily intended to be used as an interface to GCC for just-in-time code generation, it can also be used for ahead-of-time compilation, as we do here.
 
-**Despite its name, libgccjit can be used for ahead-of-time compilation, as is used here.**
-
-## Motivation
-
-The primary goal of this project is to be able to compile Rust code on platforms unsupported by LLVM.
-A secondary goal is to check if using the gcc backend will provide any run-time speed improvement for the programs compiled using rustc.
-
-### Dependencies
-
-**rustup:** Follow the instructions on the official [website](https://www.rust-lang.org/tools/install)
-
-**DejaGnu:** Consider to install DejaGnu which is necessary for running the libgccjit test suite. [website](https://www.gnu.org/software/dejagnu/#downloading)
-
-
+### Goals
+* Enable the compilation of Rust code for target platforms supported by GCC, but not LLVM
+* Leverage potential GCC optimizations for run-time speed improvement for Rust applications
+* Reduce dependence of the Rust ecosystem on LLVM
 
 ## Building
+### Dependencies
+* **[rustup](https://www.rust-lang.org/tools/install)**: The rustup tool is required to build this project using the following instructions; do not rely on a Rust toolchain that may have been provided by your operating system's package manager.
+* **[DejaGnu](https://www.gnu.org/software/dejagnu/#downloading)** (optional): Install the DejaGnu testing framework in order to run the libgccjit test suite.
 
-**This requires a patched libgccjit in order to work.
-You need to use my [fork of gcc](https://github.com/antoyo/gcc) which already includes these patches.**
-
+### Building GCC with libgccjit.so
+#### On Linux x86-64
+When using an x86-64 Linux host to target x86-64 Linux, building `libgccjit.so` is unnecessary -- in that case, a precompiled version may be downloaded automatically. Simply copy the provided `config.example.toml` file to `config.toml` to enable the automatic downloading of `libgccjit.so`.
 ```bash
 $ cp config.example.toml config.toml
 ```
+Now you may skip ahead to [Building rustc_codegen_gcc](#building-rustc_codegen_gcc).
 
-If don't need to test GCC patches you wrote in our GCC fork, then the default configuration should
-be all you need. You can update the `rustc_codegen_gcc` without worrying about GCC.
+#### Other architectures
+If you are on a host arch other than x86-64 Linux or are targetting an arch other than x86-64 Linux, you will need to build a custom GCC with `libgccjit.so`. **At this time, this requires the use of the [rust-lang/gcc](https://github.com/rust-lang/gcc) fork of GCC, which includes patches to libgccjit to enable the use of this codegen.**
 
-### Building with your own GCC version
-
-If you wrote a patch for GCC and want to test it without this backend, you will need
-to do a few more things.
-
-To build it (most of these instructions come from [here](https://gcc.gnu.org/onlinedocs/jit/internals/index.html), so don't hesitate to take a look there if you encounter an issue):
+Full instructions to build libgccjit are provided in the [libgccjit documentation](https://gcc.gnu.org/onlinedocs/jit/internals/index.html), so check there if you encounter issues, however brief directions for Debian-based Linux follow below. You may need to adapt them to your operating system and package manager. If you need to build a cross-compiler, see [Building a cross-compiler with rustc_codegen_gcc](./doc/cross.md).
 
 ```bash
-$ git clone https://github.com/antoyo/gcc
+$ git clone https://github.com/rust-lang/gcc gcc-source
 $ sudo apt install flex libmpfr-dev libgmp-dev libmpc3 libmpc-dev
 $ mkdir gcc-build gcc-install
 $ cd gcc-build
-$ ../gcc/configure \
+$ ../gcc-source/configure \
     --enable-host-shared \
     --enable-languages=jit \
-    --enable-checking=release \ # it enables extra checks which allow to find bugs
+    --enable-checking=release \
     --disable-bootstrap \
     --disable-multilib \
     --prefix=$(pwd)/../gcc-install
-$ make -j4 # You can replace `4` with another number depending on how many cores you have.
+$ make -j4
+$ make install
 ```
 
-If you want to run libgccjit tests, you will need to also enable the C++ language in the `configure`:
+Notes:
+* If you want to run libgccjit tests, you must also add C++ to the enabled languages: `--enable-languages=jit,c++`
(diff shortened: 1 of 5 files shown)


Readme.md
```diff
@@ -1,155 +1,135 @@
...
 
+### Goals
+* Enable the compilation of Rust code for target platforms supported by GCC, but not LLVM
+* Leverage potential GCC optimizations for run-time speed improvement for Rust applications
+* Reduce dependence of the Rust ecosystem on LLVM
 
 ## Building
+### Dependencies
+* **[rustup](https://www.rust-lang.org/tools/install)**: The rustup tool is required to build this project using the following instructions; do not rely on a Rust toolchain that may have been provided by your operating system's package manager.
+* **[DejaGnu](https://www.gnu.org/software/dejagnu/#downloading)** (optional): Install the DejaGnu testing framework in order to run the libgccjit test suite.
```
GuillaumeGomez, line 16: I didn't install it and yet tests are working fine for me?

Readme.md
```diff
@@ -1,155 +1,135 @@
...
+This GCC codegen enables the official rustc compiler frontend to leverage GCC as a compiler backend. This is accomplished through interfacing with GCC's [libgccjit](https://gcc.gnu.org/wiki/JIT) library. While libgccjit is primarily intended to be used as an interface to GCC for just-in-time code generation, it can also be used for ahead-of-time compilation, as we do here.
 
+### Goals
+* Enable the compilation of Rust code for target platforms supported by GCC, but not LLVM
+* Leverage potential GCC optimizations for run-time speed improvement for Rust applications
+* Reduce dependence of the Rust ecosystem on LLVM
 
 ## Building
+### Dependencies
+* **[rustup](https://www.rust-lang.org/tools/install)**: The rustup tool is required to build this project using the following instructions; do not rely on a Rust toolchain that may have been provided by your operating system's package manager.
```
GuillaumeGomez, line 15: I think `https://rustup.rs/` is better for installing rustup.

Readme.md
```diff
@@ -1,155 +1,135 @@
...
-If you want to run libgccjit tests, you will need to also enable the C++ language in the `configure`:
+Notes:
+* If you want to run libgccjit tests, you must also add C++ to the enabled languages: `--enable-languages=jit,c++`
+* `--enable-host-shared` builds the compiler as position independent code and is required to build libgccjit. This results in a slower compiler; however, if building GCC in multiple passes, `jit` may be built in the first pass with `--enable-host-shared`, with both disabled in subsequent passes.
+* `--enable-checking=release` and `--disable-bootstrap` speed up compilation by disabling self-checks and may be omitted.
+* `--disable-multilib` is used as libgccjit only supports targeting one arch ABI variant.
+* Adjust the `4` in `make -j4` to the number of threads available on your system to speed up compilation.
+* Change `make install` to `make install-strip` to remove debug symbols from GCC, including `libgccjit.so`, to save on disk space.
 
+Once the build is complete, one may run libgccjit tests like so (requires DejaGnu to be installed):
```
GuillaumeGomez, line 55: No it's not needed.

Readme.md
```diff
@@ -1,155 +1,135 @@
...
 ```bash
---enable-languages=jit,c++
+$ make -C gcc check-jit
```
GuillaumeGomez, line 57: Why adding `-C gcc`?

Readme.md
```diff
@@ -1,155 +1,135 @@
...
+### Dependencies
+* **[rustup](https://www.rust-lang.org/tools/install)**: The rustup tool is required to build this project using the following instructions; do not rely on a Rust toolchain that may have been provided by your operating system's package manager.
+* **[DejaGnu](https://www.gnu.org/software/dejagnu/#downloading)** (optional): Install the DejaGnu testing framework in order to run the libgccjit test suite.
 
-**This requires a patched libgccjit in order to work.
-You need to use my [fork of gcc](https://github.com/antoyo/gcc) which already includes these patches.**
-
+### Building GCC with libgccjit.so
+#### On Linux x86-64
+When using an x86-64 Linux host to target x86-64 Linux, building `libgccjit.so` is unnecessary -- in that case, a precompiled version may be downloaded automatically. Simply copy the provided `config.example.toml` file to `config.toml` to enable the automatic downloading of `libgccjit.so`.
```
GuillaumeGomez, line 20:
```suggestion
When using an x86-64 Linux host to target x86-64 Linux, building `libgccjit.so` is unnecessary (unless you patched it) -- in that case, a precompiled version may be downloaded automatically. Simply copy the provided `config.example.toml` file to `config.toml` to enable the automatic downloading of `libgccjit.so`.
```

Readme.md
```diff
@@ -1,155 +1,135 @@
...
-
+Now that you've compiled GCC with libgccjit support, the installed GCC toolchain with `libgccjit.so` can be found in the `gcc-install` directory created above. You will need to provide the path to the specific subdirectory containing `libgccjit.so` to rustc_codegen_gcc. Use the following commands to leave the `gcc-build` directory and find the absolute path within the `gcc-install` directory:
 ```bash
-$ dirname $(readlink -f `find . -name libgccjit.so`)
+$ cd ..
+$ dirname $(readlink -f `find ./gcc-install -name libgccjit.so`)
 ```
 
-and to comment the `download-gccjit` setting:
+If desired, you may now delete the `gcc-source` and `gcc-build` directories to reclaim disk space, but keep the `gcc-install` directory.
```
GuillaumeGomez, line 70: I wouldn't recommend that.

Readme.md
```diff
@@ -1,155 +1,135 @@
...
-# download-gccjit = true
 ```
 
-Then you can run commands like this:
+### Building rustc_codegen_gcc
+Now that you have a `config.toml` file set up to use `libgccjit.so`, you can proceed to building the build system and sysroot. `prepare` will retrieve and patch the sysroot source, while `build` will build the sysroot.
 
 ```bash
-$ ./y.sh prepare # download and patch sysroot src and install hyperfine for benchmarking
+$ ./y.sh prepare
```
GuillaumeGomez, line 81: Please leave the comment.

Readme.md
```diff
@@ -1,155 +1,135 @@
...
 $ ./y.sh build --sysroot --release
 ```
 
-To run the tests:
-
+You may also run the tests:
```
GuillaumeGomez, line 85: Please leave the original sentence.

Readme.md
```diff
@@ -1,155 +1,135 @@
...
 ```bash
 $ ./y.sh test --release
 ```
 
 ## Usage
-
-You have to run these commands, in the corresponding order:
-
+The following example commands are run with `$CG_GCCJIT_DIR` representing the path to your rustc_codegen_gcc directory.
```
GuillaumeGomez, line 91: I think we should remove all mentions to `$CG_GCCJIT_DIR`.

Readme.md
```diff
@@ -1,155 +1,135 @@
...
-
-You have to run these commands, in the corresponding order:
-
+The following example commands are run with `$CG_GCCJIT_DIR` representing the path to your rustc_codegen_gcc directory. 
+### Cargo
+To invoke `cargo`, run the following example command:
 ```bash
-$ ./y.sh prepare
-$ ./y.sh build --sysroot
+$ CHANNEL="release" $CG_GCCJIT_DIR/y.sh cargo run
```
GuillaumeGomez, line 95: No need for `CHANNEL="release"`.

Readme.md
```diff
@@ -1,155 +1,135 @@
...
+You may verify your build is working properly by building a test project:
  ```bash
-$ ./y.sh cargo build --manifest-path tests/hello-world/Cargo.toml
-```
-
-### Cargo
-
-```bash
-$ CHANNEL="release" $CG_GCCJIT_DIR/y.sh cargo run
+$ CHANNEL="release" $CG_GCCJIT_DIR/y.sh cargo build --manifest-path $CG_GCCJIT_DIR/tests/hello-world/Cargo.toml
```
GuillaumeGomez, line 100: Please revert.

Readme.md
```diff
@@ -1,155 +1,135 @@
...
 ```
 
-If you compiled cg_gccjit in debug mode (aka you didn't pass `--release` to `./y.sh test`) you should use `CHANNEL="debug"` instead or omit `CHANNEL="release"` completely.
+Note: If you compiled rustc_codegen_gcc in debug mode (i.e., you didn't pass `--release` to `./y.sh` above), you should use `CHANNEL="debug"` or omit `CHANNEL="release"` completely.
```
GuillaumeGomez, line 103: Would be interesting to instead explain here that the `CHANNEL` environment variable can be used.

config.example.toml
```diff
@@ -1,2 +1,2 @@
-#gcc-path = "gcc-build/gcc"
+#gcc-path = "gcc-install/lib"
```
GuillaumeGomez, line 1: Please revert. I never run `make install` and I'd like to keep it that way. You can mention the install folder too though.

antoyo: I believe `make install` can be necessary for e.g. using LTO.

Readme.md
antoyo: If you're talking about DejaGnu, it is required and I added this dependency in a previous PR because some people did not have this installed.

Readme.md
antoyo: As I mentioned elsewhere, it is required and some systems don't have it installed by default.

Readme.md
antoyo: I guess that may be useful for people wanting the use the project, but not contribute on it.

Readme.md
antoyo: Since it seems like a recurring wish to be able to run `y.sh` from anywhere (@darcagn: is this why you mention this variable?), perhaps we could have a section explaining what I did to alleviate this issue, which is I created an alias `y = /path/to/rustc_codegen_gcc/y.sh`?

PR #547, opened 2024-08-03
Update git-subtree doc to install in /opt

antoyo:


antoyo:
To keep the home directory clean. The current approach will add 3 directories to the home directory.

GuillaumeGomez:
Why not a folder in user's home containing the three folders?

PR #549, opened 2024-08-04
Replace stack_var_count and RETURN_VALUE_COUNT with a single counter

sapir:


antoyo:
Thanks for your contribution!

PR #551, opened 2024-08-05
Fix load alignment being applied to referenced value

sapir:
In some code I compiled, the `align` argument wasn't the same as the pointer's regular alignment, and gcc ended up dereferencing the wrong address. So I think the `align` argument is meant to only apply to the loaded value and not to the one that's being read.

For comparison, rustc_codegen_llvm [does the following](https://github.com/rust-lang/rust/blob/f7eefec4e03f5ba723fbc04d94dbc1203b7c9bff/compiler/rustc_codegen_llvm/src/builder.rs#L496C1-L498C1):

```rust
let load = llvm::LLVMBuildLoad2(self.llbuilder, ty, ptr, UNNAMED);
llvm::LLVMSetAlignment(load, align.bytes() as c_uint);
```


antoyo:
> ```rust
> llvm::LLVMSetAlignment(load, align.bytes() as c_uint);
> ```

The load alignment is somewhat confusing.
I don't remember for sure, but I believe the `LLVMSetAlignment` function applies to the `load` instruction and not the location where the value is loaded.
Since your fix makes some test fails, I would tend to think that the code is correct as is, but I'm might be wrong.

> In some code I compiled, the align argument (edit: in Builder::load) wasn't the same as the pointer's regular alignment, and gcc ended up dereferencing the wrong address.

Could you please share an example where this happens?

sapir:
I'll try to make an example tomorrow.

antoyo:
Can you explain what's the expected result with this code?

I get:
```
[src/main.rs:15:5] b1 = 2
[src/main.rs:16:5] b2 = 2
```
which seems good.

sapir:
I'm getting 2 on the first line and some random value on the second.

antoyo:
Which architecture are you on?

sapir:
I was trying mips-unknown-linux-gnu. Since you asked, I also tried some other architectures including m68k but so far it hasn't happened anywhere except the mips one.

Also, it only happens with `--release` (I mean for cargo, e.g. `y.sh cargo build --release`). Without the `--release`, the argument gets passed fine.

antoyo:
Do you know if it would be easy to add a test for this fix?

sapir:
I think a test for the behavior might not be perfectly reliable because the bug could accidentally be optimized out, but I'll try. I think maybe with a higher field alignment it'll happen on m68k, too.

antoyo:
Did you want to add a test or would you want to merge this as is?

sapir:
I'd like to try to add a test that works on the architectures currently being tested in CI.

antoyo:
Ok, please tell me when it's ready!
Thanks a lot for your work!

PR #552, opened 2024-08-05
Implement int extension

sapir:
I tried to fix the FIXME in `zext`. Would this be acceptable?

Also, how should I add tests?


antoyo:
For the test, you can either add it to [this file](https://github.com/rust-lang/rustc_codegen_gcc/blob/master/tests/run/int.rs) or you can create another file in that directory.

antoyo:
Did you want to add a test or would you want to merge this as is?

sapir:
I did try to add a test and found that the test passes either way, so maybe I should change the PR to just add the test instead of changing the code?

antoyo:
You mean, to make sure the test fails without the implementation of int extension?
If so, yes, you can push a test, remove the implementation to make sure the test fails and push back the implementation.

PR #554, opened 2024-08-07
Remove dummy value locals for function ptr calls

sapir:
This makes the generated code cleaner so it's easier to debug.

### DIFF
--- a/src/builder.rs
+++ b/src/builder.rs
@@ -341,7 +341,7 @@ impl<'a, 'gcc, 'tcx> Builder<'a, 'gcc, 'tcx> {
             self.block
                 .add_eval(self.location, self.cx.context.new_call(self.location, func, &args));
             // Return dummy value when not having return value.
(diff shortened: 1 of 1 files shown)


GuillaumeGomez (approved):

GuillaumeGomez:
Nice improvement!

antoyo:
I believe I had a good reason to do this.
You can see where I did this [here](https://github.com/rust-lang/rustc_codegen_gcc/commit/0a4122fbbac91011467ca0c386525f3709d31a74#diff-e4f794ca308aa712cbad31feb714df7cada1bbe453a30e232e45891571430263R385-R387).

I'm not sure I remember correctly, but this was certainly for debugging reasons, possibly to clearly indicate that this value is never used because I might have thought otherwise at the time.

This is possibly not needed anymore, though.

antoyo (approved):

antoyo:
Thanks for your contribution!

PR #576, opened 2024-12-10
stabilize `lang_tests_common` config parsing logic

onur-ozkan:
`config.toml` isn't there (and shouldn't be there) by default, which makes this part of the code to fail on compilation, and that fails upstream rust to bootstrap with certain commands (e.g., `x fix`). This change should fix that problem.


GuillaumeGomez (approved):

GuillaumeGomez:
Nice improvement, thanks!

PR #580, opened 2024-12-12
Simplify intrinsics generation tool and regenerate intrinsics list

### LINKED ISSUE #575: Simplify generate_intrinsics.py
Currently it tries to fetch data from three sources:

* The `Intrinsics.td` file of the latest commit on https://github.com/llvm/llvm-project
* https://github.com/GuillaumeGomez/llvmint
* https://github.com/antoyo/llvmint

The last two sources are because the last commit on llvm-project may have removed some intrinsics that stdarch uses. It should be possible to simplify this to just `Intrinsics.td` from llvm-project by picking the exact commit that the nightly version against which cg_gcc will build uses as LLVM version. If any intrinsics are removed there, stdarch has had to stop using them before, so the removal of the intrinsic wouldn't matter.

GuillaumeGomez:
Fixes #575.


GuillaumeGomez:
Interestingly enough, a lot of intrinsics used by rustc are not listed in `tblgen`, which is very weird. So I think for now we can't really simplify the script...

antoyo:
Should we close this PR and the corresponding issue, then?

GuillaumeGomez:
Likely not. I think we need to list intrinsics that should be generated manually because they're used by rust but not listed by LLVM. Maybe @bjorn3 got some new info about this situation (I suppose not but who knows?).

GuillaumeGomez:
Superseeded by https://github.com/rust-lang/rustc_codegen_gcc/pull/823.

PR #600, opened 2025-01-15
Grep check for LTO Test

### LINKED ISSUE #591: Grep for call my_func in the LTO test
[Here](https://github.com/rust-lang/rustc_codegen_gcc/blob/93f7f2fdeba33c748cbda750cdbaa2c80bb34be1/.github/workflows/release.yml#L78)

Anon258:
Fixes #591


.github/workflows/release.yml
```diff
@@ -76,4 +76,10 @@ jobs:
     - name: Run y.sh cargo build
       run: |
         EMBED_LTO_BITCODE=1 CHANNEL="release" ./y.sh cargo build --release --manifest-path tests/hello-world/Cargo.toml
-        # TODO: grep the asm output for "call my_func" and fail if it is found.
+        asm=$(objdump -dj .text tests/hello-world/target/release/hello_world)
+        if echo asm | grep -q "call .*mylib.*my_func" ; then 
+          echo "call my_func found in asm"
```
antoyo, line 81: Please add in this string something saying that it should not be found because LTO should have removed it and that this tests LTO.

.github/workflows/release.yml
```diff
@@ -76,4 +76,10 @@ jobs:
     - name: Run y.sh cargo build
       run: |
         EMBED_LTO_BITCODE=1 CHANNEL="release" ./y.sh cargo build --release --manifest-path tests/hello-world/Cargo.toml
-        # TODO: grep the asm output for "call my_func" and fail if it is found.
+        asm=$(objdump -dj .text tests/hello-world/target/release/hello_world)
+        if echo asm | grep -q "call .*mylib.*my_func" ; then
```
antoyo, line 80: It's probably more efficient to do:
```suggestion
        call_found=$(objdump -dj .text tests/hello-world/target/release/hello_world | grep -q "call .*mylib.*my_func")
        if call_found then 
```

.github/workflows/release.yml
```diff
@@ -76,4 +76,10 @@ jobs:
     - name: Run y.sh cargo build
       run: |
         EMBED_LTO_BITCODE=1 CHANNEL="release" ./y.sh cargo build --release --manifest-path tests/hello-world/Cargo.toml
-        # TODO: grep the asm output for "call my_func" and fail if it is found.
+        asm=$(objdump -dj .text tests/hello-world/target/release/hello_world)
+        if echo asm | grep -q "call .*mylib.*my_func" ; then 
+          echo "call my_func found in asm"
+          exit 1 
+        else 
+          echo "success"
```
antoyo, line 84: No need to print this.
```suggestion
```

antoyo (commented):
Thanks. Here are a few suggestions.

PR #608, opened 2025-01-24
Add `--gcc-path` option

GuillaumeGomez:


build_system/src/config.rs
```diff
@@ -273,6 +282,9 @@ impl ConfigInfo {
     }
 
     pub fn setup_gcc_path(&mut self) -> Result<(), String> {
+        if self.gcc_path.is_some() {
```
antoyo, line 285: Please add a comment to explain that this means the path was provided via `--gcc-path`, so the config file is not needed in this case.

build_system/src/build.rs
```diff
@@ -184,8 +184,9 @@ pub fn build_sysroot(env: &HashMap<String, String>, config: &ConfigInfo) -> Resu
 fn build_codegen(args: &mut BuildArg) -> Result<(), String> {
     let mut env = HashMap::new();
 
-    env.insert("LD_LIBRARY_PATH".to_string(), args.config_info.gcc_path.clone());
-    env.insert("LIBRARY_PATH".to_string(), args.config_info.gcc_path.clone());
+    let gcc_path = args.config_info.gcc_path.clone().unwrap();
```
antoyo, line 187: Please use `expect()` here and in the other places or add a comment to explain that this cannot happen since an error would be emitted in `config.rs` if the path is not specified somehow.

build_system/src/config.rs
antoyo: Would it make sense to also print a message saying "--gcc-path was provided, ignoring config.toml"?

build_system/src/config.rs
```diff
@@ -299,10 +311,11 @@ impl ConfigInfo {
     ) -> Result<(), String> {
         env.insert("CARGO_INCREMENTAL".to_string(), "0".to_string());
 
-        if self.gcc_path.is_empty() && !use_system_gcc {
+        if self.gcc_path.is_none() && !use_system_gcc {
             self.setup_gcc_path()?;
         }
-        env.insert("GCC_PATH".to_string(), self.gcc_path.clone());
+        let gcc_path = self.gcc_path.clone().expect("GCC path wasn't provided");
```
antoyo, line 317: Just to make sure, since it's still not clear to me, this `expect()` is never supposed to fail because we check for errors in `config.rs`, right?
If so, please change to something like this:
```suggestion
        let gcc_path = self.gcc_path.clone().expect("The config module should have emitted an error if the GCC path wasn't provided");
```

build_system/src/build.rs
GuillaumeGomez: I forgot to save the file... XD

build_system/src/config.rs
GuillaumeGomez: We don't display a message by default in this path, but I can display on in both cases to know where it comes from.

antoyo: Well, since this argument skips some code that usually run, I believe it would be useful to know that this is skipped.
I guess you can print a message in the other case, but I believe it's less useful since the behavior is less surprising.

GuillaumeGomez:
Fixed the `unwrap` I forgot and I added messages to know the GCC path used by the build script.

GuillaumeGomez:
Updated all `expect` messages.

antoyo (approved):

antoyo:
Thanks for your contribution!

PR #612, opened 2025-01-27
Make globals always have 2+ chars as suffix

fluxchief:
As discussed on chat, there currently is the bug where certain suffixes are interpreted by the (m68k) assembler.

Example: `move.l #global.w,-44(%fp)`
`.w` is interpreted by the assembler as a size hint for #global.


GuillaumeGomez (approved):

antoyo:
Thanks for your contribution!

PR #621, opened 2025-02-06
fix: handle pointee_align

### LINKED ISSUE #606: Handle pointee_align
I noticed we don't handle [`pointee_align`](https://github.com/rust-lang/rust/blob/c074d8eee765cfd64e6e143d2894c85c7f3ddc1d/compiler/rustc_target/src/callconv/mod.rs#L148).

g4titanx:
this pr fixes #606


src/abi.rs
```diff
@@ -138,6 +138,13 @@ impl<'gcc, 'tcx> FnAbiGccExt<'gcc, 'tcx> for FnAbi<'tcx, Ty<'tcx>> {
             if attrs.regular.contains(rustc_target::abi::call::ArgAttribute::NonNull) {
                 non_null_args.push(arg_index as i32 + 1);
             }
+            if let Some(align) = attrs.pointee_align {
```
antoyo, line 141: I don't believe this is correct to put this here since it will be ignored when optimizations are disabled.

PR #627, opened 2025-02-12
handle NaN in unordered comparisons

### LINKED ISSUE #75: Incorrect `f32::NAN as u64/u128` conversion
The code generated for `black_box(f32::NAN) as u64/u128` converts it to some junk, but correct result is `0`. This happens only with `f32::NAN` and only when converted to `u64/u128`. Conversion into `i64/i128` works fine, as well as conversions into`u8, i8, ... u32,i32`.

### LINKED ISSUE #626: Unordered float comparisons are emitted as normal, ordered comparisons
In other words, I've checked and ordered vs. non-ordered is not respected:

https://github.com/rust-lang/rustc_codegen_gcc/blob/cb499a29a8d26cf58e4b26ea70fd49c9486e7528/src/builder.rs#L2391-L2413

That means that various operations involving `NaN` behave incorrectly: the unordered comparisons always return true when one of the arguments is `NaN`, while the ordered ones return false. The easiest way I've found to verify the behavior is the expression `f64::NAN as u64`, which should evaluate to `0u64`, but the GCC backend in debug mode evaluates to some large constant.

## When does this actually cause issues

From what I can tell, unordered comparisons, today, are used in two cases:

- `RealULT`: in the `rustc_codegen_gcc` function `fptoint_sat`
- `RealUNE`: in the lowering of `!=` on floats (in `bin_op_to_fcmp_predicate` in `rustc_codegen_ssa`)

The behavior of `!=` seems correct to me, so the only problematic case is `fptoint_sat`

## So, now what?

It looks like GCC just does not have the unordered comparisons

https://github.com/gcc-mirror/gcc/blob/3880271e94b7598b4f5d98c615b7fcddddee6d4c/gcc/jit/libgccjit.h#L1263-L1282

My suggestion is to map these operators to unreachable (they are not used today besides in `fptoint_sat`)

    RealPredicate::RealUGT => unreachable!(),
    RealPredicate::RealUGE => unreachable!(),
    RealPredicate::RealULT => unreachable!(),
    RealPredicate::RealULE => unreachable!(),

and in `fptoint_sat`, use the logic currently only used for signed integers also for unsigned integers, to correct the nan case

https://github.com/rust-lang/rustc_codegen_gcc/blob/cb499a29a8d26cf58e4b26ea70fd49c9486e7528/src/builder.rs#L1883-L1892

Alternatively, `fcmp` could correct the result for unordered operations

```rust
    fn fcmp(&mut self, op: RealPredicate, lhs: RValue<'gcc>, rhs: RValue<'gcc>) -> RValue<'gcc> {
        let correct_for_nan = match op {
            RealPredicate::RealPredicateFalse => unreachable!(),
            RealPredicate::RealOEQ => false,
            RealPredicate::RealOGT => false,
            RealPredicate::RealOGE => false,
            RealPredicate::RealOLT => false,
            RealPredicate::RealOLE => false,
            RealPredicate::RealONE => false,
            RealPredicate::RealORD => unreachable!(),
            RealPredicate::RealUNO => unreachable!(),
            RealPredicate::RealUEQ => false,
            RealPredicate::RealUGT => true,
            RealPredicate::RealUGE => true,
            RealPredicate::RealULT => true,
            RealPredicate::RealULE => true,
            RealPredicate::RealUNE => false,
            RealPredicate::RealPredicateTrue => unreachable!(),
        };

        let cmp = self.context.new_comparison(self.location, op.to_gcc_comparison(), lhs, rhs);

        if correct_for_nan {
            let is_nan_lhs = self.fcmp(RealPredicate::RealOEQ, lhs, lhs);
            let is_nan_rhs = self.fcmp(RealPredicate::RealOEQ, rhs, rhs);

            let is_nan = self.new_binary_op(
                self.location,
                BinaryOp::LogicalOr,
                self.cx.bool_type,
                is_nan_lhs,
                is_nan_rhs,
            );

            self.new_binary_op(self.location, BinaryOp::LogicalOr, self.cx.bool_type, is_nan, cmp)
        } else {
            cmp
        }
    }
```

I'd be happy to implement either of these options (or something else, if I'm missing something here)

sidenote: is there a quick way to use `./y test` to run tests in a particular file in `tests/run`? or just all the tests in `tests/`? Currently std is rebuilt every time and many tests build/run, and that's not a great feedback loop.

folkertdev:
fixes #75 
fixes #626 

The `tests/ui/numbers-arithmetic/saturating-float-casts.rs` test now succeeds.


tests/run/float.rs
```diff
@@ -0,0 +1,31 @@
...
+                const EXPECTED: $ty = $expr;
+                assert_eq!($expr, EXPECTED);
+            }
+        };
+    }
+
+    check!(i32, (black_box(0.0f32) as i32));
+
+    check!(i64, (black_box(f64::NAN) as i64));
+    check!(u64, (black_box(f64::NAN) as u64));
```
antoyo, line 27: Does this also work for `black_box(f32::NAN) as u64/u128`, which is what is mentionned in #75?

antoyo (approved):
Looks good to me. Please confirm it also work in the above case.

src/builder.rs
```diff
@@ -1284,7 +1284,53 @@ impl<'a, 'gcc, 'tcx> BuilderMethods<'a, 'tcx> for Builder<'a, 'gcc, 'tcx> {
...
+            RealPredicate::RealUNE => false,
+            RealPredicate::RealPredicateTrue => unreachable!(),
+        };
+
+        let cmp = self.context.new_comparison(self.location, op.to_gcc_comparison(), lhs, rhs);
+
+        if must_handle_nan {
+            // compare a value to itself to check whether it is NaN
+            let is_nan_lhs = self.fcmp(RealPredicate::RealONE, lhs, lhs);
+            let is_nan_rhs = self.fcmp(RealPredicate::RealONE, rhs, rhs);
```
antoyo, line 1314: Would it make more sense to just actually call `self.context.new_comparison()` here?

tests/run/float.rs
folkertdev: it does, I've added those tests just to make sure.

folkertdev:
as an additional note, this still fails:

```rust
#![feature(repr_simd, intrinsics)]

extern "rust-intrinsic" {
    fn simd_as<T, U>(x: T) -> U;
}

#[derive(Copy, Clone)]
#[repr(simd)]
struct V<T>([T; 2]);

fn main() {
    unsafe {
        let f = V::<f32>([f32::MIN, f32::MAX]);
        let i: V<i16> = simd_as(f);
        assert_eq!(i.0[0], f.0[0] as i16);
        assert_eq!(i.0[1], f.0[1] as i16);
    }
}
```

The `i` vector will just be `[0, 0]`. It looks like `convert_vector`, which is used for this conversion internally, does not match the LLVM semantics

https://github.com/rust-lang/gccjit.rs/blob/ac3ffb0b698ea24c9bbf32e3565fd72c0a1e79b7/src/context.rs#L925-L938

antoyo:
> The `tests/ui/numbers-arithmetic/saturating-float-casts.rs` test now succeeds.

In this case, please remove this test from [here](https://github.com/rust-lang/rustc_codegen_gcc/blob/438056890df8696cfa20e0058b8093dfbc3e9ec9/tests/failing-ui-tests.txt#L8).

After that, I'll merge.

Thanks.

Issue #451, opened 2024-02-21
Non-default ABIs aren't supported


sadlerap:
[The following function miscompiles][godbolt]:

```rust
#[no_mangle]
extern "efiapi" fn efi(x: u32, y: u32) -> u32 {
    x * y
}
```

LLVM emits this:

```asm
efi:
        mov     eax, ecx
        imul    eax, edx
        ret
```

GCC emits this (notice the incorrect registers used for parameters):

```asm
efi:
        mov     eax, edi
        imul    eax, esi
        ret
```

As far as I can tell, we don't preserve the declared ABI since libgccjit doesn't have a way to annotate the ABI of functions with an annotated ABI.  This needs to be implemented for `ui/asm/x86_64/multiple-clobber-abi.rs` to pass, which requires both the `sysv64` and `win64` extern abi to be supported.

[godbolt]: https://godbolt.org/z/vsvYosnbz

antoyo:
Yeah, for that, we would need to support a few new [function attributes](https://gcc.gnu.org/onlinedocs/gcc/x86-Function-Attributes.html). If you want to take a look, you would need to add the attributes in a few places:

 * https://github.com/antoyo/gcc/blob/8e0113788aca4824e39acbc85237ab275075835d/gcc/jit/libgccjit.h#L2111
 * https://github.com/antoyo/gcc/blob/8e0113788aca4824e39acbc85237ab275075835d/gcc/jit/jit-recording.cc#L4822
 * https://github.com/antoyo/gcc/blob/8e0113788aca4824e39acbc85237ab275075835d/gcc/jit/jit-playback.cc#L523

antoyo:
I believe we [now support most ABIs](https://github.com/rust-lang/rustc_codegen_gcc/blob/5c50a42bab47dd2ff206cc6e5a9b53909004ff5a/src/abi.rs#L241) and the Godbolt example now compiles properly.

PR #628, opened 2025-02-12
fix clobbered or lateout registers overlapping with input registers

folkertdev:
Apparently, in GCC, input registers cannot also be clobbered. 

this previously errored https://godbolt.org/

this fixes half of `tests/ui/asm/x86_64/multiple-clobber-abi.rs`, but it also runs into https://github.com/rust-lang/rustc_codegen_gcc/issues/451


GuillaumeGomez (approved):

antoyo (commented):
Looks good to me.
There are possibly two things to change.

antoyo:
> this fixes half of `tests/ui/asm/x86_64/multiple-clobber-abi.rs`, but it also runs into #451

I implemented the ABIs used in this test in #629, so rebasing should allow this test to pass.

folkertdev:
locally I can't quite get that to work, but maybe something else needs to be updated?
```
---- [ui] tests/ui/asm/x86_64/multiple-clobber-abi.rs stdout ----
Saved the actual stderr to "/home/folkertdev/rust/rustc_codegen_gcc/build/rust/build/x86_64-unknown-linux-gnu/test/ui/asm/x86_64/multiple-clobber-abi/multiple-clobber-abi.stderr"
normalized stderr:
gcc_jit_function_add_attribute: attribute should be a `gcc_jit_fn_attribute` enum value
gcc_jit_function_add_attribute: attribute should be a `gcc_jit_fn_attribute` enum value
```

I did remove the line to see whether CI maybe did update correctly already.

antoyo:
Oh, I forgot to update [this file](https://github.com/rust-lang/rustc_codegen_gcc/blob/master/libgccjit.version) with the latest commit.
Do you use the default `config.toml` which downloads `libgccjit.so`?

folkertdev:
I think so yes

src/asm.rs
```diff
@@ -183,25 +194,39 @@ impl<'a, 'gcc, 'tcx> AsmBuilderMethods<'tcx> for Builder<'a, 'gcc, 'tcx> {
...
+                            if input_registers.contains(&reg_name) {
+                                // the `clobber_abi` operand is converted into a series of
+                                // `lateout("reg") _` operands. Of course, a user could also
+                                // explicitly define such an output operand.
+                                //
+                                // GCC does not allow input registers to be clobbered, so if this out register
+                                // is also used as an in register, do not add it to the clobbers list.
+                                // it will be treated as a lateout register with `out_place: None`
+                                if !late {
+                                    bug!("input registers can only be used as lateout regisers");
```
antoyo, line 206: Oh sorry: I was wrong here. `bug!` will also abort the compiler.
Please use `create_err()` like is done [here](https://github.com/rust-lang/rustc_codegen_gcc/blob/b08a0c6afc33d16f530f379828bcc06bffc3dee8/src/asm.rs#L121).

folkertdev: hmm, we don't expect this to ever be hit, so is a user-facing message really needed then? I think it really is a bug if this were hit.

folkertdev:
> Could you please update the file with the latest commit from [the gcc repo](https://github.com/rust-lang/gcc) and see if this works? If so, please add this change to this PR.

yes that does work, I've added it now.

src/asm.rs
antoyo: Oh I misunderstood, then. Keep it as `bug!` in this case.

antoyo:
Is there any other change you would like to make or is this ready to merge?

folkertdev:
no I think we're good

antoyo:
Thanks for your contribution!

Issue #485, opened 2024-04-04
x86_64 asm: Build error when using r[8-15]b


taiki-e:
The following code passes build with cg_llvm, but fails with cg_gcc.
```rs
#[inline(never)]
pub unsafe fn f() {
    core::arch::asm!(
        "",
        out("al") _,
        out("bl") _,
        out("cl") _,
        out("dl") _,
        out("sil") _,
        out("dil") _,
        // out("bpl") _, // frame pointer
        out("r8b") _,
        out("r9b") _,
        out("r10b") _,
        out("r11b") _,
        out("r12b") _,
        out("r13b") _,
        out("r14b") _,
        out("r15b") _,
    );
}
```

The following is the error message when building with cg_gcc.

```
libgccjit.so: error: unknown register name 'r15b' in 'asm'
libgccjit.so: error: unknown register name 'r14b' in 'asm'
libgccjit.so: error: unknown register name 'r13b' in 'asm'
libgccjit.so: error: unknown register name 'r12b' in 'asm'
libgccjit.so: error: unknown register name 'r11b' in 'asm'
libgccjit.so: error: unknown register name 'r10b' in 'asm'
libgccjit.so: error: unknown register name 'r9b' in 'asm'
libgccjit.so: error: unknown register name 'r8b' in 'asm'
error: could not copy "/app/output.example.fb61ce7e3f5ff483-cgu.0.rcgu.s" to "/app/output.s": No such file or directory (os error 2)
```

[playground](https://godbolt.org/z/rf5WMa1aT)

I could reproduce this with both playground and the latest master branch (75f0ab5c55aeaee52a794b02a0d7d7aaf70c4bed).

Reference: https://doc.rust-lang.org/nightly/reference/inline-assembly.html#register-operands

> | Architecture | Register class | Registers | LLVM constraint code |
> | ------------ | -------------- | --------- | -------------------- |
> | x86-64 | `reg_byte`\* | `al`, `bl`, `cl`, `dl`, `sil`, `dil`, `bpl`, `r[8-15]b` | `q` |

antoyo:
Thanks for the bug report!

It seems GCC doesn't support the `b` prefix, so we'll need to convert those to the GCC way.

folkertdev:
I wrote a small example that, i believe, shows that no changes happen to the upper bits of the register when LLVM is used: they are not explicitly cleared when `reg_byte` is used, it is just UB to rely on the upper bits.

[...]

PR #630, opened 2025-02-13
support `reg_byte` registers

folkertdev:
fixes https://github.com/rust-lang/rustc_codegen_gcc/issues/485

I believe (see the issue) that the `b` suffix can just be trimmed, without causing invalid behavior.

I also refactored a bit to reduce the level of nesting, given that we probably need much more logic here to support more register classes that don't translate 1-to-1 between llvm and gcc. Hopefully the rustc assembly tests will guide us towards the edge cases.

### DIFF
--- a/src/asm.rs
+++ b/src/asm.rs
@@ -611,114 +611,127 @@ fn estimate_template_length(
 }
 
 /// Converts a register class to a GCC constraint code.
-fn reg_to_gcc(reg: InlineAsmRegOrRegClass) -> ConstraintOrRegister {
-    let constraint = match reg {
-        // For vector registers LLVM wants the register name to match the type size.
+fn reg_to_gcc(reg_or_reg_class: InlineAsmRegOrRegClass) -> ConstraintOrRegister {
+    match reg_or_reg_class {
         InlineAsmRegOrRegClass::Reg(reg) => {
-            match reg {
-                InlineAsmReg::X86(_) => {
-                    // TODO(antoyo): add support for vector register.
-                    //
-                    // // For explicit registers, we have to create a register variable: https://stackoverflow.com/a/31774784/389119
-                    return ConstraintOrRegister::Register(match reg.name() {
-                        // Some of registers' names does not map 1-1 from rust to gcc
-                        "st(0)" => "st",
+            ConstraintOrRegister::Register(explicit_reg_to_gcc(reg))
+        }
+        InlineAsmRegOrRegClass::RegClass(reg_class) => {
+            ConstraintOrRegister::Constraint(reg_class_to_gcc(reg_class))
+        }
+    }
+}
 
-                        name => name,
-                    });
+fn explicit_reg_to_gcc(reg: InlineAsmReg) -> &'static str {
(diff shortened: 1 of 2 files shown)


src/asm.rs
```diff
@@ -611,114 +611,126 @@ fn estimate_template_length(
 }
 
 /// Converts a register class to a GCC constraint code.
-fn reg_to_gcc(reg: InlineAsmRegOrRegClass) -> ConstraintOrRegister {
-    let constraint = match reg {
-        // For vector registers LLVM wants the register name to match the type size.
```
folkertdev, line 616: this comment is lost right now, I'm not sure what it was supposed to say, and why what llvm does is relevant?

antoyo: Oh, that was copy-pasted from the LLVM codegen, but I forgot to change that to GCC.

folkertdev: ok, but what does this comment mean? What type is talked about here? it just doesn't seem like the right place for this comment?

antoyo: It comes from [here](https://github.com/rust-lang/rust/blob/6ac88786155774da429213a1362d4cb085047af3/compiler/rustc_codegen_llvm/src/asm.rs#L587) originally.
Maybe the type is the `layout` parameter that we don't have in cg_gcc.

folkertdev: yeah I think that's right, e.g. 

```rust
  match layout.size.bytes() {
                        64 => 'z',
                        32 => 'y',
                        _ => 'x',
```

this really does pick the register width based on a type. Do you think GCC will have to do the same? So far that does not seem necessary.

antoyo: I do not know. I would need to check if needed.

src/asm.rs
```diff
@@ -611,114 +611,126 @@ fn estimate_template_length(
...
-                        name => name,
-                    });
+fn explicit_reg_to_gcc(reg: InlineAsmReg) -> &'static str {
+    // For explicit registers, we have to create a register variable: https://stackoverflow.com/a/31774784/389119
+    match reg {
+        InlineAsmReg::X86(reg) => {
+            // TODO(antoyo): add support for vector register.
+            match reg.reg_class() {
+                X86InlineAsmRegClass::reg_byte => {
+                    // see https://github.com/rust-lang/rustc_codegen_gcc/issues/485
```
antoyo, line 632: Could you also explain in the comment why we do this?

tests/run/asm.rs
```diff
@@ -205,6 +205,29 @@ fn asm() {
...
+    unsafe {
+        core::arch::asm!(
+            "",
+            out("al") _,
+            out("bl") _,
+            out("cl") _,
+            out("dl") _,
+            out("sil") _,
+            out("dil") _,
+            // out("bpl") _, // frame pointer
```
antoyo, line 220: Can this line be removed?

antoyo (approved):

antoyo:
Is this ready to be merged?

folkertdev:
yes

antoyo:
Thanks for your contribution!

PR #635, opened 2025-02-21
fix 128bits ctlz intrinsincs UB

### LINKED ISSUE #604: UB in the implementation of the 128 bit ctlz intrinsic
*Originally discovered by running compiler tests with `-Cllvm-args=santize-undefined`.*

Running the following piece of code:
```rust
#[no_mangle]
fn ub()->u32{
    black_box(1_u128).leading_zeros()
}
```
will trip UB checks. This is caused by an improper implementation of the 128 bit ctlz intrinsic.  Currently, this intrinsic first checks if the argument is not 0, and then counts the leading zeroes in the high and low 8 bytes of the value - separately. 

This is incorrect: if either the low xor the high bytes are zero, then counting the leading zeroes in those 8 byte numbers is UB.

Here is the relevant part of GIMPLE, which shows the underlying issue:
```gimple
  if (param0 == 0) goto then; else goto else;
      then:
      zeros = 128;
      goto after;
      else:
      _2 = param0 >> 64;
      _3 = (size_t) _2;
      _4 = __builtin_clzll (_3);
      _5 = (uint128_t) _4;
      count_loading_zeroes_results[0] = _5;
      _6 = (size_t) param0;
      _7 = __builtin_clzll (_6);
```
In the case of `param0 = 1_u128`, the top `if`  will jump to `else`. There, the value `_2` will be zero(the high 8 bytes of 1_u128 are zero). This means that the value of `_3` will also be zero, and on this line:
```gimple
_4 = __builtin_clzll (_3);
```
`__builtin_clzll` will be called with an argument of `0`, which is UB.

dvermd:
Fixes #604

This is the GIMPLE code generated
```
      if (param0 == 0) goto then; else goto else;
      then:
      zeros = 128;
      goto after;
      else:
      _2 = param0 >> 64;
      _3 = (size_t) _2;
      if (_3 != 0) goto ctlz_then; else goto ctlz_else;
      after:
      stack_var_1.15_4 = &stack_var_1;
      MEM[(unsigned int *)stack_var_1.15_4] = zeros;
      stack_var_1.16_5 = &stack_var_1;
      loadedValue2 = MEM[(unsigned int *)stack_var_1.16_5];
      D.4444 = loadedValue2;
      return D.4444;
      ctlz_then:
      _6 = param0 >> 64;
      _7 = (size_t) _6;
      _8 = __builtin_clzll (_7);
      zeros = (unsigned int) _8;
      goto ctlz_after;
      ctlz_else:
      _9 = (size_t) param0;
      _10 = __builtin_clzll (_9);
      _11 = (unsigned int) _10;
      zeros = _11 + 64;
      goto ctlz_after;
      ctlz_after:
      zeros = zeros;
      goto after;
```

### DIFF
--- a/src/intrinsic/mod.rs
+++ b/src/intrinsic/mod.rs
@@ -726,51 +726,42 @@ impl<'a, 'gcc, 'tcx> Builder<'a, 'gcc, 'tcx> {
                 "__builtin_clzll"
             }
             else if width == 128 {
-                // Algorithm from: https://stackoverflow.com/a/28433850/389119
-                let array_type = self.context.new_array_type(None, arg_type, 3);
-                let result = self.current_func()
-                    .new_local(None, array_type, "count_loading_zeroes_results");
-
+                // Algorithm from: https://stackoverflow.com/a/28433850/389119 updated to check for high 64bits being 0
+                let result = self.current_func().new_local(None, self.u32_type, "zeros");
+                let ctlz_then_block = self.current_func().new_block("ctlz_then");
+                let ctlz_else_block = self.current_func().new_block("ctlz_else");
+                let ctlz_after_block = self.current_func().new_block("ctlz_after");
                 let sixty_four = self.const_uint(arg_type, 64);
                 let shift = self.lshr(arg, sixty_four);
                 let high = self.gcc_int_cast(shift, self.u64_type);
-                let low = self.gcc_int_cast(arg, self.u64_type);
-
-                let zero = self.context.new_rvalue_zero(self.usize_type);
-                let one = self.context.new_rvalue_one(self.usize_type);
-                let two = self.context.new_rvalue_from_long(self.usize_type, 2);
 
                 let clzll = self.context.get_builtin_function("__builtin_clzll");
 
-                let first_elem = self.context.new_array_access(None, result, zero);
-                let first_value = self.gcc_int_cast(self.context.new_call(None, clzll, &[high]), arg_type);
-                self.llbb()
-                    .add_assignment(self.location, first_elem, first_value);
+                let zero = self.const_uint(high.get_type(), 0);
+                let cond = self.gcc_icmp(IntPredicate::IntNE, high, zero);
+                self.llbb().end_with_conditional(
+                    self.location,
+                    cond,
+                    ctlz_then_block,
+                    ctlz_else_block,
+                );
 
-                let second_elem = self.context.new_array_access(self.location, result, one);
-                let cast = self.gcc_int_cast(self.context.new_call(self.location, clzll, &[low]), arg_type);
-                let second_value = self.add(cast, sixty_four);
-                self.llbb()
-                    .add_assignment(self.location, second_elem, second_value);
+                let leading_zeroes =
+                    self.gcc_int_cast(self.context.new_call(None, clzll, &[high]), result_type);
 
-                let third_elem = self.context.new_array_access(self.location, result, two);
-                let third_value = self.const_uint(arg_type, 128);
-                self.llbb()
-                    .add_assignment(self.location, third_elem, third_value);
+                ctlz_then_block.add_assignment(None, result, leading_zeroes);
+                ctlz_then_block.end_with_jump(None, ctlz_after_block);
 
-                let not_high = self.context.new_unary_op(self.location, UnaryOp::LogicalNegate, self.u64_type, high);
-                let not_low = self.context.new_unary_op(self.location, UnaryOp::LogicalNegate, self.u64_type, low);
-                let not_low_and_not_high = not_low & not_high;
-                let index = not_high + not_low_and_not_high;
-                // NOTE: the following cast is necessary to avoid a GIMPLE verification failure in
-                // gcc.
-                // TODO(antoyo): do the correct verification in libgccjit to avoid an error at the
-                // compilation stage.
-                let index = self.context.new_cast(self.location, index, self.i32_type);
+                let low = self.gcc_int_cast(arg, self.u64_type);
+                let low_leading_zeroes =
+                    self.gcc_int_cast(self.context.new_call(None, clzll, &[low]), result_type);
+                let leading_zeroes = self.add(low_leading_zeroes, sixty_four);
+                ctlz_else_block.add_assignment(None, result, leading_zeroes);
+                ctlz_else_block.end_with_jump(None, ctlz_after_block);
 
-                let res = self.context.new_array_access(self.location, result, index);
+                self.switch_to_block(ctlz_after_block);
 
-                return self.gcc_int_cast(res.to_rvalue(), result_type);
+                return result.to_rvalue();
             }
             else {
                 let count_leading_zeroes = self.context.get_builtin_function("__builtin_clzll");


dvermd:
It looks like the some CI jobs are failing on 
```
thread 'rustc' panicked at src/int.rs:176:13:
assertion failed: a_type.dyncast_array().is_some()
```
which is some code I didn't touch.

Can my changes have some side effects ?
I see that #631 also have these jobs failing. Perhaps it's not my changes which bring these failures.

Any hints ?

antoyo:
Not sure it's related. I opened a new dummy [PR](https://github.com/rust-lang/rustc_codegen_gcc/pull/636) to check if the tests pass, though.

antoyo:
The [CI](https://github.com/rust-lang/rustc_codegen_gcc/pull/636) passed, so the problem is in here.

I suspect this is because you changed the return type: it used to return an array type and your change makes it return a `u32`.

dvermd:
The return type of the `if width == 128` was already a u32 because the block ended with
```
                let res = self.context.new_array_access(self.location, result, index);

                return self.gcc_int_cast(res.to_rvalue(), result_type);
```
which will get an element in the array and cast it as result_type.

dvermd:
Ok, finally found the line pointing in my change `let leading_zeroes = self.add(low_leading_zeroes, sixty_four);` where `low_leading_zeroes`and `sixty_four` are of different types.

The problem I see is that the [if branch](https://github.com/rust-lang/rustc_codegen_gcc/blob/fcac22978ad507b8a7124f0a1600ff55417a34a7/src/int.rs#L162) should have been taken.

src/intrinsic/mod.rs
```diff
@@ -726,51 +726,43 @@ impl<'a, 'gcc, 'tcx> Builder<'a, 'gcc, 'tcx> {
                 "__builtin_clzll"
             }
             else if width == 128 {
-                // Algorithm from: https://stackoverflow.com/a/28433850/389119
-                let array_type = self.context.new_array_type(None, arg_type, 3);
-                let result = self.current_func()
-                    .new_local(None, array_type, "count_loading_zeroes_results");
-
+                // Algorithm from: https://stackoverflow.com/a/28433850/389119 updated to check for high 64bits being 0
```
antoyo, line 729: It seems your new algorithm significantly diverges from the one linked here.
If so, could you please provide the link to the algorithm you used? If you can't, could you please add comments below to make this easier to understand?

antoyo (commented):
Sorry for the delay: I've been very busy.

dvermd:
Hi, sorry for the delay, I missed your reply notification and got caught up by IRL stuff.

I'm going to add a comment on the new algorithm which is:
```
// if arg is 0 return 128
// else if the 64 high bits of arg are not 0, return clzll(64 high bits of arg)
//      else return 64 + clzll(64 low bits of arg)
```
This adds an if to the generated code but we need to check for 0 before calling `ctzll`

I think there is also a problem of UB in [intrinsic/simd.rs](https://github.com/rust-lang/rustc_codegen_gcc/blob/master/src/intrinsic/simd.rs#L359) 
```rust
                let index = bx.context.new_rvalue_from_long(bx.i32_type, i as i64);
                let value = bx.extract_element(vector, index).to_rvalue();
                let value_type = value.get_type();
                let element = if name == sym::simd_ctlz {
                    bx.count_leading_zeroes(value_type.get_size() as u64 * 8, value)
                } else {
                    bx.count_trailing_zeroes(value_type.get_size() as u64 * 8, value)
                };
```

where 0 is not special cased like in [intrinsic/mod.rs](https://github.com/rust-lang/rustc_codegen_gcc/blob/master/src/intrinsic/mod.rs#L413).


So I'll move the 0 special case handling in `count_leading_zeroes` and remove it from `codegen_intrinsic_call`

As found by @FractalFir, the same problems apply to trailing_zeros / cttz

dvermd:
There is still a diff in the handling of clz and ctz intrinsics in the `if` to select the first builtin function

```rust
    fn count_leading_zeroes(&mut self, width: u64, arg: RValue<'gcc>) -> RValue<'gcc> {
...
            if arg_type.is_uint(self.cx) {
                "__builtin_clz"
            }
```
vs 
```rust
    fn count_trailing_zeroes(&mut self, _width: u64, arg: RValue<'gcc>) -> RValue<'gcc> {
...
            if arg_type.is_uchar(self.cx) || arg_type.is_ushort(self.cx) || arg_type.is_uint(self.cx) {
                // NOTE: we don't need to & 0xFF for uchar because the result is undefined on zero.
                ("__builtin_ctz", self.cx.uint_type)
            }
```

They both have the same comment
```rust
            // TODO(antoyo): write a new function Type::is_compatible_with(&Type) and use it here
            // instead of using is_uint().
```

Should I add the `uchar` and `ushort` types to the leading zeros test ?

dvermd:
@antoyo Do you have any hints to reproduce the `--without-int128` environment locally ? I can't figure out by myself what makes the CI fail.

antoyo:
> @antoyo Do you have any hints to reproduce the `--without-int128` environment locally ? I can't figure out by myself what makes the CI fail.

If you compile gcc yourself, you can apply [this patch](https://github.com/rust-lang/gcc/blob/master/patches/0001-Disable-128-bit-integers-for-testing-purposes.patch).

Otherwise, you can try setting [this variable](https://github.com/rust-lang/rustc_codegen_gcc/blob/2b640216aea9527c67c82cfc781d979247ac6028/src/base.rs#L207) to false, but I'm not sure if this will have the same behavior.

src/intrinsic/mod.rs
```diff
@@ -886,82 +857,159 @@ impl<'a, 'gcc, 'tcx> Builder<'a, 'gcc, 'tcx> {
...
-        self.context.new_cast(self.location, res, result_type)
+            let sixty_four = self.const_uint(arg_type, 64);
+            let shift = self.lshr(arg, sixty_four);
+            let high = self.gcc_int_cast(shift, self.u64_type);
+
+            let clzll = self.context.get_builtin_function("__builtin_clzll");
+
+            let zero_hi = self.const_uint(high.get_type(), 0);
+            let cond = self.gcc_icmp(IntPredicate::IntNE, high, zero_hi);
+            block.end_with_conditional(self.location, cond, ctlz_then_block, ctlz_else_block);
```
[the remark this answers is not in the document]
dvermd: Well, the condition must be created on the else_block that is passed as an argument like in [gcc_icmp](https://github.com/rust-lang/rustc_codegen_gcc/blob/master/src/int.rs#L450) for example.

But that gave me a hint. 

If I make the code look like
```rust 
if width == 128 && !self.cx.supports_128bit_integers {
    // this code will exhibit UB when arg is 0
    self.count_leading_zeroes_nonzero_impl(...) 
} else {
    let result_type = self.u32_type;
    let result = self.current_func().new_local(None, result_type, "zeros");

    let then_block = self.current_func().new_block("zero_then");
    let else_block = self.current_func().new_block("zero_else");
    let after_block = self.current_func().new_block("zero_after");

    let zero = self.cx.const_uint(arg.get_type(), 0);
    let cond = self.gcc_icmp(IntPredicate::IntEQ, arg, zero);

    self.llbb().end_with_conditional(None, cond, then_block, else_block);

    let zero_result = self.cx.gcc_uint(self.u32_type, width);
    then_block.add_assignment(None, result, zero_result);
    then_block.end_with_jump(None, after_block);
    self.switch_to_block(else_block);

    let res = self.count_leading_zeroes_nonzero_impl(width, arg);
    else_block.add_assignment(None, result, res);
    else_block.end_with_jump(None, after_block);
    self.switch_to_block(after_block);
    result.to_rvalue()
}
``` 

The compilation pass. So, you're right, there must be something with the comparison of the 128bit arg when the context doesn't support 128bit integers.

What's weird is that I made exactly the same change on the `count_trailing_zeros` function and it compiles fine. 
What's even weirder is that when I copy and rename `count_trailing_zeros*` functions to `count_leading_zeros`, I get the compiler error.

That leads me to think there is something elsewhere that treats `count_trailing_zeros` and `count_leading_zeros` differently regarding 128bit integers.

Making a special case for 128bit integer on platform without 128bit integers doesn't look like a good idea and I'd prefer to find the real root of the error.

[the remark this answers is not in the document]
antoyo: What I meant is that `gcc_icmp` will not only create new blocks, but will end the current block because it has conditions.
In order to workaround this issue, you need to use `llbb()` to get the new current block that will still be in your `else` branch.

dvermd:
I finally made through it once I realized `lshr` ends the current block just like `gcc_icmp`. I moved it out of the `then` and `else` blocks, thus it is computed even when not used.

src/intrinsic/mod.rs
```diff
@@ -992,50 +1010,43 @@ impl<'a, 'gcc, 'tcx> Builder<'a, 'gcc, 'tcx> {
...
-                let not_low = self.context.new_unary_op(self.location, UnaryOp::LogicalNegate, self.u64_type, low);
-                let not_high = self.context.new_unary_op(self.location, UnaryOp::LogicalNegate, self.u64_type, high);
-                let not_low_and_not_high = not_low & not_high;
-                let index = not_low + not_low_and_not_high;
-                // NOTE: the following cast is necessary to avoid a GIMPLE verification failure in
-                // gcc.
-                // TODO(antoyo): do the correct verification in libgccjit to avoid an error at the
-                // compilation stage.
-                let index = self.context.new_cast(self.location, index, self.i32_type);
+                let result = self.current_func()
```
antoyo: There is also a lot duplication with `count_leading_zeroes_nonzero` here, but I'm not sure a refactoring would make it clearer since we need to use high first in one method and low first in the other.
So, I'll let you judge if a refactoring would help clarity here.

src/intrinsic/mod.rs
```diff
@@ -886,20 +858,43 @@ impl<'a, 'gcc, 'tcx> Builder<'a, 'gcc, 'tcx> {
     }
 
     fn count_leading_zeroes(&mut self, width: u64, arg: RValue<'gcc>) -> RValue<'gcc> {
+        let func = self.current_func();
```
antoyo: Please add a comment to this method to describe what it does (return 0 if arg is zero, otherwise call the leading_zero builtin).

antoyo (commented):
This looks good!
Again, I'm very sorry for the long delay it took me to review this.
Thanks a lot for your contribution.

src/intrinsic/mod.rs
dvermd: I can see many ways to factorize this code:

1. Add a boolean and several ifs to switch between trailing and leading zeros
2. Create an enum with functions to implement specific behaviour for trailing or leading zeros
3. Create a macro and generate 2 different functions

Any preference on your side on which refactoring path to take ?

dvermd:
There is a compilation error on m68k because the `__ctzdi2` symbol is not found. From my research it looks like adding `-lgcc` would fix the error. Is this something we want for the tests ?

I also tried to add `-C llvm-args=sanitize-undefined` and `-C link-args=-lubsan` to compiler args to make the test fail if the UB is re-introduced but m68k doesn't find libubsan.so so I remove it for now.

antoyo:
> There is a compilation error on m68k because the `__ctzdi2` symbol is not found. From my research it looks like adding `-lgcc` would fix the error. Is this something we want for the tests ?

Correct me if I'm wrong, but I was under the impression that a previous version of this PR passed this test. Is this the case?
If so, do you know what change caused this failure?

dvermd:
> > There is a compilation error on m68k because the `__ctzdi2` symbol is not found. From my research it looks like adding `-lgcc` would fix the error. Is this something we want for the tests ?
> 
> Correct me if I'm wrong, but I was under the impression that a previous version of this PR passed this test. Is this the case? If so, do you know what change caused this failure?

I think the last green on m68k was on the commit [97aaccd](https://github.com/rust-lang/rustc_codegen_gcc/commit/97aaccdd5e6b33d5e96bd23271e8a4ec39310e6c) in this [ workflow](https://github.com/rust-lang/rustc_codegen_gcc/actions/runs/16799184499/workflow).

Then I added some tests like you suggested and the error occurred while compiling it.

antoyo:
This intrinsic seems to be in [compiler-builtins](https://github.com/rust-lang/compiler-builtins/blob/eba1a3f3d2824739f07fe434624b0d2fee917d89/compiler-builtins/README.md?plain=1#L66), so I believe if you change the test to use `core` (and possibly `std`), this should work without any flag needed.

tests/lang_tests_common.rs
```diff
@@ -115,7 +115,19 @@ pub fn main_inner(profile: Profile) {
...
             match profile {
-                Profile::Debug => {}
+                Profile::Debug => {
+                    if test_target.is_ok() {
+                        // m68k doesn't have lubsan for now
+                        compiler.args(["-C", "llvm-args=sanitize-undefined"]);
+                    } else {
+                        compiler.args([
+                            "-C",
+                            "llvm-args=sanitize-undefined",
```
antoyo: Are you sending the correct flags here?
I ran the new test on the master branch and it passes while I assumed it should fail on the master branch.

antoyo (commented):
Thanks!
Everything looks good now.
There's only this question left regarding the new test:

antoyo:
And sorry again for taking some time to review this.

PR #647, opened 2025-04-11
Remove -fuse-linker-plugin since we do LTO before the linker is invoked

antoyo:


src/back/lto.rs
```diff
@@ -279,6 +281,15 @@ fn fat_lto(
             context.set_optimization_level(to_gcc_opt_level(config.opt_level));
             context.add_command_line_option("-flto=auto");
             context.add_command_line_option("-flto-partition=one");
+            context.add_driver_option("-flto=auto");
+            context.add_driver_option("-flto-partition=one");
+            // FIXME: we apparently generate fat objects since we had trouble with a sync:
```
antoyo, line 286: Fix this first.

antoyo:
This was attempted again in #775.

Issue #658, opened 2025-04-30
Make GitHub loads diff of `Cargo.lock` by default


antoyo:

albus-droid:
Could we implement this as a GitHub Action that shows diffs in a more readable way?

bjorn3:
Putting `Cargo.lock linguist-generated=false` in `.gitattributes` should be enough: https://github.com/rust-lang/rust/blob/6e23095adf9209614a45f7f75fea36dad7b92afb/.gitattributes#L12

antoyo:
> Could this be implemented as a GitHub Action that shows diffs in a more readable way or is there any other way this can be done

The solution I had in mind was what @bjorn3 mentioned.

albus-droid:
I wasn't aware of this before. Thank you.

antoyo:
Fixed in #659.

PR #659, opened 2025-05-01
Added a .gitattributes file for showing cargo.lock diffs

albus-droid:
This solves rust-lang/rustc_codegen_gcc#658

Also added a line to treat `Cargo.lock` as a text file and normalize eol.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,1 +1,1 @@
-Cargo.lock text eol=lf linguist-generated=false
+Cargo.lock linguist-generated=false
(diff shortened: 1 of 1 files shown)


.gitattributes
```diff
@@ -0,0 +1 @@
+Cargo.lock text eol=lf linguist-generated=false
```
antoyo, line 1:
```suggestion
Cargo.lock linguist-generated=false
```
to match what is done in the [Rust repo](https://github.com/rust-lang/rust/blob/6e23095adf9209614a45f7f75fea36dad7b92afb/.gitattributes#L12)

antoyo:
Thanks for your contribution!

Issue #665, opened 2025-05-09
Create a y.sh command for the tests that are currently ran via "cargo test"


antoyo:
Currently, users need to manually set `LD_LIBRARY_PATH` and `LIBRARY_PATH` and compile the `mini-core` crate via `./y.sh test --mini-tests` before they can run `cargo test`.

I originally thought about documenting this, but I find it better to just create a command that will do this automatically.

FractalFir:
I think this is done now

GuillaumeGomez:
Closing then!

PR #667, opened 2025-05-09
Improved the contrubiting docs

FractalFir:
Added some information about the steps required to run `cargo test`s.


antoyo:
Thanks a lot for sending this PR.

I changed [the issue](https://github.com/rust-lang/rustc_codegen_gcc/issues/665) to mention that I'd prefer to create a `y.sh` command to run those tests, so that this setup is not needed.
Sorry for not thinking about this earlier.

I'm OK with merging this as is, but this will be reverted when that command is created.

FractalFir:
Closed in favour of https://github.com/rust-lang/rustc_codegen_gcc/pull/669

PR #668, opened 2025-05-09
Add a workaround for 128 bit switches

FractalFir:
Due to a libgccjit limitation, 128 bit switches are currently not supported.

As a workaround, this PR:
1. Detects switches on 16 byte(128 bit) values.
2. Replaces such switches with an "if ladder".

Effectively, we turn this:
```rust
let val = black_box(64_u128);
match val {
    0 => return 1,
    1 => return 2,
    64 => (),
    _ => return 3,
}
```
Into this:
```rust
let val = black_box(64_u128);
if val == 0{
  return 1;
}
else if val == 1{
   return 2;
}
else if val ==   64 {
} else {
   return 3; 
}
```
This is less efficient than a switch, but it **only** applies to previously unsupported edge case: 128 bit switches. So, the small perf penalty is acceptable in this case.

Besides that, this PR contains a regression test for this issue, and a few tiny changes to minicore. 

1. Added support for the `black_box` intrinsic - it is strictly necessary for correct tests. **Even now**, MIR optimizations sometimes optimize the crash I test for away.  Using `black_box` guarantees we will not miss any issues in the future. 

2. Added copy impls for `i64`, `u128`, and `i128`. The `u128` impl was needed to properly test 128 bit switches. Adding one for `i128` too seemed reasonable.

### DIFF
--- a/example/mini_core.rs
+++ b/example/mini_core.rs
@@ -1,6 +1,14 @@
 #![feature(
-    no_core, lang_items, intrinsics, unboxed_closures, extern_types,
-    decl_macro, rustc_attrs, transparent_unions, auto_traits, freeze_impls,
+    no_core,
+    lang_items,
+    intrinsics,
+    unboxed_closures,
+    extern_types,
+    decl_macro,
+    rustc_attrs,
+    transparent_unions,
+    auto_traits,
+    freeze_impls,
     thread_local
 )]
 #![no_core]
@@ -35,13 +43,13 @@ impl<T: ?Sized + Unsize<U>, U: ?Sized> CoerceUnsized<*mut U> for *mut T {}
 pub trait DispatchFromDyn<T> {}
 
 // &T -> &U
-impl<'a, T: ?Sized+Unsize<U>, U: ?Sized> DispatchFromDyn<&'a U> for &'a T {}
+impl<'a, T: ?Sized + Unsize<U>, U: ?Sized> DispatchFromDyn<&'a U> for &'a T {}
 // &mut T -> &mut U
-impl<'a, T: ?Sized+Unsize<U>, U: ?Sized> DispatchFromDyn<&'a mut U> for &'a mut T {}
+impl<'a, T: ?Sized + Unsize<U>, U: ?Sized> DispatchFromDyn<&'a mut U> for &'a mut T {}
 // *const T -> *const U
-impl<T: ?Sized+Unsize<U>, U: ?Sized> DispatchFromDyn<*const U> for *const T {}
+impl<T: ?Sized + Unsize<U>, U: ?Sized> DispatchFromDyn<*const U> for *const T {}
 // *mut T -> *mut U
-impl<T: ?Sized+Unsize<U>, U: ?Sized> DispatchFromDyn<*mut U> for *mut T {}
+impl<T: ?Sized + Unsize<U>, U: ?Sized> DispatchFromDyn<*mut U> for *mut T {}
 impl<T: ?Sized + Unsize<U>, U: ?Sized> DispatchFromDyn<Box<U, ()>> for Box<T, ()> {}
 
 #[lang = "legacy_receiver"]
@@ -52,8 +60,7 @@ impl<T: ?Sized> LegacyReceiver for &mut T {}
 impl<T: ?Sized, A: Allocator> LegacyReceiver for Box<T, A> {}
 
 #[lang = "receiver"]
-trait Receiver {
-}
+trait Receiver {}
 
 #[lang = "copy"]
 pub trait Copy {}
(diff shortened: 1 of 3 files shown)


src/builder.rs
```diff
@@ -568,11 +568,28 @@ impl<'a, 'gcc, 'tcx> BuilderMethods<'a, 'tcx> for Builder<'a, 'gcc, 'tcx> {
     ) {
         let mut gcc_cases = vec![];
         let typ = self.val_ty(value);
-        for (on_val, dest) in cases {
-            let on_val = self.const_uint_big(typ, on_val);
-            gcc_cases.push(self.context.new_case(on_val, on_val, dest));
+        // FIXME: This is a workaround for a libgccjit limitation.
```
antoyo, line 571: It is preferred in the Rust project to specify a username, so please use something like this:
```suggestion
        // FIXME(antoyo): This is a workaround for a libgccjit limitation.
```

src/builder.rs
```diff
@@ -568,11 +568,28 @@ impl<'a, 'gcc, 'tcx> BuilderMethods<'a, 'tcx> for Builder<'a, 'gcc, 'tcx> {
...
+        // Currently, libgccjit can't directly create 128 bit intigers.
```
antoyo, line 572:
```suggestion
        // Currently, libgccjit can't directly create 128 bit integers.
```

src/builder.rs
```diff
@@ -568,11 +568,28 @@ impl<'a, 'gcc, 'tcx> BuilderMethods<'a, 'tcx> for Builder<'a, 'gcc, 'tcx> {
...
+        // Since switch cases must be values, and casts are not constant, we can't use 128 bit switch cases.
+        // In such a case, we will simply fall back to an if-ladder.
+        // This *may* be slower than a native switch, but a slow working solution is better than none at all.
+        if typ.is_i128(self) || typ.is_u128(self) {
+            for (on_val, dest) in cases {
+                let on_val = self.const_uint_big(typ, on_val);
+                let is_case =
+                    self.context.new_comparison(None, ComparisonOp::Equals, value, on_val);
```
antoyo, line 580: Please use `self.location` to have better debug info.
```suggestion
                    self.context.new_comparison(self.location, ComparisonOp::Equals, value, on_val);
```

src/builder.rs
```diff
@@ -568,11 +568,28 @@ impl<'a, 'gcc, 'tcx> BuilderMethods<'a, 'tcx> for Builder<'a, 'gcc, 'tcx> {
...
+                let next_block = self.current_func.borrow().unwrap().new_block("");
+                self.block.end_with_conditional(None, is_case, dest, next_block);
```
antoyo, line 582: Same here:
```suggestion
                self.block.end_with_conditional(self.location, is_case, dest, next_block);
```

src/builder.rs
```diff
@@ -568,11 +568,28 @@ impl<'a, 'gcc, 'tcx> BuilderMethods<'a, 'tcx> for Builder<'a, 'gcc, 'tcx> {
...
+                self.block = next_block;
+            }
+            self.block.end_with_jump(None, default_block);
```
antoyo, line 585: And here:
```suggestion
            self.block.end_with_jump(self.location, default_block);
```

src/builder.rs
```diff
@@ -568,11 +568,28 @@ impl<'a, 'gcc, 'tcx> BuilderMethods<'a, 'tcx> for Builder<'a, 'gcc, 'tcx> {
...
+        // Currently, libgccjit can't directly create 128 bit intigers.
+        // Since switch cases must be values, and casts are not constant, we can't use 128 bit switch cases.
+        // In such a case, we will simply fall back to an if-ladder.
+        // This *may* be slower than a native switch, but a slow working solution is better than none at all.
+        if typ.is_i128(self) || typ.is_u128(self) {
+            for (on_val, dest) in cases {
+                let on_val = self.const_uint_big(typ, on_val);
+                let is_case =
+                    self.context.new_comparison(None, ComparisonOp::Equals, value, on_val);
+                let next_block = self.current_func.borrow().unwrap().new_block("");
```
antoyo, line 581: I would give the block a dummy name just to make it clearer in the generated IR.
(We can also use the method `current_func()` to make this simpler.)
```suggestion
                let next_block = self.current_func().new_block("case");
```

tests/run/switchint_128bit.rs
```diff
@@ -0,0 +1,36 @@
...
+
+#![feature(no_core)]
+#![no_std]
+#![no_core]
+#![no_main]
+
+extern crate mini_core;
+use intrinsics::black_box;
+use mini_core::*;
+#[no_mangle]
```
antoyo, line 14:
```suggestion
use mini_core::*;

#[no_mangle]
```

example/mini_core.rs
```diff
@@ -1,6 +1,14 @@
 #![feature(
-    no_core, lang_items, intrinsics, unboxed_closures, extern_types,
-    decl_macro, rustc_attrs, transparent_unions, auto_traits, freeze_impls,
+    no_core,
```
antoyo, line 2: In the future, please put the formatting changes in a different commit to make the review easier.

antoyo (commented):
Thanks a lot!
Nice work as always.

A few nitpicks.

PR #669, opened 2025-05-09
modifed y.sh to allow for running cargo tests.

FractalFir:
Added a new option to `./y.sh test`: ` ./y.sh test --cargo-tests`.

When this option is selected, `y.sh` will build & test minicore(it is a dependency). After that, it will simply call `cargo test` with the selected environment flags.  

For now  ` ./y.sh test --cargo-tests`  ignores `RUSTFLAGS`. 

Normally, `RUSTFLAGS` contain the `-Zcodegen-backend` option, which would make `cg_gcc` rebuild itself. This is undesirable, and ignoring `RUSTFLAGS` altogether was simply the easiest option. 

In the future, we maybe could filter out `-Zcodegen-backend`, and keep the rest of the flags intact.

### DIFF
--- a/build_system/src/test.rs
+++ b/build_system/src/test.rs
@@ -42,7 +42,7 @@ fn get_runners() -> Runners {
     );
     runners.insert("--extended-regex-tests", ("Run extended regex tests", extended_regex_tests));
     runners.insert("--mini-tests", ("Run mini tests", mini_tests));
-
+    runners.insert("--cargo-tests", ("Run cargo tests", cargo_tests));
     runners
 }
 
@@ -88,6 +88,8 @@ struct TestArg {
     use_system_gcc: bool,
     runners: Vec<String>,
     flags: Vec<String>,
+    /// Additional arguments, to be passed to commands like `cargo test`.
+    test_args: Vec<String>,
     nb_parts: Option<usize>,
     current_part: Option<usize>,
     sysroot_panic_abort: bool,
@@ -144,6 +146,7 @@ impl TestArg {
                     show_usage();
                     return Ok(None);
                 }
+                "--" => test_arg.test_args.extend(&mut args),
                 x if runners.contains_key(x)
                     && !test_arg.runners.iter().any(|runner| runner == x) =>
                 {
@@ -201,6 +204,33 @@ fn clean(_env: &Env, args: &TestArg) -> Result<(), String> {
     let _ = remove_dir_all(&args.config_info.cargo_target_dir);
     let path = Path::new(&args.config_info.cargo_target_dir).join("gccjit");
     create_dir(&path)
+}
+
+fn cargo_tests(test_env: &Env, test_args: &TestArg) -> Result<(), String> {
+    // First, we call `mini_tests` to build minicore for us. This ensures we are testing with a working `minicore`,
+    // and that any changes we have made affect `minicore`(since it would get rebuilt).
+    mini_tests(test_env, test_args)?;
+    // Then, we copy some of the env vars from `test_env`
+    // We don't want to pass things like `RUSTFLAGS`, since they contain the -Zcodegen-backend flag.
+    // That would force `cg_gcc` to *rebuild itself* and only then run tests, which is undesirable.
+    let mut env = HashMap::new();
+    env.insert(
+        "LD_LIBRARY_PATH".into(),
+        test_env.get("LD_LIBRARY_PATH").expect("LD_LIBRARY_PATH missing!").to_string(),
+    );
+    env.insert(
+        "LIBRARY_PATH".into(),
+        test_env.get("LIBRARY_PATH").expect("LIBRARY_PATH missing!").to_string(),
+    );
+    env.insert(
+        "CG_RUSTFLAGS".into(),
+        test_env.get("CG_RUSTFLAGS").map(|s| s.as_str()).unwrap_or("").to_string(),
+    );
+    // Pass all the default args + the user-specified ones.
+    let mut args: Vec<&dyn AsRef<OsStr>> = vec![&"cargo", &"test"];
+    args.extend(test_args.test_args.iter().map(|s| s as &dyn AsRef<OsStr>));
+    run_command_with_output_and_env(&args, None, Some(&env))?;
+    Ok(())
 }
 
 fn mini_tests(env: &Env, args: &TestArg) -> Result<(), String> {
@@ -1217,7 +1247,9 @@ fn run_all(env: &Env, args: &TestArg) -> Result<(), String> {
     // asm_tests(env, args)?;
     test_libcore(env, args)?;
     extended_sysroot_tests(env, args)?;
+    cargo_tests(env, args)?;
     test_rustc(env, args)?;
+
     Ok(())
 }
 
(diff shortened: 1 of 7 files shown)


antoyo (commented):
Good job!

Could you please update the doc and the CI to use this new command instead of `cargo test`?

antoyo (commented):
I like how you add a lot of comments.
Thanks!

build_system/src/test.rs
```diff
@@ -203,6 +203,19 @@ fn clean(_env: &Env, args: &TestArg) -> Result<(), String> {
...
+fn cargo_tests(test_env: &Env, args: &TestArg) -> Result<(), String> {
+    // First, we call `mini_tests` to build minicore for us. This ensures we are testing with a working `minicore`,
+    // and that any changes we have made affect `minicore`(since it would get rebuilt).
+    mini_tests(test_env, args)?;
+    // Then, we copy the env vars from `test_env`
+    // We don't want to pass things like `RUSTFLAGS`, since they contain the -Zcodegen-backend flag.
+    // That would force `cg_gcc` to *rebuild itself* and only then run tests, which is undesirable.
+    let mut env = test_env.clone();
+    env.remove("RUSTFLAGS");
+    // FIXME: maybe we could allow for only a select test to be run?
```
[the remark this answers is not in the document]
FractalFir: Args can now be forwarded by using the `--` symbol.
Eg.
```sh
./y.sh test --cargo-tests --use-system-gcc -- all flags after this get passed to cargo test
```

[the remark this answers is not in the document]
antoyo: Nice.
I believe only this is missing for this PR:

> Could you please update the doc and the CI to use this new command instead of cargo test?

build_system/src/test.rs
```diff
@@ -203,6 +206,21 @@ fn clean(_env: &Env, args: &TestArg) -> Result<(), String> {
     create_dir(&path)
 }
 
+fn cargo_tests(test_env: &Env, test_args: &TestArg) -> Result<(), String> {
```
antoyo, line 209: The following is not necessary before merging this PR, but would be a nice to have:
currently, this seems to always recompile `rustc_codegen_gcc` and the `tests/lang_tests_*.rs` stuff, even when there are no changes.
Do you know what causes this?
Would that be easy to fix?
Maybe that's because we have some environment variables set that affect the build, which makes it different than the normal build? (It could be worth just extract `LD_LIBRARY_PATH` and `LIBRARY_PATH` from `test_env` and only sending those to `run_command_with_output_and_env` to see if this is the case and fixes this issue).

FractalFir: Sadly, that is *unlikely* to fix this, at least from my (rather amateur) analysis of the cargo logs. 

(One of) the rebuild reasons is this:
```
fingerprint dirty for rustc_codegen_gcc v0.1.0 
cargo::core::compiler::fingerprint:     dirty: ProfileConfigurationChanged
```

Now, if it was a flag issue, I'd expect this to be the recompilation reason:

```rust
   if self.rustflags != old.rustflags {
            return DirtyReason::RustflagsChanged {
                old: old.rustflags.clone(),
                new: self.rustflags.clone(),
            };
        }
```

I guess I will have to ask about what `ProfileConfigurationChanged` *means* - it is compared as a 64 bit hash in the cargo code, I will have to see what it is a hash of.

FractalFir: BTW, all the other rebuild reasons in the logs only tell me that the *tests* are getting rebuilt, because `cg_gcc` got rebuilt.
```
 target="lang_tests_common"}: cargo::core::compiler::fingerprint:     dirty: FsStatusOutdated(StaleDependency { name: "rustc_codegen_gcc",
```

antoyo: I tested my idea and it does seem to help.
With it, running `--cargo-tests` a second time gives:
```
[BUILD] build system
    Finished `release` profile [optimized] target(s) in 0.01s
GCC path retrieved from `config.toml`. Using `/opt/gcc/lib` as path for libgccjit
    Finished `dev` profile [optimized + debuginfo] target(s) in 0.01s
[…]
    Finished `test` profile [optimized + debuginfo] target(s) in 0.03s
     Running unittests src/lib.rs (target/debug/deps/rustc_codegen_gcc-e4ba054637ab4819)
```
while without it:
```
[BUILD] build system
    Finished `release` profile [optimized] target(s) in 0.03s
GCC path retrieved from `config.toml`. Using `/opt/gcc/lib` as path for libgccjit
   Compiling rustc_codegen_gcc v0.1.0 (/home/bouanto/Ordinateur/Programmation/Rust/Projets/rustc_codegen_gcc)
    Finished `dev` profile [optimized + debuginfo] target(s) in 1.37s
[…]
   Compiling rustc_codegen_gcc v0.1.0 (/home/bouanto/Ordinateur/Programmation/Rust/Projets/rustc_codegen_gcc)
    Finished `test` profile [optimized + debuginfo] target(s) in 8.51s
     Running unittests src/lib.rs (target/debug/deps/rustc_codegen_gcc-4d02d698ebbfa187)
```

build_system/src/test.rs
```diff
@@ -42,7 +42,7 @@ fn get_runners() -> Runners {
     );
     runners.insert("--extended-regex-tests", ("Run extended regex tests", extended_regex_tests));
     runners.insert("--mini-tests", ("Run mini tests", mini_tests));
-
+    runners.insert("--cargo-tests", ("Run cargo tests", cargo_tests));
```
antoyo, line 45: As a reminder:
Could you please update the doc and the CI to use this new command instead of cargo test?

antoyo (commented):
This looks good!

build_system/src/test.rs
```diff
@@ -1218,6 +1247,7 @@ fn run_all(env: &Env, args: &TestArg) -> Result<(), String> {
     test_libcore(env, args)?;
     extended_sysroot_tests(env, args)?;
     test_rustc(env, args)?;
+    cargo_tests(env, args)?;
```
antoyo, line 1250: Is this ran? I'm asking because `test_rustc` might return a non-zero exit code, so perhaps the `?` will return from this function.

FractalFir:
Should be good to merge now.

build_system/src/test.rs
```diff
@@ -203,6 +206,32 @@ fn clean(_env: &Env, args: &TestArg) -> Result<(), String> {
...
+    env.insert(
+        "CG_RUSTFLAGS".into(),
+        test_env.get("CG_RUSTFLAGS").map(|s| s.as_str()).unwrap_or("").to_string(),
+    );
+    // Pass all the default args + the user-specified ones.
+    let mut args: Vec<&dyn AsRef<OsStr>> = vec![&"cargo", &"test"];
+    args.extend(test_args.test_args.iter().map(|s| s as &dyn AsRef<OsStr>));
+    run_command_with_output_and_env(&args, None, Some(&env))?;
+    Ok(())
+}
```
antoyo, line 234:
```suggestion
}

```

antoyo (commented):
A small nitpick:

antoyo (approved):

antoyo:
Thanks a lot!
This is a very nice improvement!

PR #672, opened 2025-05-10
Fix UB in 128 bit cttz/ctlz intrinsics

FractalFir:
Fixes https://github.com/rust-lang/rustc_codegen_gcc/issues/604.

This PR changes a few things. 

First of all, it enables UB checks for tests in debug mode: This should allow us to catch any regressions.

Second, I extracted the common implementation of the "safe" `ctlz`/`cttz` intrinsics into a separate function. 

This makes handling the 128 bit edge case a bit easier, and allows us to use those intrinsics.

The third change modifies the 128 bit  `ctlz`/`cttz` emulation to use the "safe" version of the 64 bit intrinsic, instead of calling the GCC built-in directly.

Effectively, this turns a piece of IR like this:

```rust
// Get the ctlz for the low and high half of the 128 bit integer
let ctlz_low = ctlz_nonzero(low);
let ctlz_high = ctlz_nonzero(high);
```

Into a piece of code like this:

```rust
// Get the ctlz for the low and high half of the 128 bit integer
let ctlz_low = if low != 0 {ctlz_nonzero(low)} else {64};
let ctlz_high = if high != 0 {ctlz_nonzero(high)} else {64};
```

This solution is not the prettiest(maybe a few less branches could be used?), but it is UB-free, and relatively easy to implement.

### DIFF
--- a/src/builder.rs
+++ b/src/builder.rs
@@ -444,6 +444,55 @@ impl<'a, 'gcc, 'tcx> Builder<'a, 'gcc, 'tcx> {
         );
         result.to_rvalue()
     }
+    /// Performs a *bitcount* intrinsic on a given arg, properly handling the zero case.  
+    pub fn checked_bitcount_intrinsic(
+        &mut self,
+        name: rustc_span::Symbol,
+        arg: RValue<'gcc>,
+        width: u64,
+    ) -> RValue<'gcc> {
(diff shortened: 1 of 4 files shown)


antoyo:
Thanks for sending this PR.

How does the generated code compare to the fix made in [this PR](https://github.com/rust-lang/rustc_codegen_gcc/pull/635)?

antoyo:
Done in #635.

PR #680, opened 2025-05-21
Changes to constant handling - faster deduplication, more compact represtntation

FractalFir:
This PR changes a few things.

First of all, it caches ConstAllocations, ensuring that for a given ConstAllocation(or one an identical one with a different aligement), you will get the exact same Rvalue.

This allows us to simplify the deduplication code in `static_addr_of`, and remove the formating (which caused memory leaks). 

Additonally, this allows constant byte sequences to *sometimes* be represented in a more compact fashion. From a few checks, it looks like ~80 % of constant byte sequences can be optimized this way.

### DIFF
--- a/src/common.rs
+++ b/src/common.rs
@@ -9,7 +9,6 @@ use rustc_middle::mir::Mutability;
 use rustc_middle::mir::interpret::{ConstAllocation, GlobalAlloc, Scalar};
 use rustc_middle::ty::layout::LayoutOf;
 
-use crate::consts::const_alloc_to_gcc;
 use crate::context::CodegenCx;
 use crate::type_of::LayoutGccExt;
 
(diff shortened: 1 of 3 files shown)


FractalFir:
As a nice bonus, this PR seems to drastically reduce the ammount of memory used to build `regex-automata`.

I can now build `rustc` with all threads on, and run on ~28 GB of RAM. This is still suboptimal, but a fair bit better than before.

antoyo:
> As a nice bonus, this PR seems to drastically reduce the ammount of memory used to build `regex-automata`.
> 
> I can now build `rustc` with 10 threads, and run on ~28 GB of RAM. This is still suboptimal, but a fair bit better than before.

Very nice improvement!
Is this with `-Copt-level=0` and do you still disable some GCC passes?

FractalFir:
Should be good to merge now.

src/common.rs
```diff
@@ -46,12 +45,58 @@ impl<'gcc, 'tcx> CodegenCx<'gcc, 'tcx> {
 }
 
 pub fn bytes_in_context<'gcc, 'tcx>(cx: &CodegenCx<'gcc, 'tcx>, bytes: &[u8]) -> RValue<'gcc> {
-    let context = &cx.context;
-    let byte_type = context.new_type::<u8>();
-    let typ = context.new_array_type(None, byte_type, bytes.len() as u64);
-    let elements: Vec<_> =
-        bytes.iter().map(|&byte| context.new_rvalue_from_int(byte_type, byte as i32)).collect();
-    context.new_array_constructor(None, typ, &elements)
+    // Instead of alaways using an array of bytes, use an array of larger intgers of target endianess
```
antoyo, line 48:
```suggestion
    // Instead of always using an array of bytes, use an array of larger integers of target endianness
```

antoyo, line 48: Please add a `TODO` (or create an issue) to remind us about the other solution involving `global_set_initializer`.

src/common.rs
```diff
@@ -46,12 +45,58 @@ impl<'gcc, 'tcx> CodegenCx<'gcc, 'tcx> {
...
+    // if possible.
```
antoyo, line 49: Please mention that we do this to improve the memory usage.

antoyo (commented):
After this, it should be good to merge.
Thanks!

src/common.rs
```diff
@@ -46,12 +45,65 @@ impl<'gcc, 'tcx> CodegenCx<'gcc, 'tcx> {
...
 
 pub fn bytes_in_context<'gcc, 'tcx>(cx: &CodegenCx<'gcc, 'tcx>, bytes: &[u8]) -> RValue<'gcc> {
-    let context = &cx.context;
-    let byte_type = context.new_type::<u8>();
-    let typ = context.new_array_type(None, byte_type, bytes.len() as u64);
-    let elements: Vec<_> =
-        bytes.iter().map(|&byte| context.new_rvalue_from_int(byte_type, byte as i32)).collect();
-    context.new_array_constructor(None, typ, &elements)
+    // Instead of always using an array of bytes, use an array of larger integers of target endianness
+    // if possible. This reduces the ammount of `rvalues` we use, which reduces memory usage significantly.
```
antoyo, line 49:
```suggestion
    // if possible. This reduces the amount of `rvalues` we use, which reduces memory usage significantly.
```

src/common.rs
```diff
@@ -46,12 +45,65 @@ impl<'gcc, 'tcx> CodegenCx<'gcc, 'tcx> {
...
+    //
+    // FIXME(FractalFir): Consider using `global_set_initializer` instead. Before this is done, we need to confoirm that
```
antoyo, line 51:
```suggestion
    // FIXME(FractalFir): Consider using `global_set_initializer` instead. Before this is done, we need to confirm that
```

src/common.rs
```diff
@@ -46,12 +45,65 @@ impl<'gcc, 'tcx> CodegenCx<'gcc, 'tcx> {
...
+            let context = &cx.context;
+            let byte_type = context.new_type::<u64>();
+            let typ = context.new_array_type(None, byte_type, bytes.len() as u64 / 8);
+            let elements: Vec<_> = bytes
+                .chunks_exact(8)
+                .map(|arr| {
+                    let arr: [u8; 8] = arr.try_into().unwrap();
+                    context.new_rvalue_from_long(
+                        byte_type,
+                        // Since we are representin g arbitrary byte runs as intigers, we need to follow the target
```
antoyo, line 66:
```suggestion
                        // Since we are representing arbitrary byte runs as integers, we need to follow the target
```

src/common.rs
```diff
@@ -46,12 +45,65 @@ impl<'gcc, 'tcx> CodegenCx<'gcc, 'tcx> {
...
+                        // endianess.
```
antoyo, line 67:
```suggestion
                        // endianness.
```

antoyo (commented):
A few nitpicks:

antoyo (commented):
After this last nitpick is fixed, I believe I can merge.
Thanks!

bjorn3:
In your commit message: *representation

FractalFir:
> In your commit message: *representation

Did not see that, thanks :)!

antoyo:
Thanks a lot for this nice improvement!

Issue #683, opened 2025-05-22
Segfault caused by `packed`


FractalFir:
This piece of code:
```rust
#[derive(Clone, Copy, PartialEq)]
#[repr(packed(1))]
pub struct ScalarInt {
    data: u128,
    size: u8,
}
#[inline(never)]
fn eq(a: &ScalarInt, b: &ScalarInt) -> bool {
    a == b
}
fn main() {
    let data = [std::hint::black_box(ScalarInt { data: 0, size: 1 }); 2];
    data.iter().all(|v| eq(v, &ScalarInt { data: 0, size: 1 }));
}
```
Crashes under GCC when optimizations are enabled.

PR #684, opened 2025-05-23
Fix to 128 bit int unaligned loads

FractalFir:
This PR fixes the segfault caused by packed structs containing u128 / i128.  https://github.com/rust-lang/rustc_codegen_gcc/issues/683

I also added a test to detect any regressions like this in the future.

### DIFF
--- a/src/builder.rs
+++ b/src/builder.rs
@@ -924,7 +924,12 @@ impl<'a, 'gcc, 'tcx> BuilderMethods<'a, 'tcx> for Builder<'a, 'gcc, 'tcx> {
         // dereference after a drop, for instance.
         // FIXME(antoyo): this check that we don't call get_aligned() a second time on a type.
         // Ideally, we shouldn't need to do this check.
-        let aligned_type = if pointee_ty == self.cx.u128_type || pointee_ty == self.cx.i128_type {
+        // FractalFir: the `align == self.int128_align` check ensures we *do* call `get_aligned` if
+        // the alignment of a `u128`/`i128` is not the one mandated by the ABI. This ensures we handle
+        // under-aligned loads correctly.
+        let aligned_type = if (pointee_ty == self.cx.u128_type || pointee_ty == self.cx.i128_type)
+            && align == self.int128_align
+        {
             pointee_ty
         } else {
             pointee_ty.get_aligned(align.bytes())
--- a/src/context.rs
+++ b/src/context.rs
@@ -4,7 +4,7 @@ use std::collections::HashMap;
 use gccjit::{
     Block, CType, Context, Function, FunctionPtrType, FunctionType, LValue, Location, RValue, Type,
 };
-use rustc_abi::{HasDataLayout, PointeeInfo, Size, TargetDataLayout, VariantIdx};
+use rustc_abi::{Align, HasDataLayout, PointeeInfo, Size, TargetDataLayout, VariantIdx};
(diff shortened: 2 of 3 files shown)


src/builder.rs
```diff
@@ -924,7 +924,9 @@ impl<'a, 'gcc, 'tcx> BuilderMethods<'a, 'tcx> for Builder<'a, 'gcc, 'tcx> {
         // dereference after a drop, for instance.
         // FIXME(antoyo): this check that we don't call get_aligned() a second time on a type.
         // Ideally, we shouldn't need to do this check.
-        let aligned_type = if pointee_ty == self.cx.u128_type || pointee_ty == self.cx.i128_type {
+        let aligned_type = if (pointee_ty == self.cx.u128_type || pointee_ty == self.cx.i128_type)
```
antoyo, line 927: Could you please add a comment to explain why you do this?

src/builder.rs
```diff
@@ -924,7 +924,9 @@ impl<'a, 'gcc, 'tcx> BuilderMethods<'a, 'tcx> for Builder<'a, 'gcc, 'tcx> {
...
+            && align == Align::from_bytes(16).unwrap()
```
antoyo, line 928: Instead of hard-coding the alignment, could you query it from the Rust compiler like we do [here](https://github.com/rust-lang/rustc_codegen_gcc/blob/9aec231fbaf09ecd9ee6350100045cfccd4f4689/src/context.rs#L156)?
Seems it is possibly executed many times, it might be better to store this result in `CodegenCx`.

antoyo (commented):
Nice find!

src/context.rs
```diff
@@ -135,6 +135,9 @@ pub struct CodegenCx<'gcc, 'tcx> {
 
     #[cfg(feature = "master")]
     pub cleanup_blocks: RefCell<FxHashSet<Block<'gcc>>>,
+    /// The aligement of a u128/i128 type.
```
antoyo, line 138:
```suggestion
    /// The alignment of a u128/i128 type.
```

src/context.rs
```diff
@@ -135,6 +135,9 @@ pub struct CodegenCx<'gcc, 'tcx> {
...
+    // We cache this, since it is needed for aligement checks during loads.
```
antoyo, line 139:
```suggestion
    // We cache this, since it is needed for alignment checks during loads.
```

src/context.rs
```diff
@@ -226,6 +229,14 @@ impl<'gcc, 'tcx> CodegenCx<'gcc, 'tcx> {
         }
 
         let mut cx = Self {
+            int128_align: Align::from_bytes(
+                tcx.layout_of(ty::TypingEnv::fully_monomorphized().as_query_input(tcx.types.i128))
+                    .unwrap()
```
antoyo, line 234: Could you please avoid `unwrap()` in general and use `expect()` to provide a description?
Something like:
```suggestion
                    .expect("could not get the layout of i128")
```

src/context.rs
```diff
@@ -226,6 +229,14 @@ impl<'gcc, 'tcx> CodegenCx<'gcc, 'tcx> {
...
+                    .align
+                    .abi
+                    .bytes(),
+            )
+            .unwrap(),
```
antoyo, line 239: Same here.

FractalFir:
Should be OK to merge now.

antoyo (approved):
Good job!

PR #688, opened 2025-05-28
Add support for easily fuzzing `cg_gcc` with rustlantis

FractalFir:
This PR adds a new `y` command: `./y.sh fuzz`.

This command clones the [rustlantis](https://github.com/cbeuw/rustlantis) repo, and uses it to generate Rust programs to fuzz `cg_gcc` with. 

It takes in 3 optional arguments: search space start(`--start`), search space length(`--count`), and the parallelism(`--jobs` / `-j`).

By default, the fuzzer will search the first 100 programs(this takes about a minute, and functions as a small test for more obscure semantics of rust, like NaN handling). 

The script compiles programs with LLVM debug mode and GCC release mode, and then compares the result of their execution. If a compiler error occurs, or there is a runtime difference between the two backends, a message will be displayed, and the problematic source file will be copied over to either `target/fuzz/compiletime_error` or `target/fuzz/runtime_error`.

The fuzzer script respects `CG_RUSTFLAGS`, which should allow us to test things like software floating point emulation, or 128 bit intiger emulation. We can also enable UB checks.

# Limitations

`rustlantis` does not support 32 bit systems *without patches* , so we can't fuzz on 32 bit platforms. 

Additonally, on platforms not supported by LLVM, we have nothing to compare to. We *could* compare against debug GCC runs, but that may miss things.

We might consider adding  support for cross compilation / emulation, and compare scripts run on x86_64 with scripts run elsewhere.

### DIFF
--- a/build_system/src/fuzz.rs
+++ b/build_system/src/fuzz.rs
@@ -1,0 +1,238 @@
+use std::ffi::OsStr;
+use std::path::Path;
+
+use crate::utils::run_command_with_output;
+
+fn show_usage() {
+    println!(
+        r#"
+`fuzz` command help:
+    --help                 : Show this help"#
+    );
+}
+
+pub fn run() -> Result<(), String> {
+    // We skip binary name and the `fuzz` command.
+    let mut args = std::env::args().skip(2);
+    let mut start = 0;
+    let mut count = 100;
+    let mut threads =
+        std::thread::available_parallelism().map(|threads| threads.get()).unwrap_or(1);
+    while let Some(arg) = args.next() {
+        match arg.as_str() {
+            "--help" => {
+                show_usage();
+                return Ok(());
+            }
+            "--start" => {
+                start =
+                    str::parse(&args.next().ok_or_else(|| "Fuzz start not provided!".to_string())?)
+                        .map_err(|err| (format!("Fuzz start not a number {err:?}!")))?;
+            }
+            "--count" => {
+                count =
+                    str::parse(&args.next().ok_or_else(|| "Fuzz count not provided!".to_string())?)
+                        .map_err(|err| (format!("Fuzz count not a number {err:?}!")))?;
+            }
+            "-j" | "--jobs" => {
+                threads = str::parse(
+                    &args.next().ok_or_else(|| "Fuzz thread count not provided!".to_string())?,
+                )
+                .map_err(|err| (format!("Fuzz thread count not a number {err:?}!")))?;
+            }
+            _ => return Err(format!("Unknown option {}", arg)),
+        }
+    }
+
+    // Ensure that we have a cloned version of rustlantis on hand.
+    crate::utils::git_clone(
+        "https://github.com/cbeuw/rustlantis.git",
+        Some("clones/rustlantis".as_ref()),
+        true,
+    )
+    .map_err(|err| (format!("Git clone failed with message: {err:?}!")))?;
+
+    // Ensure that we are on the newest rustlantis commit.
+    let cmd: &[&dyn AsRef<OsStr>] = &[&"git", &"pull", &"origin"];
+    run_command_with_output(cmd, Some(&Path::new("clones/rustlantis")))?;
+
+    // Build the release version of rustlantis
+    let cmd: &[&dyn AsRef<OsStr>] = &[&"cargo", &"build", &"--release"];
+    run_command_with_output(cmd, Some(&Path::new("clones/rustlantis")))?;
+    // Fuzz a given range
+    fuzz_range(start, start + count, threads);
+    Ok(())
+}
+
+/// Fuzzes a range `start..end` with `threads`.
+fn fuzz_range(start: u64, end: u64, threads: usize) {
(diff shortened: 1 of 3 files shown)


build_system/src/fuzz.rs
```diff
@@ -0,0 +1,185 @@
...
+    // Ensure that we are on the newest rustlantis commit.
+    let cmd: &[&dyn AsRef<OsStr>] = &[&"git", &"pull", &"origin"];
+    run_command_with_output(cmd, Some(&Path::new("rustlantis")))?;
+    // Build the release version of rustlantis
+    let cmd: &[&dyn AsRef<OsStr>] = &[&"cargo", &"build", &"--release"];
+    run_command_with_output(cmd, Some(&Path::new("rustlantis")))?;
+    fuzz_range(start, start + count, threads);
+    Ok(())
+}
+fn fuzz_range(start: u64, end: u64, threads: usize) {
```
antoyo, line 61:
```suggestion

fn fuzz_range(start: u64, end: u64, threads: usize) {
```

build_system/src/fuzz.rs
```diff
@@ -0,0 +1,185 @@
...
+    use std::sync::atomic::{AtomicU64, Ordering};
+    use std::time::{Duration, Instant};
+
+    let total = end - start;
+    let start = Arc::new(AtomicU64::new(start));
+    let start_time = Instant::now();
+    for _ in 0..threads {
+        let start = start.clone();
+        std::thread::spawn(move || {
+            while start.load(Ordering::Relaxed) < end {
```
antoyo, line 72: I guess it is not very important, but if you're not sure which memory ordering to use, please use `Ordering::SeqCst` here and elsewhere.

build_system/src/fuzz.rs
```diff
@@ -0,0 +1,185 @@
...
+        let fuzzed = total - remaining;
+        let iter_per_sec = fuzzed as f64 / start_time.elapsed().as_secs_f64();
+        println!(
+            "fuzzed {fuzzed} cases({}%), at rate {iter_per_sec} iter/s, remaining ~{}s",
+            (100 * fuzzed) as f64 / total as f64,
+            (remaining as f64) / iter_per_sec
+        )
+    }
+}
+/// Builds & runs a file with LLVM.
```
antoyo, line 109:
```suggestion

/// Builds & runs a file with LLVM.
```

build_system/src/fuzz.rs
```diff
@@ -0,0 +1,185 @@
...
+        std::process::Command::new(&exe_path).output().map_err(|err| format!("{err:?}"))?;
+    if !output.status.success() {
+        return Err(format!(
+            "The program at {path:?}, compiled with GCC, exited unsuccessfully:{output:?}"
+        ));
+    }
+    std::fs::remove_file(exe_path).map_err(|err| format!("{err:?}"))?;
+    Ok(output.stdout)
+}
+/// Generates a new rustlantis file, & compares the result of running it with GCC and LLVM.
```
antoyo, line 156:
```suggestion

/// Generates a new rustlantis file, & compares the result of running it with GCC and LLVM.
```

build_system/src/fuzz.rs
```diff
@@ -0,0 +1,185 @@
...
+    let llvm_res = debug_llvm(&source_file)?;
+    let gcc_res = release_gcc(&source_file)?;
+    if llvm_res != gcc_res {
+        Ok(Err(source_file))
+    } else {
+        std::fs::remove_file(source_file).map_err(|err| format!("{err:?}"))?;
+        Ok(Ok(()))
+    }
+}
+/// Generates a new rustlantis file for us to run tests on.
```
antoyo, line 168:
```suggestion

/// Generates a new rustlantis file for us to run tests on.
```

antoyo (commented):
Thanks for your work!

Here's a first review:

build_system/src/fuzz.rs
FractalFir: I use Relaxed on purpose here. 

Since this function only deals with one atomic at a time, the ordering  of that atomic in relation to other atomic does not matter.

So, I use Relaxed here.

build_system/src/fuzz.rs
```diff
@@ -0,0 +1,238 @@
...
+        }
+    }
+
+    // Ensure that we have a cloned version of rustlantis on hand.
+    crate::utils::git_clone(
+        "https://github.com/cbeuw/rustlantis.git",
+        Some("clones/rustlantis".as_ref()),
+        true,
+    )
+    .map_err(|err| (format!("Fuzz thread count not a number {err:?}!")))?;
```
antoyo, line 53: This message seems to be copy/pasted from above. Please fix it.

antoyo (commented):
One nitpick.
After that, I can merge it.
Thanks for the good work!
This command is going to be very useful.

FractalFir:
Should be all ready for merge now.

antoyo (approved):

antoyo:
Thanks a lot for the amazing work!

PR #692, opened 2025-06-02
Added some bare-bones support for explict registers in ARM inline assembly

FractalFir:
A tiny patch adding the very basic support for explicit registers in ARM/AArch64 assembly.

### DIFF
--- a/src/asm.rs
+++ b/src/asm.rs
@@ -641,7 +641,8 @@ fn explicit_reg_to_gcc(reg: InlineAsmReg) -> &'static str {
                 },
             }
         }
-
+        InlineAsmReg::Arm(reg) => reg.name(),
+        InlineAsmReg::AArch64(reg) => reg.name(),
         _ => unimplemented!(),
     }
 }


src/asm.rs
```diff
@@ -641,7 +641,8 @@ fn explicit_reg_to_gcc(reg: InlineAsmReg) -> &'static str {
                 },
             }
         }
-
+        InlineAsmReg::Arm(reg) => reg.name(),
+        InlineAsmReg::AArch64(reg) => reg.name(),
         _ => unimplemented!(),
```
antoyo, line 646: The [LLVM codegen](https://github.com/antoyo/rust/blob/bf64d66bd58719fac2585eae5e546e5e1d9947f5/compiler/rustc_codegen_llvm/src/asm.rs#L603-L624) requires some special handling for `InlineAsmReg::Arm(ArmInlineAsmReg::r14)` and `InlineAsmReg::AArch64(AArch64InlineAsmReg::x30)`.
Do you know if this is also the case for GCC?

PR #697, opened 2025-06-06
Add `--toolchain` option for `test` command

GuillaumeGomez:


antoyo:
This does not appear to work.
When running:
```
./y.sh test --toolchain cg_gcc-stage1
```
I get a bunch of errors like:
```
error[E0599]: no method named `add_used_global` found for reference `&CodegenCx<'gcc, 'tcx>` in the current scope
   --> src/consts.rs:167:14
    |
167 |         self.add_used_global(global);
    |              ^^^^^^^^^^^^^^^
```
as though it would compile (the local) cg_gcc itself with the (remote) toolchain.

PR #698, opened 2025-06-06
Raise command signal to the current process

GuillaumeGomez:
Fixes https://github.com/rust-lang/rustc_codegen_gcc/issues/496.

### DIFF
--- a/build_system/src/rust_tools.rs
+++ b/build_system/src/rust_tools.rs
@@ -99,6 +99,10 @@ impl RustcTools {
 fn exec(input: &[&dyn AsRef<OsStr>], env: &HashMap<String, String>) -> Result<(), String> {
     #[cfg(unix)]
     {
+        // We use `exec` to call the `execvp` syscall instead of creating a new process where the
+        // command will be executed because very few signals can actually kill a current process,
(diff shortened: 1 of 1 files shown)


build_system/src/rust_tools.rs
```diff
@@ -97,6 +96,22 @@ impl RustcTools {
     }
 }
 
+fn exec(input: &[&dyn AsRef<OsStr>], env: &HashMap<String, String>) -> Result<(), String> {
+    #[cfg(unix)]
+    {
+        let error = crate::utils::get_command_inner(input, None, Some(env)).exec();
+        eprintln!("Command failed: {error:?}");
```
antoyo, line 103: I would rephrase the error to make this clear that it's the `execvp` syscall that failed.

antoyo:
Nice work!
This has the correct behavior, but just to make sure I understand, does this work because of the call to `exec()`?
If so, could you please document why we call `exec()` (i.e. to forward the signals)?

GuillaumeGomez:
Updated error message.

GuillaumeGomez:
Added docs too.

antoyo:
This is a very nice improvement!

PR #700, opened 2025-06-07
Add support for automatically reducing found fuzz cases.

FractalFir:
When reducing the issues found via fuzzing, I often find myself repeating certain cleanup steps, needed to much more quickly reduce a given file.

This PR automates those steps for all cases.

1. The **saved** examples use `println` and not hashing(when searching for bugs, hashing is still used). This is slower to run by a factor of 2, but it makes analyzing the bug samples much, much easier. 
2. Remove most calls to `dump_var`. This drastically speeds up the execution of the sample(both native and in MIRI). This more than makes up for the slow-down of using println. 
3. The `remove_dup_assign` step is not needed, but it exploits some properties of Rust to quickly reduce the sample by about 10-15%. 
4. `match_to_goto` uses the way Rustlantis generates MIR to very quickly turn almost all condtional branching in the sample into unconditional jumps. This makes analysing the generated MIR much easier, and allows for further reduction.
5. `block_abort` attempts to replace decoy blocks with calls to `abort`. In code generated by `rustlantis`, the decoy blocks may never execute, and are there only to confuse the compiler. So, replacing them with abort is always sound. This greatly reduces the generated sample, and makes reasoning about control flow simpler. 
6. `remove_block` tries to remove small blocks allotoghter. It runs after `block_abort`, because non-rechable blocks are still often needed(eg. they are the target of a jump).
7. `linearize_cf` tries to merge blocks, removing terminators. This, once again, makes reasoning about the control flow much easier. In some cases, it can reduce a function to a single, continuous block. 

Since those reductions exploit certain properties of Rust, more general tools like `creduce` are often incapable of applying them. So, this set of reduction steps should both speed up the production of minimized samples, and also automate some of the steps of the process. 

The resulting file still needs to be feed to a proper reducer for further processing, but, sinice it is considerably simpler already, the whole process will be quicker. 

Additionally, since I know that none of the transformations here can introduce UB, and that most of them don't alter the behaviour of the program, I can avoid running MIRI & rustc + LLVM when reducing samples, which further speeds up the process.


build_system/src/fuzz/reduce.rs
```diff
@@ -0,0 +1,307 @@
...
+        // Find the beginning of the next block to find the end of this block.
+        let Some(block_ends) = file[(block_starts + 1)..]
+            .iter()
+            .position(|line| line.starts_with("bb") && line.trim_end().ends_with(" = {"))
+        else {
+            // No more `block`s to remove - exit early.
+            break;
+        };
+        let block_ends = block_starts + block_ends + 1;
+        // Large blocks are likely to be neccsarry.
```
antoyo, line 218:
```suggestion
        // Large blocks are likely to be necessary.
```

build_system/src/fuzz/reduce.rs
```diff
@@ -0,0 +1,307 @@
...
+use std::path::{Path, PathBuf};
+fn save_reduction(lines: &[String], path: &PathBuf, ext: &str) {
+    let mut path = path.clone();
+    path.set_extension(&format!("rs.{ext}"));
+    let mut file = std::fs::File::create(&path).expect("Could not create the reduced example file");
+    for line in lines {
+        file.write_all(line.as_bytes()).expect("Could not save the reduced example");
+    }
+}
+/// Checks if a given reduction is valid.
```
antoyo, line 11:
```suggestion

/// Checks if a given reduction is valid.
```

build_system/src/fuzz/reduce.rs
```diff
@@ -0,0 +1,307 @@
...
+    if test_reduction(&file_copy, &path, cache) {
+        eprintln!("Reduced {path:?} by {} lines `remove_dup_assign`", reduction_count);
+        *file = file_copy;
+    } else {
+        remove_dup_assign(file, path, starts, (starts + ends) / 2, cache);
+        remove_dup_assign(file, path, (starts + ends) / 2, ends, cache);
+    }
+    save_reduction(file, path, "remove_dup_assign");
+}
+/// Removes all the unneeded calls to `dump_var`. This is not something tools like `cvise` can do,
```
antoyo, line 69:
```suggestion

/// Removes all the unneeded calls to `dump_var`. This is not something tools like `cvise` can do,
```

build_system/src/fuzz/reduce.rs
```diff
@@ -0,0 +1,307 @@
...
+            eprintln!("Reduced {path:?} by 3 lines `remove_dump_var`");
+            *file = file_copy;
+            curr = line;
+        } else {
+            curr = line + 1;
+        }
+    }
+    save_reduction(file, path, "remove_dump_var");
+}
+/// Replaces matches with gotos where possible.
```
antoyo, line 99:
```suggestion

/// Replaces matches with gotos where possible.
```

build_system/src/fuzz/reduce.rs
```diff
@@ -0,0 +1,307 @@
...
+            eprintln!("Reduced {path:?} by {} lines `match_to_goto`", match_ends - match_starts);
+            *file = file_copy;
+            curr = match_starts;
+        } else {
+            curr = match_ends;
+        }
+    }
+    save_reduction(file, path, "match_to_goto");
+}
+/// At this point, we can try "killing" blocks, by replacing their bodies with calls to `abort`.
```
antoyo, line 148:
```suggestion

/// At this point, we can try "killing" blocks, by replacing their bodies with calls to `abort`.
```

build_system/src/fuzz/reduce.rs
```diff
@@ -0,0 +1,307 @@
...
+            eprintln!("Reduced {path:?} by {} lines `block_abort`", block_ends - block_starts - 2);
+            *file = file_copy;
+            curr = block_starts;
+        } else {
+            curr = block_ends;
+        }
+    }
+    save_reduction(file, path, "block_abort");
+}
+/// Removes unreachable basic blocks
```
antoyo, line 195:
```suggestion

/// Removes unreachable basic blocks
```

build_system/src/fuzz/reduce.rs
```diff
@@ -0,0 +1,307 @@
...
+            eprintln!("Reduced {path:?} by {} lines `remove_blocks`", block_ends - block_starts);
+            *file = file_copy;
+            curr = block_starts;
+        } else {
+            curr = block_starts + 1;
+        }
+    }
+    save_reduction(file, path, "remove_block");
+}
+/// Merges blocks ending with unconditional jumps.
```
antoyo, line 235:
```suggestion

/// Merges blocks ending with unconditional jumps.
```

build_system/src/fuzz/reduce.rs
```diff
@@ -0,0 +1,307 @@
...
+            eprintln!("Reduced {path:?} by 3 lines `linearize_cf`");
+            *file = file_copy;
+            curr = block_starts;
+        } else {
+            curr = block_starts + 1;
+        }
+    }
+    save_reduction(file, path, "linearize_cf");
+}
+pub(super) fn reduce(path: impl AsRef<Path>) {
```
antoyo, line 281:
```suggestion

pub(super) fn reduce(path: impl AsRef<Path>) {
```

antoyo (commented):
A few nitpicks.
(And sorry about all the noise for new lines between functions: I checked and `rustfmt` doesn't seem to support adding a new line between functions. I'll try to find another solution to check this in the CI.)
Looks good!
Thanks for your work!

FractalFir:
Fixed the issues you pointed out, and improved the reduction by a tiny bit. 

Now, the reducer will try to remove function calls, and then functions. This can significantly reduce the example at a pretty fast pace.

FractalFir:
> I checked and `rustfmt` doesn't seem to support adding a new line between functions.

I suppose we could open an issue about that. `tidy` enforces this(I think), why shouldn't `rustfmt`?

antoyo:
> I suppose we could open an issue about that. `tidy` enforces this(I think), why shouldn't `rustfmt`?

There's already a bunch of issues about this. The most up-to-date seems to be [this one](https://github.com/rust-lang/rustfmt/issues/3382).
The fix apparently was made [a couple years ago](https://github.com/rust-lang/rustfmt/pull/4295), but was not merged to the main branch, for some reasons.

build_system/src/fuzz/reduce.rs
```diff
@@ -278,28 +306,126 @@ fn linearize_cf(file: &mut Vec<String>, path: &PathBuf) {
     }
     save_reduction(file, path, "linearize_cf");
 }
+
+/// Replaces a call to a given function with a 0 assigement to the destination place, and a Goto.
```
antoyo, line 310:
```suggestion
/// Replaces a call to a given function with a 0 assignment to the destination place, and a Goto.
```

build_system/src/fuzz/reduce.rs
```diff
@@ -278,28 +306,126 @@ fn linearize_cf(file: &mut Vec<String>, path: &PathBuf) {
...
+        // Find a function start
+        let Some(fn_start) = file[curr..].iter().position(|line| {
+            line.contains("#[custom_mir(dialect = \"runtime\", phase = \"initial\")]")
+        }) else {
+            // No more functions to remove - exit early.
+            break;
+        };
+        // Find the next function(and use that to find the end of this one).
+        // FIXME: this check is flawed: it will never remove the very last function(the one before main).
+        // The other checks will turn that function into a single call to abort, but it is still annoying it is kept.
```
antoyo, line 374: Is it missing a word at the end of that sentence around "annoying it is kept"?

antoyo (commented):
A few more nitpicks.
After that, it should be good to merged.
Thanks!

FractalFir:
Should be good to go now

antoyo (approved):

antoyo:

Issue #538, opened 2024-07-05
Re-enable run-make tests


antoyo:
They were disabled in the [last sync](https://github.com/rust-lang/rustc_codegen_gcc/pull/535) because it caused an assert failure in GCC.

PR #709, opened 2025-06-13
Reenable run-make tests

antoyo:
Fix rust-lang/rustc_codegen_gcc#538.


patches/tests/0001-Workaround-to-make-a-run-make-test-pass.patch
```diff
@@ -0,0 +1,25 @@
...
+diff --git a/tests/run-make/linker-warning/rmake.rs b/tests/run-make/linker-warning/rmake.rs
+index bc21739fefc..0946a7e2a48 100644
+--- a/tests/run-make/linker-warning/rmake.rs
++++ b/tests/run-make/linker-warning/rmake.rs
+@@ -55,7 +55,7 @@ fn main() {
+         diff()
+             .expected_file("short-error.txt")
+             .actual_text("(linker error)", out.stderr())
+-            .normalize(r#"/rustc[^/]*/"#, "/rustc/")
++            .normalize(r#"/tmp/rustc[^/]*/"#, "/tmp/rustc/")
```
antoyo, line 19: @bjorn3: Since we both need this patch, do you know if there would be any way to have this change upstream?

PR #710, opened 2025-06-14
Added support for testing the backend with abi-cafe

FractalFir:
I added a new command(abi-test) that runs the abi-cafe test suite, comparing the ABI implemented by the LLVM version of Rustc, and the GCC version of rustc.

Additionally, as a base sanity check, the command will also compare the implementation of the C abi in cg_gcc and in the system's C compiler.

### DIFF
--- a/build_system/src/abi_test.rs
+++ b/build_system/src/abi_test.rs
@@ -1,0 +1,62 @@
+use std::ffi::OsStr;
+use std::path::Path;
+
+use crate::utils::run_command_with_output;
+
+fn show_usage() {
+    println!(
+        r#"
+`abi-test` command help:
+    --help                 : Show this help"#
+    );
+}
+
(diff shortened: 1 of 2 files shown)


antoyo (approved):
Nice job!

bjorn3:
You could also add it to CI. Takes about 5min for cg_clif when only testing Linux. For handling the known test failures, there is a section in `src/report.rs` of abi-cafe that you can patch like so: https://github.com/rust-lang/rustc_codegen_cranelift/blob/master/patches/0002-abi-cafe-Disable-broken-tests.patch

FractalFir:
Getting this to run reliably in the CI is proving a bit difficult. Could we get this PR merged for now, and worry about adding new checks to the CI later?

antoyo:
I'm good with this.
Thanks for your contribution!

Issue #711, opened 2025-06-14
x86_64 libgccjit error when the C ABI passes an argument larger than 16 bytes


FractalFir:
This code causes libgccjit errors:
```c
#[derive(Copy, Clone,Default)]
#[repr(C)]
struct Many3 {
    f0: u64,
    f1: u64,
    f2: u8,
}
extern "C" fn struct_in_3(_: Many3){}
#[no_mangle]
fn do_test() {
    struct_in_3(Many3::default());
}
```

The `repr(C)` on the struct is not needed, but it sliences the imporper-ctype warning.

error:
```
libgccjit.so: error: : bitcast with types of different sizes
input expression (size: 64):
 <addr_expr 0x7efdb68e95c0
    type <pointer_type 0x7efdb68f29d8
        type <array_type 0x7efdb68f2930 type <integer_type 0x7efdb68ecc78 signed char>
            BLK
            size <integer_cst 0x7efdb6b60690 constant 192>
            unit-size <integer_cst 0x7efdb6b606c0 constant 24>
            user align:64 warn_if_not_align:0 symtab:0 alias-set -1 canonical-type 0x7efdb68f2888 domain <integer_type 0x7efdb68f2738>
            pointer_to_this <pointer_type 0x7efdb68f29d8>>
        unsigned DI
        size <integer_cst 0x7efdb6b60228 constant 64>
        unit-size <integer_cst 0x7efdb6b60240 constant 8>
        align:64 warn_if_not_align:0 symtab:0 alias-set -1 canonical-type 0x7efdb68f2a80>
    visited
    arg:0 <var_decl 0x7efdb68f8000 stack_var_1 type <array_type 0x7efdb68f2930>
        addressable visited BLK (null):0:0 size <integer_cst 0x7efdb6b60690 192> unit-size <integer_cst 0x7efdb6b606c0 24>
        align:64 warn_if_not_align:0 context <function_decl 0x7efdb68f4200 do_test>>>
requested type (size: 192):
 <record_type 0x7efdb68eca80  BLK
    size <integer_cst 0x7efdb6b60690 type <integer_type 0x7efdb6b4b690 bitsizetype> constant 192>
    unit-size <integer_cst 0x7efdb6b606c0 type <integer_type 0x7efdb6b4b5e8 sizetype> constant 24>
    user align:64 warn_if_not_align:0 symtab:0 alias-set -1 canonical-type 0x7efdb68eca80
[...]

PR #715, opened 2025-06-18
Added some sanity checks to function call compilation. This will catch *some* cases of ABI mismatch.

FractalFir:
This PR introduces 2 changes:

1. The function `known_eq` is a fallible form of equality. `known_eq` does not suffer from the issues the current implementation of `Eq` suffers from(sometimes, a `float` would be compared as uneqal to `float` simply because the pointer was different). `known_eq` reimplements some of that logic by itself, returning Some(true) if and only if types are known to be equal, and Some(false) if and only if it is known that the types can't be possibly equal. 

2. This is later used in a sanity check in `Builder::check_ptr_call`. Instead of checking if both types are / are not vectors, it will check if they are "known to be unequal". So, it will also painc if this code path is taken with types struct and pointer(ABI mismatch) but will still allow primitive types to pass trough. Since we only panic if the types are not vectors, and are known to be uneqal, this will effectively only catch ABI issues. 

This change allows us to catch ABI bugs like rust-lang/rustc_codegen_gcc#711 on the `cg_gcc` side, giving us a proper backtrace, and more detalied information about the bug(the types in play, the argument), and allowing us to use a debuger to see exactly what went wrong.


src/common.rs
```diff
@@ -524,4 +528,43 @@ impl<'gcc, 'tcx> TypeReflection<'gcc, 'tcx> for Type<'gcc> {
 
         false
     }
+    fn known_eq(&self, other: &Self, cx: &CodegenCx<'gcc, 'tcx>) -> Option<bool> {
```
antoyo, line 531: Would the use of `is_compatible_with` help reduce the size of this function?

src/common.rs
```diff
@@ -524,4 +528,58 @@ impl<'gcc, 'tcx> TypeReflection<'gcc, 'tcx> for Type<'gcc> {
...
+            }
+            return Some(
+                (self.get_size() == other.get_size())
+                    && (self.is_signed(cx) == other.is_signed(cx)),
+            );
+        } else if self.is_vector() {
+            if !other.is_vector() {
+                return Some(false);
+            }
+            return None;
```
antoyo, line 550: Do you think it would be worth it to check the sizes match?
We can probably do it with `get_element_type()` and `get_num_units()`?

src/common.rs
```diff
@@ -524,4 +528,58 @@ impl<'gcc, 'tcx> TypeReflection<'gcc, 'tcx> for Type<'gcc> {
...
+        } else if self.is_struct().is_some() {
+            if other.is_struct().is_none() {
+                return Some(false);
+            }
+            return None;
+        } else if let Some(s_ptr) = self.get_pointee() {
+            let Some(other_ptr) = other.get_pointee() else {
+                return Some(false);
+            };
+            return s_ptr.known_eq(&other_ptr, cx);
```
antoyo, line 560: I guess it doesn't matter much since there's a check for pointers in the assert in the other file, but is this the expected behavior to check if the pointee types are equal since we can bitcast pointers of different types?

antoyo (commented):
Some questions.

src/common.rs
FractalFir: The code path calling this funcition in the assert should not be hit by pointers at all - it is meant to bitcast vectors. 

The idea behind `known_eq` is to serve as a saner replacement for the current `Eq` implementation of `Type`. 

As we discussed before, comparing `Types` / `Rvalues` on the `libgccjit` side is quite hard to do, and would require a lot of code to be added. 

Here, I am seeing if I could possibly work around that issue, by doing some of that comparison logic on the Rust side.

Really, I am trying to see if I could implement the functions we require entirely on the Rust side, subverting the limitations of the GCC API. 

One thing I am considering attempting is field-vise comparison of structs(so that structs with identical elements compare as equal).

src/common.rs
FractalFir: Might be worth checking. Will add now.

src/common.rs
antoyo: I still believe that using that using `global_set_initializer()` would be better than this approach. Was there any reason it would not work?

Also, I would be very wary of a struct comparison done like that: we need to take some attributes into account like `packed` and possibly some others: I much prefer to let that job to the gcc backend even though I'm not sure right now how that could work since it would involve the target part of the compiler.

FractalFir: `global_set_initializer` required globals to have unique names, and I think that messed with the string-based comparisons. I also belive could not get it to ever run right - `global_set_initializer` always errored out when I used it. I belive that was caused by some kind of type issue - it always complained about the type of the global and the type of the initializer rvalue being different. Maybe this is related to us calling `get_aligned` needlessly. 

> Also, I would be very wary of a struct comparison done like that: we need to take some attributes into account like packed and possibly some others: I much prefer to let that job to the gcc backend even though I'm not sure right now how that could work since it would involve the target part of the compiler.

Which is why `known_eq` will not return `Some(true)` for structs(at least for now). For structs, the current version only allows us to check that structs are *unequal* which is fine for our purposes.

FractalFir: Also `global_set_initializer` will not help us with deduplication, at all. We will still have to manage the lvalues by ourselves, and deduplicate them ourselves.

PR #722, opened 2025-06-26
Remove unnecesary uses of the 'current_func' field, replacing it with  a call to function.

FractalFir:
All uses of the `current_func` field of `Context` in `Builder` can be safely replaced with `Builder::current_func`. This is done in preparation for removal of that field. 

Only one use of `current_func` remains - within `const_undef`, which is a function of `Context`, and not `Builder`.


antoyo (approved):
Nice improvement!
Thanks!

antoyo:
Thanks for your contribution!

PR #730, opened 2025-07-05
Add native CI for Aarch64

antoyo:
* [ ] Many UI test errors are caused by `__atomic_load_16` not being found.
 * [ ] libcore tests crash with SIGABRT (`store_illegal_rt_store_acq_rel_ordering` aborts whereas it should panic with `there is no such thing as an acquire-release store`)
 * [ ] Segfault in extended rand tests
 * [ ] Segfault in project tests (`log` crate: in the doc test src/lib.rs - (line 97), possibly at `__memcpy_sve () at ../sysdeps/aarch64/multiarch/memcpy_sve.S:75`, called by `proc_macro::bridge::selfless_reify::reify_to_extern_c_fn_hrt_bridge::wrapper`, seems like trying to read a null pointer in `x1`. This might be the [`sret` issue](https://rust-lang.zulipchat.com/#narrow/channel/421156-gsoc/topic/Project.3A.20Bootstrap.20of.20rustc.20with.20rustc_codegen_gcc/search/x1)).


tests/compile/simd-ffi.rs
```diff
@@ -1,5 +1,9 @@
 // Compiler:
 
+fn main() {
```
antoyo, line 3: Do not forget to re-enable this test.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustc_codegen_gcc/730/3aea6dce9935004372b634ff7cff896c26c73137..fc964cb976a5bc99a0cf5198a9d36cde0b9e9641)*

src/gcc_util.rs
```diff
@@ -141,6 +150,23 @@ pub fn target_cpu(sess: &Session) -> &str {
     }
 }
 
+/*pub fn add_baseline_flags<'gcc>(context: &Context<'gcc>, sess: &Session) {
```
antoyo, line 153: Either remove or re-enable this.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustc_codegen_gcc/730/3aea6dce9935004372b634ff7cff896c26c73137..fc964cb976a5bc99a0cf5198a9d36cde0b9e9641)*

PR #733, opened 2025-07-08
Inserted a local variable in volatile_load, to ensure reads don't move across blocks.

### LINKED ISSUE #717: Does `volatile_load` handle blocks correctly?
When looking at the implementation of `Builder::load`, I noticed it is quite a bit more complex than `volatile_load`.

`load` creates a temporary rvalue to ensure the read remains in the correct block:
```rust
   let deref = ptr.dereference(self.location).to_rvalue();
        let loaded_value = function.new_local(
            self.location,
            aligned_type,
            format!("loadedValue{}", self.next_value_counter()),
        );
        block.add_assignment(self.location, loaded_value, deref);
```
`volatile_load` does no such thing:
```
    fn volatile_load(&mut self, ty: Type<'gcc>, ptr: RValue<'gcc>) -> RValue<'gcc> {
        let ptr = self.context.new_cast(self.location, ptr, ty.make_volatile().make_pointer());
        ptr.dereference(self.location).to_rvalue()
    }
```
Is this correct?

FractalFir:
Fixes rust-lang/rustc_codegen_gcc#717

### DIFF
--- a/src/builder.rs
+++ b/src/builder.rs
@@ -980,7 +980,11 @@ impl<'a, 'gcc, 'tcx> BuilderMethods<'a, 'tcx> for Builder<'a, 'gcc, 'tcx> {
 
     fn volatile_load(&mut self, ty: Type<'gcc>, ptr: RValue<'gcc>) -> RValue<'gcc> {
         let ptr = self.context.new_cast(self.location, ptr, ty.make_volatile().make_pointer());
-        ptr.dereference(self.location).to_rvalue()
+        // (FractalFir): We insert a local here, to ensure this volatile load can't move across
+        // blocks.
+        let local = self.current_func().new_local(self.location, ty, "volatile_tmp");
+        self.block.add_assignment(self.location, local, ptr.dereference(self.location).to_rvalue());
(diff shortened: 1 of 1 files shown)


src/builder.rs
```diff
@@ -980,7 +980,9 @@ impl<'a, 'gcc, 'tcx> BuilderMethods<'a, 'tcx> for Builder<'a, 'gcc, 'tcx> {
 
     fn volatile_load(&mut self, ty: Type<'gcc>, ptr: RValue<'gcc>) -> RValue<'gcc> {
         let ptr = self.context.new_cast(self.location, ptr, ty.make_volatile().make_pointer());
-        ptr.dereference(self.location).to_rvalue()
+        let local = self.current_func().new_local(self.location, ty, "volatile_tmp");
```
antoyo, line 983: Please add a note like in `load()` to explain why we do this so that someone doesn't remove it at some point.

FractalFir:
Ready for review.

antoyo:
Thanks a lot for this PR!

PR #734, opened 2025-07-09
Skip needlessly setting the default visibility on functions

FractalFir:
The default visibility of a function is ` __attribute__((visibility(default)))`, so explicitly setting that attribute just wastes RAM and CPU cycles.


FractalFir:
Fixed some formatting issues.

antoyo (approved):

antoyo:
Nice find and thanks a lot for this PR!

PR #743, opened 2025-07-25
ci: Don't set `compiler-builtins-no-f16-f128`

tgross35:
Since rust-lang/rust be35d37d8b6c ("Use the compiler to determine whether or not to enable f16 and f128"), `compiler-builtins` relies on `rustc` to report whether or not `f16` and `f128` are supported, which is reported by the backend. This means that there should no longer be any need to set this config in CI.

Backend config: https://github.com/rust-lang/rustc_codegen_gcc/blob/f682d09eefc6700b9e5851ef193847959acf4fac/src/lib.rs#L499-L510

### DIFF
--- a/.github/workflows/m68k.yml
+++ b/.github/workflows/m68k.yml
@@ -82,14 +82,14 @@ jobs:
     - name: Build sample project with target defined as JSON spec
       run: |
         ./y.sh prepare --only-libcore --cross
-        ./y.sh build --sysroot --features compiler-builtins-no-f16-f128 --target-triple m68k-unknown-linux-gnu --target ${{ github.workspace }}/target_specs/m68k-unknown-linux-gnu.json
+        ./y.sh build --sysroot --target-triple m68k-unknown-linux-gnu --target ${{ github.workspace }}/target_specs/m68k-unknown-linux-gnu.json
         CG_RUSTFLAGS="-Clinker=m68k-unknown-linux-gnu-gcc" ./y.sh cargo build --manifest-path=./tests/hello-world/Cargo.toml --target ${{ github.workspace }}/target_specs/m68k-unknown-linux-gnu.json
         ./y.sh clean all
 
     - name: Build
       run: |
         ./y.sh prepare --only-libcore --cross
-        ./y.sh build --sysroot --features compiler-builtins-no-f16-f128 --target-triple m68k-unknown-linux-gnu
+        ./y.sh build --sysroot --target-triple m68k-unknown-linux-gnu
         ./y.sh test --mini-tests --target-triple m68k-unknown-linux-gnu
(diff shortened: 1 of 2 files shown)


antoyo:
@FractalFir: Is this the error you got on Dec Alpha or something: https://github.com/rust-lang/rustc_codegen_gcc/actions/runs/16532424105/job/46760590729?pr=743#step:11:64 ?

tgross35:
Yeah, that's the same error. Looks like GCC is reporting it supports f128 on m68k when it actually doesn't?

tgross35:
Actually, I think the toolchain here is from before https://github.com/rust-lang/rust/pull/143405  merged, meaning https://github.com/rust-lang/rust/commit/be35d37d8b6c206576ff3cdcad956d921b47cc6b isn't available. Should have known since that is what we were looking at in [#t-compiler/help > How to correctly disable &#96;f16&#96; and &#96;f128&#96; when bootstraping](https://rust-lang.zulipchat.com/#narrow/channel/182449-t-compiler.2Fhelp/topic/How.20to.20correctly.20disable.20.60f16.60.20and.20.60f128.60.20when.20bootstraping/with/530850893).

tgross35:
Looks like this is getting stuck because running `rustc` in the `build.rs` for Cargo.toml doesn't account for codegen backend, and we can't check it because build scripts don't have `RUSTFLAGS`. So the compiler-builtins configure part is unfortunately apparently broken for non-llvm backends, and sadly this needs some Cargo tie in to fix it.

Started discussion at [#t-cargo > Detecting &#96;-Zcodegen-backend&#96; from build.rs](https://rust-lang.zulipchat.com/#narrow/channel/246057-t-cargo/topic/Detecting.20.60-Zcodegen-backend.60.20from.20build.2Ers/with/532978463).

tgross35:
🎉 it works!

The first commit is https://github.com/rust-lang/compiler-builtins/commit/87a66ec9699e5ddf2c660277b8078099efd01311. It's in the compiler-builtins repo now but will take a while to sync to rust-lang/rust then rust-lang/rustc_codegen_gcc, so using a patch unblocks this.

antoyo:
Could you please explain what this unblocks?
What is the reason to merge this vs waiting for this to get merged upstream?

tgross35:
> Could you please explain what this unblocks?
>
> What is the reason to merge this vs waiting for this to get merged upstream?

This isn't technically a blocker now that the build here looks okay, but I plan to delete the `compiler-builtins-no-f16-f128` feature since it shouldn't be needed anymore. No strong need to merge this with the patch if you are okay waiting for the sync, the benefit is just that it should make it easier for you to test on targets that have broken f16/f128 (which I believe helps some of @FractalFir's work).

antoyo:
Ok, I'll let @FractalFir tell us if that would help him and if so, I'll merge this.

tgross35:
The commit is in r-l/r now, in case it's possible to do a sync soon

antoyo:
Done in #751.
Thanks!

Issue #590, opened 2025-01-11
Remove the environment variable EMBED_LTO_BITCODE


antoyo:
It should not be needed anymore.

Issue #541, opened 2024-07-05
LTO tests should not require -fat-lto-objects when only bitcode is requested


antoyo:
In [the last sync](https://github.com/rust-lang/rustc_codegen_gcc/pull/535), we needed a [workaround](https://github.com/rust-lang/rustc_codegen_gcc/blob/98ed962c7d3eebe12c97588e61245273d265e72f/src/back/write.rs#L60-L61) to keep the tests passing in the CI.

PR #750, opened 2025-08-13
Cleanup lto dead code

dvermd:
This should address #541 and #590. I hope it doesn't collide with work in progress on your side


antoyo:
This PR seems to remove the support for LTO altogether, which is not desired.

I have a PR [here](https://github.com/rust-lang/rustc_codegen_gcc/pull/678) that does fix those issue, but I'm not yet satisfied with the fact that it outputs some stuff to `stderr` when LTO is not supported.

Issue #777, opened 2025-10-11
OOM when compiling cargo


antoyo:
More specifically the dependency `p384`.

thesamesam:
Thank you for filing this! For completeness, this is from [#rustc-codegen-gcc > Wiring up gcc backend in Gentoo @ 💬](https://rust-lang.zulipchat.com/#narrow/channel/386786-rustc-codegen-gcc/topic/Wiring.20up.20gcc.20backend.20in.20Gentoo/near/544272928) where I had `rustc` use 64GB with this stacktrace:
```
Thread 2 (Thread 0x7f9549e2c6c0 (LWP 1310744) "opt cgu.0"):
#0  __memcpy_avx_unaligned () at ../sysdeps/x86_64/multiarch/memmove-vec-unaligned-erms.S:225
#1  0x00007f95691dba8c in copy_node(tree_node*) () from target:/var/tmp/portage/dev-lang/rust-1.91.0_beta20250928/work/rustc-beta-src/build/x86_64-unknown-linux-gnu/stage1/lib/libgccjit.
so.0
#2  0x00007f9568ef8eb6 in copy_tree_r(tree_node**, int*, void*) () from target:/var/tmp/portage/dev-lang/rust-1.91.0_beta20250928/work/rustc-beta-src/build/x86_64-unknown-linux-gnu/stage
1/lib/libgccjit.so.0
#3  0x00007f9568b28273 in mostly_copy_tree_r(tree_node**, int*, void*) () from target:/var/tmp/portage/dev-lang/rust-1.91.0_beta20250928/work/rustc-beta-src/build/x86_64-unknown-linux-gn
u/stage1/lib/libgccjit.so.0
#4  0x00007f95691e7476 in walk_tree_1(tree_node**, tree_node* (*)(tree_node**, int*, void*), void*, hash_set<tree_node*, false, default_hash_traits<tree_node*> >*, tree_node* (*)(tree_no
de**, int*, tree_node* (*)(tree_node**, int*, void*), void*, hash_set<tree_node*, false, default_hash_traits<tree_node*> >*)) [clone .localalias] () from target:/var/tmp/portage/dev-lang
/rust-1.91.0_beta20250928/work/rustc-beta-src/build/x86_64-unknown-linux-gnu/stage1/lib/libgccjit.so.0
[...]

PR #783, opened 2025-10-26
Fix RAM usage by splitting deeply nested expressions

antoyo:
Fix rust-lang/rustc_codegen_gcc#777

 * [x] try only setting `TREE_VISITED` on the `ADDR_EXPR` in `playback::function::get_address` and the `ADDR_EXPR` in `playback::lvalue::get_address`. (**edit:** doesn't work)

I was able to build `cargo` with 5-10 GB RAM while it used to take +64 GB.

This [function](https://github.com/mit-plv/fiat-crypto/blob/5d98fcc6da685ee06302ec43bdcd7c861539b41a/fiat-rust/src/p384_64.rs#L219) would generate deeply nested expressions which made GCC use huge amount of RAM.

### DIFF
--- a/src/builder.rs
+++ b/src/builder.rs
@@ -668,32 +668,38 @@ impl<'a, 'gcc, 'tcx> BuilderMethods<'a, 'tcx> for Builder<'a, 'gcc, 'tcx> {
     }
 
     fn add(&mut self, a: RValue<'gcc>, b: RValue<'gcc>) -> RValue<'gcc> {
-        self.gcc_add(a, b)
+        self.assign_to_var(self.gcc_add(a, b))
     }
 
     fn fadd(&mut self, a: RValue<'gcc>, b: RValue<'gcc>) -> RValue<'gcc> {
-        a + b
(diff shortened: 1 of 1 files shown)


src/builder.rs
```diff
@@ -2409,6 +2427,15 @@ impl<'a, 'gcc, 'tcx> Builder<'a, 'gcc, 'tcx> {
         let res = then_vals | else_vals;
         self.bitcast_if_needed(res, result_type)
     }
+
+    // GCC doesn't like deeply nested expressions.
+    // By assigning intermediate expressions to a variable, this allow us to avoid deeply nested
+    // expressions and GCC will use much less RAM.
+    fn assign_to_var(&self, value: RValue<'gcc>) -> RValue<'gcc> {
+        let var = self.current_func().new_local(self.location, value.get_type(), "addResult");
```
antoyo:
```suggestion
        let var = self.current_func().new_local(self.location, value.get_type(), "opResult");
```

PR #816, opened 2025-12-10
Fix #537: UI test removed logic update

### LINKED ISSUE #537: Better removal of UI tests with errors emitted by the frontend
The test `tests/ui/consts/const-eval/parse_ints.rs` in the [file `failing-ui-tests.txt`](https://github.com/rust-lang/rustc_codegen_gcc/blob/98ed962c7d3eebe12c97588e61245273d265e72f/tests/failing-ui-tests.txt#L93) even though this is not something to fix on our end.

The reason it is not removed is because the test file doesn't contain any error pattern like `//@ error-pattern:` and [we only remove tests containing those patterns](https://github.com/rust-lang/rustc_codegen_gcc/blob/98ed962c7d3eebe12c97588e61245273d265e72f/build_system/src/test.rs#L845-L847).
According to the [documentation](https://rustc-dev-guide.rust-lang.org/tests/ui.html#controlling-passfail-expectations), a UI test is expected to generate a compile error by default, so perhaps we should only keep the tests with one of the pass headers:

 * `//@ check-pass`
 * `//@ build-pass`
 * `//@ run-pass`

harin-ramesh:
This fixes https://github.com/rust-lang/rustc_codegen_gcc/issues/537

### DIFF
--- a/build_system/src/test.rs
+++ b/build_system/src/test.rs
@@ -834,8 +834,9 @@ fn valid_ui_error_pattern_test(file: &str) -> bool {
     .any(|to_ignore| file.ends_with(to_ignore))
 }
 
-fn contains_ui_error_patterns(file_path: &Path, keep_lto_tests: bool) -> Result<bool, String> {
-    // Tests generating errors.
+fn contains_ui_error_patterns(file_path: &Path, _keep_lto_tests: bool) -> Result<bool, String> {
+    // Inverted logic: UI tests are expected to fail by default unless marked with pass markers.
+    // Return false (keep) for tests with pass markers, true (remove) for everything else.
     let file = File::open(file_path)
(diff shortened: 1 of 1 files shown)


antoyo:
I completely forgot about this PR, sorry about that.
What's the status of this?
I remember we talked about this, but I forgot the details.

harin-ramesh:
I also don't remember exact details. With just the pass marker, not all test cases were passing because some tests include annotations like //@ error-pattern: or //@ known-bug, etc and pass marker

So I updated the existing function by adding an additional condition. Instead of returning false by default, the logic now ensures that a file is marked as a contain ui test error only if it does not contain any error-pattern (or similar annotations) and does not have a pass marker

PR #844, opened 2026-01-28
Sync from rust 2026/01/28

antoyo:


src/declare.rs
```diff
@@ -100,18 +99,14 @@ impl<'gcc, 'tcx> CodegenCx<'gcc, 'tcx> {
...
+        declare_raw_fn(
             self,
             name,
             callconv,
             return_type,
             &[self.type_i32(), const_string],
             variadic,
-        );
-        // NOTE: it is needed to set the current_func here as well, because get_fn() is not called
-        // for the main function.
```
antoyo: Check if this make some tests fail.

PR #845, opened 2026-01-29
Simplify running tests

GuillaumeGomez:

PR #841, opened 2026-01-26
Add two new target-specific intrinsics mapping

GuillaumeGomez:
Fixes https://github.com/rust-lang/rustc_codegen_gcc/issues/288.

PR #846, opened 2026-01-29
Split `compile` tests from `run` tests

GuillaumeGomez:
Follow-up of https://github.com/rust-lang/rustc_codegen_gcc/pull/845.

This PR introduces a second step to check that some tests only need to compile.

### DIFF
--- a/tests/lang_tests.rs
+++ b/tests/lang_tests.rs
@@ -11,7 +11,7 @@ fn compile_and_run_cmds(
     compiler_args: Vec<String>,
     test_target: &Option<String>,
     exe: &Path,
-    run_binary: bool,
+    test_mode: TestMode,
 ) -> Vec<(&'static str, Command)> {
     let mut compiler = Command::new("rustc");
     compiler.args(compiler_args);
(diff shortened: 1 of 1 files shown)


GuillaumeGomez:
Ah that one is interesting. It means that when I did #841, it didn't check for m68k somehow. There doesn't seem to be a direct translation for `__builtin_ia32_pmaxsd128` on m68k though. Gonna ignore this test for this target I think.

antoyo:
> Ah that one is interesting. It means that when I did #841, it didn't check for m68k somehow. There doesn't seem to be a direct translation for `__builtin_ia32_pmaxsd128` on m68k though. Gonna ignore this test for this target I think.

Are you sure it's not a rustc error? In any case, I believe it's OK to ignore it for m68k.

GuillaumeGomez:
CI finally passed! \o/

antoyo:
Rebasing on master should fix the CI.

GuillaumeGomez:
CI passed! \o/

GuillaumeGomez:
I removed one too much commit. ^^'

antoyo (approved):
Looks good.
One thing to maybe change.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustc_codegen_gcc/846/26406710580c8b573f59603852ddd5741fce2fbb..01cda7d674853dedf842dc161df02d0724aee5ca)*

antoyo (approved):

antoyo:
Thanks for your contribution.

PR #858, opened 2026-02-25
Add CI checks with LLVM sysroot

GuillaumeGomez:
Fixes https://github.com/rust-lang/rustc_codegen_gcc/issues/850.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -37,6 +37,10 @@ jobs:
           "--test-successful-rustc --nb-parts 2 --current-part 1",
           "--projects",
           "--gcc-asm-tests",
+          # With the LLVM sysroot now.
+          "--test-successful-rustc --nb-parts 2 --current-part 0 --use-llvm-sysroot",
+          "--test-successful-rustc --nb-parts 2 --current-part 1 --use-llvm-sysroot",
+          "--projects --use-llvm-sysroot",
         ]
 
     steps:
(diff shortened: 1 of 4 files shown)


GuillaumeGomez:
@Kobzol Any idea how to make this work with rust's bootstrap? :)

Kobzol:
So if I understand it correcty, you want to build rustc that supports both cg_llvm and cg_gcc. Then build the stdlib with cg_llvm, and then run the test suite in a way so that cg_gcc is used as the default codegen backend. I think that should work by setting `codegen-backends = ["llvm", "gcc"]` (in this order), and then run `x test --test-codegen-backend gcc`.

GuillaumeGomez:
@Kobzol Does `bootstrap` support using a custom path for `gcc` and `cg_gcc`?

antoyo:
Yes, with `gcc.libgccjit-libs-dir`.

**Edit:** This is only for `gcc`, though. For `cg_gcc`, I don't know.

Kobzol:
You can override the libgccjit used, but not `cg_gcc`, because that one is actually built by bootstrap itself. You'd have to implant it into the sysroot of the locally built compiler, but that's not really possible "mid-bootstrap-build", because bootstrap will be dynamically recreating the sysroot directory.

antoyo:
An idea could be to run the tests like we currently do, but without the `--sysroot` flag [from here](https://github.com/rust-lang/rustc_codegen_gcc/blob/1127416326321b489f6f1a3117bc7c4ab053c2bb/build_system/src/test.rs#L1118).

PR #861, opened 2026-03-07
Add TODO checker to cg_gcc

reddevilmidzy:
cc: https://github.com/rust-lang/rust/pull/153166
MCP: https://github.com/rust-lang/compiler-team/issues/963

After the commit https://github.com/rust-lang/rust/pull/153166/changes/69f9c6a020b3c8f21b2f3b346388d7463fdb798f is merged here, CI will pass.

### DIFF
--- a/build_system/src/todo.rs
+++ b/build_system/src/todo.rs
@@ -1,0 +1,72 @@
+use std::ffi::OsStr;
+use std::fs::File;
+use std::io::{BufRead, BufReader};
+use std::path::{Path, PathBuf};
+use std::process::Command;
+
+const EXTENSIONS: &[&str] =
+    &["rs", "py", "js", "sh", "c", "cpp", "h", "md", "css", "ftl", "toml", "yml", "yaml"];
+
+fn has_supported_extension(path: &Path) -> bool {
+    path.extension().is_some_and(|ext| EXTENSIONS.iter().any(|e| ext == OsStr::new(e)))
+}
+
+fn list_tracked_files() -> Result<Vec<PathBuf>, String> {
+    let output = Command::new("git")
+        .args(["ls-files", "-z"])
+        .output()
+        .map_err(|e| format!("Failed to run `git ls-files`: {e}"))?;
+
+    if !output.status.success() {
+        let stderr = String::from_utf8_lossy(&output.stderr);
+        return Err(format!("`git ls-files` failed: {stderr}"));
+    }
+
+    let mut files = Vec::new();
+    for entry in output.stdout.split(|b| *b == 0) {
+        if entry.is_empty() {
+            continue;
+        }
+        let path = std::str::from_utf8(entry).unwrap();
+        files.push(PathBuf::from(path));
+    }
+
+    Ok(files)
+}
+
+pub(crate) fn run() -> Result<(), String> {
+    let files = list_tracked_files()?;
+    let mut error_count = 0;
+    // Avoid embedding the task marker in source so greps only find real occurrences.
+    let todo_marker = "todo".to_ascii_uppercase();
+
+    for file in files {
+        if !has_supported_extension(&file) {
+            continue;
+        }
+
+        let file_handle =
+            File::open(&file).map_err(|e| format!("Failed to open {}: {e}", file.display()))?;
+        let reader = BufReader::new(file_handle);
+
+        for (i, line) in reader.lines().enumerate() {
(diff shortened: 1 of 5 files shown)


reddevilmidzy:
https://github.com/rust-lang/rustc_codegen_gcc/blob/66d6f7bcd2250f5041637304c6883238e68e8099/.cspell.json#L24-L26

Since TODOs are banned, can I remove TODO here too?

antoyo:
Yes

reddevilmidzy:
Ah, there are three TODOs left.

antoyo:
Is this ready for review?
I'm not sure what's the status here.

Also, could you please update the message of all TODOs that were converted to FIXMEs so that they make sense as a FIXME?

reddevilmidzy:
Yes, I think I am now ready to be reviewed!
First, I will modify the FIXME messages.

reddevilmidzy:
It seems like there are quite a few messages that need fixing. Is it okay if I continue working on them in this PR?

antoyo:
Yes, this is good.

src/int.rs
```diff
@@ -862,12 +862,13 @@ impl<'gcc, 'tcx> CodegenCx<'gcc, 'tcx> {
         self.bitwise_operation(BinaryOp::BitwiseOr, a, b, loc)
     }
 
-    // FIXME(antoyo): can we use https://github.com/rust-lang/compiler-builtins/blob/master/src/int/mod.rs#L379 instead?
```
reddevilmidzy, line 865: *[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustc_codegen_gcc/861/9886ca55fdba1339f6811bee3cbb5a75229d3ccd..4afa4afdf39da158f1ca202bf8d3f4fcf8060ac8)*

Looking at the history, it seems it was probably pointing to this code.

https://github.com/rust-lang/compiler-builtins/blob/2ad41ef64b870e32f38e13a0fd05c4d3bceb287a/src/int/mod.rs#L379

reddevilmidzy:
I only modified the FIXME messages changed in this commit. https://github.com/rust-lang/rustc_codegen_gcc/commit/241a104ce819112c46c593338e2fc09ca9a8862c

reddevilmidzy:
Hello @antoyo, Could you review this when you have time?

I changed the remaining TODOs to FIXMEs and modified the sentences to have a FIXME feel. I also added the check-todo job to CI.

thanks:)

antoyo:
> Hello @antoyo, Could you review this when you have time?

Yes, I've been a bit busy lately.
It should take me a few more days to review this PR.
Thanks.

antoyo:
Sorry for the delay.

I do have a pending review for this, but I'm not sure we're going to the right path.
What I suggest is to keep the original messages and just merge the TODO checker.
I know I asked you to update the messages, so I'm sorry for that: I thought that was the right call here.
The reason I suggest to keep the original messages is to avoid a long back-and-forth of review/addressing the review.
How do you feel about this? Would you be OK with this?
Again, sorry for this.

reddevilmidzy:
No worries! I agree with your suggestion. I'll revert the message changes and keep only the TODO checker changes.
Thanks for taking another look at this.

build_system/src/todo.rs
```diff
@@ -0,0 +1,69 @@
...
+
+    for file in files {
+        if !has_supported_extension(&file) {
+            continue;
+        }
+
+        let bytes = fs::read(&file).unwrap();
+        let contents = std::str::from_utf8(&bytes).unwrap();
+
+        for (i, line) in contents.split('\n').enumerate() {
```
antoyo: Since you split on lines anyway, I would instead suggest to use a BufReader to avoid reading the whole files in memory.

antoyo (commented):
I had this one comment that is still relevant.
After that, I can merge this.
Thanks for your work!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustc_codegen_gcc/861/cd4b93ac5052be9f53e2f8fcda7c4b097feeb950..7a755714bd560615c5dec1d173f6e390ea587283)*

build_system/src/todo.rs
reddevilmidzy: Sounds good! I've addressed it.

antoyo (approved):

antoyo:
Thanks for your contribution.
And sorry again.

PR #876, opened 2026-04-28
fix: handle zero-sized memory intrinsics

### LINKED ISSUE #516: Make sure memcpy/memmove/memset with size 0 behave correctly
Zero-sized memory accesses are [now](https://github.com/rust-lang/rust/pull/117329) always permitted, even if the pointer is NULL or dangling (but it must be aligned still). For codegen this means in particular that memcpy/memmove/memset must be lowered to operations that are never UB when the size is 0 (and the pointer is sufficiently aligned). In LLVM that's easy as LLVM's corresponding intrinsics explicitly allow size 0. However, in C, memcpy/memmove/memset with size 0 is UB on NULL (and dangling pointers are impossible to even mention in C), so GCC may use a different semantics for its builtins. For Rust's GCC backend, it's crucial that we use GCC builtins that allow size 0 with any pointer.

mizoz:
Fixes rust-lang/rustc_codegen_gcc#516.

This avoids lowering zero-sized `copy`, `copy_nonoverlapping`, and `write_bytes` intrinsics to GCC memory builtins. For zero-sized element types, the operation has no bytes to touch, so the backend can return before the fallback lowering emits `memcpy`, `memmove`, or `memset`.

A small run test covers the three zero-sized memory intrinsic paths using `()` pointers.

Tested with:

- `git diff --check`
- `cargo fmt --check`
- `./y.sh test --mini-tests`
- `cargo test -p rustc_codegen_gcc --test lang_tests zero_sized_mem -- --nocapture`

### DIFF
--- a/src/intrinsic/mod.rs
+++ b/src/intrinsic/mod.rs
@@ -338,13 +338,6 @@ impl<'a, 'gcc, 'tcx> IntrinsicCallBuilderMethods<'tcx> for Builder<'a, 'gcc, 'tc
                 );
                 return Ok(());
             }
-            sym::copy | sym::copy_nonoverlapping | sym::write_bytes => {
-                let ty = fn_args.type_at(0);
-                if self.layout_of(ty).size.bytes() == 0 {
-                    return Ok(());
-                }
-                return Err(Instance::new_raw(instance.def_id(), instance.args));
-            }
             sym::breakpoint => {
                 unimplemented!();
             }
--- a/src/intrinsic/mod.rs
+++ b/src/intrinsic/mod.rs
@@ -338,6 +338,13 @@ impl<'a, 'gcc, 'tcx> IntrinsicCallBuilderMethods<'tcx> for Builder<'a, 'gcc, 'tc
                 );
                 return Ok(());
             }
+            sym::copy | sym::copy_nonoverlapping | sym::write_bytes => {
+                let ty = fn_args.type_at(0);
+                if self.layout_of(ty).size.bytes() == 0 {
+                    return Ok(());
+                }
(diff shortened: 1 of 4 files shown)


src/intrinsic/mod.rs
```diff
@@ -338,6 +338,13 @@ impl<'a, 'gcc, 'tcx> IntrinsicCallBuilderMethods<'tcx> for Builder<'a, 'gcc, 'tc
                 );
                 return Ok(());
             }
+            sym::copy | sym::copy_nonoverlapping | sym::write_bytes => {
+                let ty = fn_args.type_at(0);
+                if self.layout_of(ty).size.bytes() == 0 {
+                    return Ok(());
+                }
+                return Err(Instance::new_raw(instance.def_id(), instance.args));
+            }
```
bjorn3: Why not change the `memset`, `memcpy` and `memmove` methods on `Builder`? Also this only helps with ZST types, not with copying zero elements.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustc_codegen_gcc/876/9886ca55fdba1339f6811bee3cbb5a75229d3ccd..88c3abfb97d16bb2db8088f970b389b241974638)*

mizoz:
You're right — this should cover the generic `memcpy`/`memmove`/`memset` paths and not only the ZST intrinsic case.

I had avoided changing the builder methods because a first pass that split the current block inside those methods caused later MIR emission to keep writing to the terminated block in local testing. I'm going to rework this instead of trying to paper over it, so the updated patch handles runtime zero sizes cleanly as well.

PR #880, opened 2026-04-30
Handle all modules being serialized during LTO

bjorn3:


bjorn3:
Blocked on https://github.com/rust-lang/rust/pull/156003

antoyo:
I forgot what's the status regarding LTO.
Can you please remind me where we're at?
Is this for ThinLTO?

bjorn3:
This is for fat LTO. This currently only fixes something that can happen when you combine incremental compilation with fat LTO and all codegen units are unchanged.

I want to make some changes to cg_llvm and cg_gcc that would cause this code path to be unconditionally hit for fat LTO. I don't know yet if the performance of that would be acceptable though. I wanted to land the fix in this PR first to make sure I don't regress anything in cg_gcc.

bjorn3:
CI passes if I force the new code path to be used. Reverted back to only using it when necessary. Ready for review.

antoyo (approved):

antoyo:
Thanks for your contribution.

Issue #825, opened 2026-01-23
Support f16 on targets where GCC doesn't support it


antoyo:
@tgross35 mentioned we could support it the following way:

```
1. Treat it as a u16 in ABI (calling convention, alignment, etc)
2. To extend: call `__extendhfsf2` for f16->f32, `__extendhfdf2` for f16->f64
3. To trunc: call `__truncsfhf2` for f32->f16, `__truncdfhf2` for f64->f16
4. For arithmetic: extend to `f32`, perform the op, truncate back to `f16`
5. For casts to/from int: go via f32
6. For math ops: if I get https://github.com/rust-lang/rust/pull/150946 over the line then this won't be needed in most cases, but anything else can go via f32
```

tgross35:
Context thread [#rustc-codegen-gcc > Fallback for f16](https://rust-lang.zulipchat.com/#narrow/channel/386786-rustc-codegen-gcc/topic/Fallback.20for.20f16/with/569656835)

PR #885, opened 2026-05-09
Treat f16 as a u16 in ABI

cijiugechu:
Part1 of rust-lang/rustc_codegen_gcc#825 .

### DIFF
--- a/build_system/src/test.rs
+++ b/build_system/src/test.rs
@@ -515,7 +515,6 @@ fn setup_rustc(env: &mut Env, args: &TestArg) -> Result<PathBuf, String> {
     // If the repository was already cloned, command will fail, so doesn't matter.
     let _ = git_clone("https://github.com/rust-lang/rust.git", Some(&rust_dir_path), false);
     let rust_dir: Option<&Path> = Some(&rust_dir_path);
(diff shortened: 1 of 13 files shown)


antoyo:
Sorry, I missed this PR (feel free to ping me after some time when this happens).

I would be more comfortable merging this if at least one of the following parts (2 to 6) was implemented as well as some CI for this (not sure what would be the best way to test this).

cijiugechu:
> Sorry, I missed this PR (feel free to ping me after some time when this happens).
> 
> I would be more comfortable merging this if at least one of the following parts (2 to 6) was implemented as well as some CI for this (not sure what would be the best way to test this).

Thanks for taking a look!
That makes sense. I'm already working on implementing part 2 and 5 on this branch, so I'll push an update to this PR once those are ready.

antoyo:
Since the CI passed on the last commit, was that ready for review?

tgross35:
I'm only reading the description here and not the code: this only takes effect when GCC doesn't already support `f16`, right?

antoyo:
Yes, that's indeed for targets where f16 is not supported.
This PR is the first steps towards having #825, which follows your idea.

PR #891, opened 2026-05-29
builder: Emit a plain jump for a switch with no cases

### LINKED ISSUE #881: ICE libgccjit.so: error: : unterminated block (`-Zmir-preserve-ub`)
snippet:
````rust
//@compile-flags: -Zcodegen-backend=gcc -Clink-dead-code=true -Zmir-preserve-ub
const LOW_RANGE: char = '0';
const HIGH_RANGE: char = '9';

fn main() {
    match '5' {
        LOW_RANGE..=HIGH_RANGE => (),
        _ => (),
    };
}

````

Version information
````
rustc 1.97.0-nightly (f53b654a8 2026-04-30)
binary: rustc
commit-hash: f53b654a8882fd5fc036c4ca7a4ff41ce32497a6
commit-date: 2026-04-30
host: x86_64-unknown-linux-gnu
release: 1.97.0-nightly
LLVM version: 22.1.4

````

Command:
`/home/matthias/.rustup/toolchains/master/bin/rustc -Zcodegen-backend=gcc -Clink-dead-code=true -Zmir-preserve-ub`

<details><summary><strong>Program output</strong></summary>
<p>

```
libgccjit.so: error: : gcc_jit_block_end_with_switch: expr: (__uint32_t  __attribute__((aligned(4))))(__uint32_t  __attribute__((aligned(4))))(__int32_t  __attribute__((aligned(4))))53 <= (__uint32_t  __attribute__((aligned(4))))(__uint32_t  __attribute__((aligned(4))))(__int32_t  __attribute__((aligned(4))))57 (type: bool) is not of integer type
libgccjit.so: error: : unterminated block in _RNvCs5Oe9U7VPi96_4mvce4main: bb1
  |
  = note:  "cc" "-m64" "/tmp/icemaker_global_tempdir.r7xD27Zz1GAO/rustc_testrunner_tmpdir_reporting.jt7Ecpyc2vaT/rustcyJOnmm/symbols.o" "<2 object files omitted>" "-Wl,--as-needed" "-Wl,-Bstatic" "<sysroot>/lib/rustlib/x86_64-unknown-linux-gnu/lib/{libstd-*,libpanic_unwind-*,libobject-*,libmemchr-*,libaddr2line-*,libgimli-*,libcfg_if-*,librustc_demangle-*,libstd_detect-*,libhashbrown-*,librustc_std_workspace_alloc-*,libminiz_oxide-*,libadler2-*,libunwind-*,liblibc-*,librustc_std_workspace_core-*,liballoc-*,libcore-*,libcompiler_builtins-*}.rlib" "-Wl,-Bdynamic" "-lgcc_s" "-lutil" "-lrt" "-lpthread" "-lm" "-ldl" "-lc" "-L" "/tmp/icemaker_global_tempdir.r7xD27Zz1GAO/rustc_testrunner_tmpdir_reporting.jt7Ecpyc2vaT/rustcyJOnmm/raw-dylibs" "-B<sysroot>/lib/rustlib/x86_64-unknown-linux-gnu/bin/gcc-ld" "-fuse-ld=lld" "-Wl,--eh-frame-hdr" "-Wl,-z,noexecstack" "-L" "<sysroot>/lib/rustlib/x86_64-unknown-linux-gnu/lib" "-o" "outputfile" "-pie" "-Wl,-z,relro,-z,now" "-nodefaultlibs"
  = note: some arguments are omitted. use `--verbose` to show all linker arguments
  = note: rust-lld: error: cannot open outputfile.mvce.43ac9ae409ec8616-cgu.0.rcgu.o: No such file or directory
          

error: aborting due to 1 previous error


```

</p>
</details>

### PR DESCRIPTION
**Pull request by Dnreikronos**
Closes rust-lang/rustc_codegen_gcc#881

A range-pattern match compiled with `-Zmir-preserve-ub` can leave a `SwitchInt` whose only target is `otherwise`. That reaches the backend as a switch with an empty case list, and in the reproducer the discriminant is the `bool` result of the range comparison. `gcc_jit_block_end_with_switch` only accepts an integer discriminant, so it errors out and leaves the block unterminated.

A switch with no cases always goes to the default block, so emit a plain jump in that case rather than building a degenerate switch. This is also what `codegen_ssa` already relies on for the LLVM backend, which tolerates the empty switch.

### Verification

Built the backend locally and ran the exact reproducer from the issue with `-Zmir-preserve-ub -Clink-dead-code=true`:

- Without this change, it reproduces the reported error verbatim:

  ```
  libgccjit.so: error: : gcc_jit_block_end_with_switch: expr: (...)53 <= (...)57 (type: bool) is not of integer type
  libgccjit.so: error: : unterminated block in ..._main: bb1
  ```

- With this change, it compiles cleanly and the resulting binary runs (exit 0).

The MIR confirms the shape. `bb0` holds `switchInt(_2: bool) -> [0: bb2, otherwise: bb1]`, which has a single case and is lowered to a conditional branch by `codegen_ssa`, so it never reaches here. `bb1` holds `switchInt(_3: bool) -> bb2`, a switch with zero cases, which is the one that hit the empty-case path.

No in-tree regression test, since the `lang_tester` harness has no per-file compile-flags and the bug needs `-Zmir-preserve-ub`. Happy to add one if there's a preferred spot.

### DIFF
--- a/tests/run/mir_preserve_ub_empty_switch.rs
+++ b/tests/run/mir_preserve_ub_empty_switch.rs
@@ -1,0 +1,35 @@
+// ignore-if: test -z "$CARGO_TEST_FLAGS"
+// Compiler:
+//
+// Run-time:
+//   status: 0
+
+// Regression test for https://github.com/rust-lang/rustc_codegen_gcc/issues/881
+//
+// This needs `-Zmir-preserve-ub`, so it is skipped unless that flag is passed
+// through `CARGO_TEST_FLAGS` (see the `ignore-if` directive above). Run it with:
+//   CARGO_TEST_FLAGS="-Zmir-preserve-ub" ./y.sh test --cargo-tests -- mir_preserve_ub_empty_switch
+
+#![feature(no_core)]
+#![no_std]
+#![no_core]
+#![no_main]
+
+extern crate mini_core;
+use intrinsics::black_box;
+use mini_core::*;
+
+#[no_mangle]
+extern "C" fn main(argc: i32, _argv: *const *const u8) -> i32 {
+    // With `-Zmir-preserve-ub`, the range pattern below is lowered to a pair of
+    // comparisons and the second one becomes a `SwitchInt` with no cases (only
+    // an `otherwise` target) whose discriminant is the `bool` comparison
+    // result. `gcc_jit_block_end_with_switch` rejects a non-integer
+    // discriminant, so the backend must emit a plain jump for it instead.
+    let value = black_box(argc);
+    match value {
+        0..=9 => (),
+        _ => (),
+    }
+    0
+}
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -98,6 +98,7 @@ jobs:
       run: |
         ./y.sh build --sysroot
         ./y.sh test --cargo-tests
+        CARGO_TEST_FLAGS="-Zmir-preserve-ub" ./y.sh test --cargo-tests -- mir_preserve_ub_empty_switch
 
     - name: Run y.sh cargo build
       run: |
--- a/src/builder.rs
+++ b/src/builder.rs
@@ -570,6 +570,18 @@ impl<'a, 'gcc, 'tcx> BuilderMethods<'a, 'tcx> for Builder<'a, 'gcc, 'tcx> {
         default_block: Block<'gcc>,
         cases: impl ExactSizeIterator<Item = (u128, Block<'gcc>)>,
     ) {
+        // A switch with no cases is equivalent to an unconditional jump to the
+        // default block. Such a `SwitchInt` (one with only an `otherwise` target)
+        // is normally simplified into a `goto`, but `-Z mir-preserve-ub` keeps it,
+        // so it can reach here with e.g. the `bool` discriminant produced by a
+        // range-pattern comparison. `gcc_jit_block_end_with_switch` rejects a
+        // discriminant that is not of integer type, so emit a plain jump instead
+        // of a (pointless) switch.
+        if cases.len() == 0 {
+            self.block.end_with_jump(self.location, default_block);
+            return;
+        }
+
         let mut gcc_cases = vec![];
         let typ = self.val_ty(value);
         // FIXME(FractalFir): This is a workaround for a libgccjit limitation.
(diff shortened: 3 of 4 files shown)

### CODE REVIEW

**Comment by antoyo** [comment]
Thanks for your work.

I believe we might be able to add the test the following way.
We could run it with adding a new env var (similar to `TEST_FLAGS`), and run it this way:
```
CARGO_TEST_FLAGS="-Zmir-preserve-ub" ./y.sh test --cargo-tests -- NEW_TEST_NAME
```
and add a commented line that does something like `// ignore-if: <check-that-env-var-CARGO_TEST_FLAGS-is-not-set>`.

**Comment by Dnreikronos** [comment]
> Thanks for your work.
> 
> I believe we might be able to add the test the following way. We could run it with adding a new env var (similar to `TEST_FLAGS`), and run it this way:
> 
> ```
> CARGO_TEST_FLAGS="-Zmir-preserve-ub" ./y.sh test --cargo-tests -- NEW_TEST_NAME
> ```
> 
> and add a commented line that does something like `// ignore-if: <check-that-env-var-CARGO_TEST_FLAGS-is-not-set>`.

Just adjusted :)

File: tests/run/mir_preserve_ub_empty_switch.rs [hunk 1]
```diff
@@ -0,0 +1,35 @@
...
+// Compiler:
+//
+// Run-time:
+//   status: 0
+
+// Regression test for https://github.com/rust-lang/rustc_codegen_gcc/issues/881
+//
+// This needs `-Zmir-preserve-ub`, so it is skipped unless that flag is passed
+// through `CARGO_TEST_FLAGS` (see the `ignore-if` directive above). Run it with:
+//   CARGO_TEST_FLAGS="-Zmir-preserve-ub" ./y.sh test --cargo-tests -- mir_preserve_ub_empty_switch
```
antoyo, line 11: Please add this line in the CI so that the test is actually run in the CI.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustc_codegen_gcc/891/c0bb5745fcd5722d06ec73217f19e82ba5667d11..fdd3536ce6084ad1d106dbb20c11d04e0b17409d)*

Dnreikronos: Adjusted!
Can you re-validate, please? :)

antoyo (approved):

antoyo:
Thanks for your contribution!

PR #893, opened 2026-05-30
Lower x86-interrupt byval first argument as pointer

### LINKED ISSUE #833: interrupt service routine should have a pointer as the first argument
snippet:
````rust
//@compile-flags: -Zcodegen-backend=gcc -Clink-dead-code=true --crate-type=lib
#![feature(abi_x86_interrupt)]
pub extern "x86-interrupt" fn f(a: i64) {}

````

Version information
````
rustc 1.95.0-nightly (a18e6d9d1 2026-01-24)
binary: rustc
commit-hash: a18e6d9d1473d9b25581dd04bef6c7577999631c
commit-date: 2026-01-24
host: x86_64-unknown-linux-gnu
release: 1.95.0-nightly
LLVM version: 21.1.8

````

Command:
`/home/matthias/.rustup/toolchains/master/bin/rustc -Zcodegen-backend=gcc -Clink-dead-code=true --crate-type=lib`

<details><summary><strong>Program output</strong></summary>
<p>

```
warning: unused variable: `a`
 --> /tmp/icemaker_global_tempdir.ixZrmjiu4rgx/rustc_testrunner_tmpdir_reporting.fd7j7VvF0nFL/mvce.rs:2:33
  |
2 | pub extern "x86-interrupt" fn f(a: i64) {}
  |                                 ^ help: if this is intentional, prefix it with an underscore: `_a`
  |
  = note: `#[warn(unused_variables)]` (part of `#[warn(unused)]`) on by default

libgccjit.so: error: : interrupt service routine should have a pointer as the first argument
error: failed to build archive at `outputfile`: failed to open object file: No such file or directory (os error 2)

error: aborting due to 1 previous error; 1 warning emitted


```

</p>
</details>

### PR DESCRIPTION
**Pull request by Dnreikronos**
Fixes rust-lang/rustc_codegen_gcc#833

An `extern "x86-interrupt"` function can have a by-value first Rust argument, such as `i64` or a by-value struct. rustc's LLVM backend lowers that first byval stack argument to a pointer-shaped backend parameter for `x86_intrcc`.

GCC's `interrupt` attribute expects the same pointer-shaped first parameter. Without that lowering, libgccjit reports `interrupt service routine should have a pointer as the first argument` and does not write an object file, which then shows up as a confusing follow-up archive/object error.

Mirror rustc's lowering in the GCC backend for the first `x86-interrupt` byval argument, while keeping the interrupt attribute. This avoids the libgccjit abort without rejecting a Rust signature that rustc accepts.

The regression test covers scalar and by-value struct first arguments.

### DIFF
--- a/src/abi.rs
+++ b/src/abi.rs
@@ -159,7 +159,10 @@ impl<'gcc, 'tcx> FnAbiGccExt<'gcc, 'tcx> for FnAbi<'tcx, Ty<'tcx>> {
         #[cfg(not(feature = "master"))]
         let apply_attrs = |ty: Type<'gcc>, _attrs: &ArgAttributes, _arg_index: usize| ty;
 
-        for arg in self.args.iter() {
+        for (source_arg_index, arg) in self.args.iter().enumerate() {
+            #[cfg(not(feature = "master"))]
+            let _ = source_arg_index;
+
             let arg_ty = match arg.mode {
                 PassMode::Ignore => continue,
                 PassMode::Pair(a, b) => {
@@ -185,9 +188,31 @@ impl<'gcc, 'tcx> FnAbiGccExt<'gcc, 'tcx> for FnAbi<'tcx, Ty<'tcx>> {
                     apply_attrs(ty, &cast.attrs, argument_tys.len())
                 }
                 PassMode::Indirect { attrs: _, meta_attrs: None, on_stack: true } => {
-                    // This is a "byval" argument, so we don't apply the `restrict` attribute on it.
-                    on_stack_param_indices.insert(argument_tys.len());
-                    arg.layout.gcc_type(cx)
+                    let x86_interrupt_first_arg = {
+                        #[cfg(feature = "master")]
+                        {
+                            source_arg_index == 0
+                                && matches!(self.conv, CanonAbi::Interrupt(InterruptKind::X86))
+                        }
+                        #[cfg(not(feature = "master"))]
+                        {
+                            false
+                        }
+                    };
+
+                    if x86_interrupt_first_arg {
+                        // Rust lowers the first `x86-interrupt` argument as a byval stack slot.
+                        // LLVM represents that as a pointer parameter with `byval`; GCC's
+                        // interrupt attribute likewise requires a pointer-shaped first parameter.
+                        // Do not add this parameter to `on_stack_param_indices`: that set is only
+                        // needed when GCC represents a byval argument as a value parameter, while
+                        // this parameter is already pointer-shaped.
+                        cx.type_ptr_to(arg.layout.gcc_type(cx))
+                    } else {
+                        // This is a "byval" argument, so we don't apply the `restrict` attribute on it.
+                        on_stack_param_indices.insert(argument_tys.len());
+                        arg.layout.gcc_type(cx)
+                    }
                 }
                 PassMode::Direct(attrs) => {
                     apply_attrs(arg.layout.immediate_gcc_type(cx), &attrs, argument_tys.len())
--- a/src/attributes.rs
+++ b/src/attributes.rs
@@ -2,6 +2,8 @@
 use gccjit::FnAttribute;
 use gccjit::Function;
 #[cfg(feature = "master")]
+use rustc_abi::{CanonAbi, InterruptKind};
+#[cfg(feature = "master")]
 use rustc_hir::attrs::InlineAttr;
 use rustc_hir::attrs::InstructionSetAttr;
 #[cfg(feature = "master")]
@@ -9,6 +11,7 @@ use rustc_middle::middle::codegen_fn_attrs::CodegenFnAttrFlags;
 #[cfg(feature = "master")]
 use rustc_middle::mir::TerminatorKind;
 use rustc_middle::ty;
+use rustc_target::callconv::FnAbi;
 #[cfg(feature = "master")]
 use rustc_target::spec::Arch;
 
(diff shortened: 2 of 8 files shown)

### CODE REVIEW

**Comment by Dnreikronos** [comment]
@antoyo, hi mate!
can you help me with this, please?

**Comment by antoyo** [comment]
Sorry for the delay.
I'll try to review this in the following days.

**Comment by Dnreikronos** [comment]
> Sorry for the delay. I'll try to review this in the following days.

all good!
ty!

**Comment by antoyo** [comment]
> LLVM only warns on the same code.

What is the warning produced by LLVM? (I don't see it.)

One question I have is: is the code in the issue supposed to work correctly, or does Rust also always expect a pointer argument?

**Comment by Dnreikronos** [comment]
yeah, i checked this again and i think the pr description is a bit off here.

i don't see an llvm warning either. with rustc 1.96.0 / llvm 22.1.2, the testcase compiles cleanly for me. dumping the llvm ir shows rustc lowers the first arg to a pointer/byval param:

define x86_intrcc void ... (ptr byval([8 x i8]) align 8 %_a)

so imo rust is not expecting the rust-level arg itself to be a pointer. it accepts something like i64 or a by-value struct, then lowers it into the pointer shape the backend wants.

tbh i think the better fix is to mirror that lowering in the gcc backend, instead of rejecting the source signature or dropping the interrupt attr. i'll update the pr in that direction.

File: src/abi.rs [hunk 1]
```diff
@@ -185,9 +188,28 @@ impl<'gcc, 'tcx> FnAbiGccExt<'gcc, 'tcx> for FnAbi<'tcx, Ty<'tcx>> {
...
                 }
                 PassMode::Indirect { attrs: _, meta_attrs: None, on_stack: true } => {
-                    // This is a "byval" argument, so we don't apply the `restrict` attribute on it.
-                    on_stack_param_indices.insert(argument_tys.len());
-                    arg.layout.gcc_type(cx)
+                    let x86_interrupt_first_arg = {
+                        #[cfg(feature = "master")]
+                        {
+                            source_arg_index == 0
+                                && matches!(self.conv, CanonAbi::Interrupt(InterruptKind::X86))
```
antoyo, line 195: Is this necessary only on x86 or will other platforms get the same issue?

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustc_codegen_gcc/893/2ca8d7f385b24c30822075b69076e69aa3c7cf6c..39f95374abf135de12ea2b49671c961c9dc5a64e)*

src/abi.rs
```diff
@@ -185,9 +188,28 @@ impl<'gcc, 'tcx> FnAbiGccExt<'gcc, 'tcx> for FnAbi<'tcx, Ty<'tcx>> {
...
+                        #[cfg(not(feature = "master"))]
+                        {
+                            false
+                        }
+                    };
+
+                    if x86_interrupt_first_arg {
+                        // Rust lowers the first `x86-interrupt` argument as a byval stack slot.
+                        // LLVM represents that as a pointer parameter with `byval`; GCC's
+                        // interrupt attribute likewise requires a pointer-shaped first parameter.
```
antoyo, line 206: Please edit the comment to make it clearer that we don't use on_stack_param_indices (byval) on purpose here.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustc_codegen_gcc/893/2ca8d7f385b24c30822075b69076e69aa3c7cf6c..39f95374abf135de12ea2b49671c961c9dc5a64e)*

src/attributes.rs
```diff
@@ -120,39 +134,48 @@ pub fn from_fn_attrs<'gcc, 'tcx>(
...
-    function_features.extend(&mut global_features);
-    let target_features = function_features
-        .iter()
-        .filter_map(|feature| {
-            // FIXME(antoyo): support soft-float.
-            if feature.contains("soft-float") {
-                return None;
-            }
+    let target_features = if x86_interrupt {
+        "general-regs-only".to_string()
```
antoyo: Was there any reason to make this replace all the target features?
I believe that it would work if we append `"general-regs-only"` instead of dropping the other features here.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustc_codegen_gcc/893/2ca8d7f385b24c30822075b69076e69aa3c7cf6c..39f95374abf135de12ea2b49671c961c9dc5a64e)*

src/abi.rs
bjorn3: I'm pretty sure x86 is an exception.

Dnreikronos: > I'm pretty sure x86 is an exception.

Yeah, it is

src/attributes.rs
bjorn3: Also why is this set in the first place? Either you kernel globally disables SSE and AVX anyway or any call you make from an `extern "x86-interrupt"` call will clobber all registers anyway. I don't think the LLVM backend enables this feature either.

src/abi.rs
Dnreikronos: yep, fair point. i updated the comment to say why we intentionally skip on_stack_param_indices here. that set is for byval args represented as GCC values, but this one is already a pointer. imo the intent is way clearer now.

src/attributes.rs
Dnreikronos: yeah, makes sense. tbh replacing the whole feature list was too broad, so i changed it to keep the normal function + global features and append general-regs-only.

btw, i checked why we need it in the first place. GCC checks the enabled target flags, not the instructions actually emitted, so default x86_64 SSE2 is enough to reject even an empty interrupt fn. the LLVM backend does not add this because x86_intrcc saves/restores the SIMD state that can be clobbered, while GCC's interrupt attribute explicitly does not preserve SSE/MMX/x87 state.

imo keeping this function-local is safer than assuming the whole kernel disabled those features. you are right that it does not make arbitrary callees safe, though; they still need the same restriction or a global no-SIMD setup. fyi, this also keeps the existing LTO feature handling.

antoyo:
@Dnreikronos: Oh it seems my last comment was not sent correctly.
Here it is:

Please add this as a comment (why we add `general-regs-only`) so that the reasoning stays in the code.
After that, I think we'll be good to merge this.

antoyo (approved):

antoyo:
@Dnreikronos Did you make any change since my last approval? (I'm asking because you requested another review.)
Is this PR ready to be merged if the CI passes?

Dnreikronos:
> @Dnreikronos Did you make any change since my last approval? (I'm asking because you requested another review.) Is this PR ready to be merged if the CI passes?

I just squashed all the commits to clean the history and be easier to understand everything. And yes, it's ready the be merged when the CI passses

antoyo:
Thanks for your contribution!

PR #908, opened 2026-07-14
Test CI

antoyo:


folkertdev:
Well I suspect the stdarch failure is related to https://github.com/rust-lang/stdarch/pull/2184?

folkertdev:
I merged that back into the main repo in https://github.com/rust-lang/rust/pull/158968, but I think the PR to remove the target feature altogether is still not merged https://github.com/rust-lang/rust/pull/158910.

`stdarch` still seems to work here, so then I suspect GCC also dropped support for this feature, and you need to sync what stdarch you're compiling?

antoyo:
So, this fails with the updated GCC version.
Are you saying I only need to sync with the Rust repo in order to fix this?

folkertdev:
I don't know your exact setup, but if you can get CI here to use the stdarch version that `rust-lang/rust` uses right now, I believe this issue should disappear.

PR #910, opened 2026-07-16
use core fallbacks for more `f16`/`f128` operations

folkertdev:
This PR removes some explicit handling of functions that since https://github.com/rust-lang/rust/pull/150946 have a fallback via libm.

We plan to do this for more operations (especially for `f16`, where possible for `f128`) to be able to remove custom logic from the backends.

### DIFF
--- a/src/intrinsic/mod.rs
+++ b/src/intrinsic/mod.rs
@@ -95,7 +95,6 @@ fn get_simple_intrinsic<'gcc, 'tcx>(
         sym::floorf64 => "floor",
         sym::ceilf32 => "ceilf",
         sym::ceilf64 => "ceil",
-        sym::powf128 => return float_intrinsic(cx, cx.type_f128(), "powf128"),
         sym::truncf32 => "truncf",
         sym::truncf64 => "trunc",
         // We match the LLVM backend and lower this to `rint`.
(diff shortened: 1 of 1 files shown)


src/intrinsic/mod.rs
```diff
@@ -263,11 +251,6 @@ impl<'a, 'gcc, 'tcx> IntrinsicCallBuilderMethods<'tcx> for Builder<'a, 'gcc, 'tc
             | sym::roundf128
             | sym::round_ties_even_f128
             | sym::sqrtf128
-            | sym::expf128
-            | sym::exp2f128
-            | sym::logf128
-            | sym::log2f128
-            | sym::log10f128
```
antoyo, line 270: Should we also remove the cases for those in get_simple_function_f128?

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustc_codegen_gcc/910/2467b6732cae3e91aaea3ee7ae7738b510cafc45..d4c83de163c9257b147088decf33d098dc03fca5)*

folkertdev: yes, fixed

antoyo (approved):

antoyo:
Nice improvement!
Thanks for your contribution!

PR #911, opened 2026-07-16
Support guaranteed tail calls

folkertdev:
Add support for `become` using GCC's `musttail` implementation. Depends on

- https://github.com/rust-lang/gccjit.rs/pull/71

### DIFF
--- a/src/builder.rs
+++ b/src/builder.rs
@@ -36,7 +36,6 @@ use rustc_target::spec::{HasTargetSpec, HasX86AbiOpt, Target, X86Abi};
 use crate::abi::FnAbiGccExt;
 use crate::common::{SignType, TypeReflection, type_is_pointer};
 use crate::context::CodegenCx;
-use crate::errors;
 use crate::intrinsic::llvm;
 use crate::type_of::LayoutGccExt;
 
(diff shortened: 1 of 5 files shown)


src/builder.rs
folkertdev, line 440: is this workaround still needed? I think this code can be refactored a bunch (sharing this assignment logic between direct and pointer calls) if it is now redundant.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustc_codegen_gcc/911/3aeb91f96a82c5ef18545a516b9bfe47fef2d2f0..9bce153ad31ecc1481de017fb29430a573b532b4)*

antoyo: You can try removing this code, but I believe this refers to some intrinsics that we translate to __builtin_trap because it's not yet supported, so this is likely still needed.
A refactor could possibly allow us to remove this hack, though.

folkertdev: well if you're ok with the changes then you can merge this. I made a separate PR to check if CI fails with that hack removed

tests/failing-ui-tests.txt
```diff
@@ -109,5 +103,4 @@ tests/ui/eii/static/default.rs
 tests/ui/eii/static/default_cross_crate.rs
 tests/ui/eii/static/default_explicit.rs
 tests/ui/eii/static/default_cross_crate_explicit.rs
-tests/ui/explicit-tail-calls/default-trait-method.rs
 tests/ui/explicit-tail-calls/tailcc-no-signature-restriction.rs
```
antoyo, line 106: I suspect this is normal that this test still fails.
Is this correct?

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustc_codegen_gcc/911/3aeb91f96a82c5ef18545a516b9bfe47fef2d2f0..61a02cd8e08d30a49de652b138f5f2effb80aad2)*

antoyo (approved):

tests/failing-ui-tests.txt
folkertdev: yeah `tailcc` is an llvm-specific calling convention, it'll hit an assert somewhere.

antoyo:
Thanks for your contribution!

PR #912, opened 2026-07-16
[experiment] refactor call code

folkertdev:


folkertdev:
Hmm, actually, would CI catch what that workaround was meant to solve?

antoyo:
If the workaround was what I think (the __builtin_trap workaround), it's in the stdarch tests, so the CI would indeed catch it.
Perhaps my memory fails me or the comment is outdated, though.

folkertdev:
I don't think I can do much better here at the moment

PR #918, opened 2026-07-21
Add support for aliases

GuillaumeGomez:
Now only 9 `tests/ui/eii` tests are failing out of 60. I think the remaining ones comes from the `link_section` that we don't support yet.

Considering most of the work is already done, I think it's worth merging this first step already. :)


antoyo (approved):

antoyo:
Thanks for your contribution!

PR #921, opened 2026-07-23
Add support for `used` attribute

GuillaumeGomez:
LLVM seems to handle "used" statics differently than GCC: you need to store them in a array manually whereas for GCC you just need to add the `used` attribute on the item.


bjorn3:
Does the GCC used attribute correspond to `used(compiler)` or `used(linker)` in rustc? The former only preserves it in object files and not optimize it out, while the latter tells the linker to preserve it in the linked executable/dylib.

GuillaumeGomez:
Thanks for the pointers @bjorn3! It now supports both. I also added a `cg_gcc` asm test to ensure that.

src/consts.rs
```diff
@@ -169,20 +169,31 @@ impl<'gcc, 'tcx> StaticCodegenMethods for CodegenCx<'gcc, 'tcx> {
...
 
-        if attrs.flags.contains(CodegenFnAttrFlags::USED_COMPILER)
-            || attrs.flags.contains(CodegenFnAttrFlags::USED_LINKER)
-        {
-            self.add_used_global(global.to_rvalue());
+        if attrs.flags.contains(CodegenFnAttrFlags::USED_LINKER) {
+            self.add_retained_global(global);
+        } else if attrs.flags.contains(CodegenFnAttrFlags::USED_COMPILER) {
+            self.add_used_global(global);
         }
```
antoyo, line 181: Could you please use the same code as cg_llvm (same function name as well)?
That will help to possibly remove duplication with it in the future.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustc_codegen_gcc/921/d9489816de25a5eef66e831fa59ea21ab6b0bdd9..bcea6f60955e5c193195bac8dd865fe9d1ee24ed)*

antoyo (approved):

PR #873, opened 2026-04-10
Fix ICE when calling transmuted function pointers

cijiugechu:
Indirect calls could keep the original libgccjit function-pointer type after a transmute instead of the ABI-lowered type expected at the call site, so ZST-elided arguments made libgccjit see the wrong arity and ICE. Cast the callee to the call-site function-pointer type before emitting the indirect call so argument checking and codegen use the normalized signature.

Closes: rust-lang/rustc_codegen_gcc#836

Issue #297, opened 2023-07-01
ICE: ` error: bitcast with types of different sizes`


matthiaskrgr:
```rust
#![allow(dead_code)]

#[repr(C)]
#[derive(Copy, Clone)]
pub struct Floats {
    a: f64,
    b: u8,
    c: f64,
}


mod rustrt {
    use super::{Floats};

    extern "C" {
        pub fn rust_dbg_abi_2(f: Floats) -> Floats;
    }
}

fn test2() {
    unsafe {
        let f = Floats { a: 1.234567890e-15_f64, b: 0b_1010_1010, c: 1.0987654321e-15_f64 };
        let ff = rustrt::rust_dbg_abi_2(f);
    }
}

pub fn main() {
    test2();
}

```

````
warning: unused variable: `ff`
  --> struct-return.rs:23:13
   |
23 |         let ff = rustrt::rust_dbg_abi_2(f);
   |             ^^ help: if this is intentional, prefix it with an underscore: `_ff`
   |
   = note: `#[warn(unused_variables)]` on by default

libgccjit.so: error: bitcast with types of different sizes
input expression (size: 64):
 <addr_expr 0x7fba6bc2c140
    type <pointer_type 0x7fba6bc2b540
        type <record_type 0x7fba6bc2b498  BLK
            size <integer_cst 0x7fba6c124120 constant 192>
            unit-size <integer_cst 0x7fba6c124150 constant 24>
            user align:64 warn_if_not_align:0 symtab:0 alias-set -1 canonical-type 0x7fba6bc28dc8 fields <field_decl 0x7fba6bc0ad10 field_0>
            pointer_to_this <pointer_type 0x7fba6bc2b540>>
        unsigned DI
        size <integer_cst 0x7fba6c118c90 constant 64>
        unit-size <integer_cst 0x7fba6c118ca8 constant 8>
        align:64 warn_if_not_align:0 symtab:0 alias-set -1 canonical-type 0x7fba6bc28d20>
    visited
    arg:0 <var_decl 0x7fba6bc275a0 stack_var_1 type <record_type 0x7fba6bc2b498 >
        addressable visited BLK (null):0:0 size <integer_cst 0x7fba6c124120 192> unit-size <integer_cst 0x7fba6c124150 24>
        align:64 warn_if_not_align:0 context <function_decl 0x7fba6bc1b500 _ZN13struct_return5test217h02d4cb3545bc47b6E>
[...]

PR #927, opened 2026-07-25
abi: Load on-stack arguments for direct calls

### LINKED ISSUE #829: ICE `bitcast with types of different sizes` with `#[repr(align))]`
````rust
#[repr(align(64))]
#[derive(Default)]
struct AlignedF32(f32);
extern fn write<T>(x : T, dest : &T) {}
fn check<T : Default>(x : T) {
  let out = T::default();
  write(x, &out)
}
fn main() { check(AlignedF32(6.0)) }
````

```
warning: `extern` declarations without an explicit ABI are deprecated
 --> code.rs:4:1
  |
4 | extern fn write<T>(x : T, dest : &T) {}
  | ^^^^^^ help: explicitly specify the "C" ABI: `extern "C"`
  |
  = note: `#[warn(missing_abi)]` on by default

warning: unused variable: `x`
 --> code.rs:4:20
  |
4 | extern fn write<T>(x : T, dest : &T) {}
  |                    ^ help: if this is intentional, prefix it with an underscore: `_x`
  |
  = note: `#[warn(unused_variables)]` (part of `#[warn(unused)]`) on by default

warning: unused variable: `dest`
 --> code.rs:4:27
  |
4 | extern fn write<T>(x : T, dest : &T) {}
  |                           ^^^^ help: if this is intentional, prefix it with an underscore: `_dest`

warning: field `0` is never read
 --> code.rs:3:19
  |
3 | struct AlignedF32(f32);
  |        ---------- ^^^
  |        |
  |        field in this struct
  |
  = help: consider removing this field
  = note: `#[warn(dead_code)]` (part of `#[warn(unused)]`) on by default

libgccjit.so: error: : bitcast with types of different sizes
input expression (size: 64):
 <addr_expr 0x7fbcf2739a60
    type <pointer_type 0x7fbcf27477e0
        type <array_type 0x7fbcf2747738 type <integer_type 0x7fbcfb9a41f8 signed char>
            BLK
            size <integer_cst 0x7fbcfb9b8420 constant 512>
            unit-size <integer_cst 0x7fbcfb9b8c18 constant 64>
            user align:512 warn_if_not_align:0 symtab:0 alias-set -1 canonical-type 0x7fbcf2747348 domain <integer_type 0x7fbcf27472a0>
            pointer_to_this <pointer_type 0x7fbcf27477e0>>
        unsigned DI
        size <integer_cst 0x7fbcfb9b8000 constant 64>
        unit-size <integer_cst 0x7fbcfb9b8018 constant 8>
        align:64 warn_if_not_align:0 symtab:0 alias-set -1 canonical-type 0x7fbcf2747540>
    visited
    arg:0 <var_decl 0x7fbcf2754098 stack_var_2 type <array_type 0x7fbcf2747738>
        addressable visited BLK (null):0:0 size <integer_cst 0x7fbcfb9b8420 512> unit-size <integer_cst 0x7fbcfb9b8c18 64>
        align:512 warn_if_not_align:0 context <function_decl 0x7fbcf2744700 _RINvCs5RvIIZZzctW_4code5checkNtB2_10AlignedF32EB2_>
        chain <var_decl 0x7fbcf2754000 stack_var_1 type <array_type 0x7fbcf27473f0>
            addressable visited BLK (null):0:0 size <integer_cst 0x7fbcfb9b8420 512> unit-size <integer_cst 0x7fbcfb9b8c18 64>
            align:512 warn_if_not_align:0 context <function_decl 0x7fbcf2744700 _RINvCs5RvIIZZzctW_4code5checkNtB2_10AlignedF32EB2_>>>>
requested type (size: 512):
 <record_type 0x7fbcf2743f18  BLK
    size <integer_cst 0x7fbcfb9b8420 type <integer_type 0x7fbcfb9a4000 bitsizetype> constant 512>
    unit-size <integer_cst 0x7fbcfb9b8c18 type <integer_type 0x7fbcfb9a4f18 sizetype> constant 64>
    user align:32 warn_if_not_align:0 symtab:0 alias-set -1 canonical-type 0x7fbcf2743f18
    fields <field_decl 0x7fbcf27463c0 field_0
        type <array_type 0x7fbcf27433f0 type <integer_type 0x7fbcf27439d8 int>
            BLK
            size <integer_cst 0x7fbcfb9b8078 constant 0>
            unit-size <integer_cst 0x7fbcfb9b8030 constant 0>
            user align:32 warn_if_not_align:0 symtab:0 alias-set -1 canonical-type 0x7fbcf2743498 domain <integer_type 0x7fbcf2743348>>
        BLK (null):0:0 size <integer_cst 0x7fbcfb9b8078 0> unit-size <integer_cst 0x7fbcfb9b8030 0>
        user align:32 warn_if_not_align:0 offset_align 256 decl_not_flexarray: 0 offset <integer_cst 0x7fbcfb9b8030 0> bit-offset <integer_cst 0x7fbcfb9b8078 0> context <record_type 0x7fbcf2743f18 >
        chain <field_decl 0x7fbcf2746460 field_1 type <real_type 0x7fbcfb9be888>
            SF (null):0:0
            size <integer_cst 0x7fbcfb9b8240 constant 32>
            unit-size <integer_cst 0x7fbcfb9b8258 constant 4>
            align:32 warn_if_not_align:0 offset_align 256 decl_not_flexarray: 0 offset <integer_cst 0x7fbcfb9b8030 0> bit-offset <integer_cst 0x7fbcfb9b8078 0> context <record_type 0x7fbcf2743f18 > chain <field_decl 0x7fbcf2746500 field_2>>>
    pointer_to_this <pointer_type 0x7fbcf2743738>>
  |
  = note:  "cc" "-m64" "/home/matthias/vcs/github/CRED/rustcHOrh7n/symbols.o" "<2 object files omitted>" "-Wl,--as-needed" "-Wl,-Bstatic" "<sysroot>/lib/rustlib/x86_64-unknown-linux-gnu/lib/{libstd-*,libpanic_unwind-*,libobject-*,libmemchr-*,libaddr2line-*,libgimli-*,libcfg_if-*,librustc_demangle-*,libstd_detect-*,libhashbrown-*,librustc_std_workspace_alloc-*,libminiz_oxide-*,libadler2-*,libunwind-*,liblibc-*,librustc_std_workspace_core-*,liballoc-*,libcore-*,libcompiler_builtins-*}.rlib" "-Wl,-Bdynamic" "-lgcc_s" "-lutil" "-lrt" "-lpthread" "-lm" "-ldl" "-lc" "-L" "/home/matthias/vcs/github/CRED/rustcHOrh7n/raw-dylibs" "-B<sysroot>/lib/rustlib/x86_64-unknown-linux-gnu/bin/gcc-ld" "-fuse-ld=lld" "-Wl,--eh-frame-hdr" "-Wl,-z,noexecstack" "-L" "<sysroot>/lib/rustlib/x86_64-unknown-linux-gnu/lib" "-o" "code" "-Wl,--gc-sections" "-pie" "-Wl,-z,relro,-z,now" "-nodefaultlibs"
  = note: some arguments are omitted. use `--verbose` to show all linker arguments
  = note: rust-lld: error: cannot open code.code.444a81fd07d3598a-cgu.0.rcgu.o: No such file or directory


error: aborting due to 1 previous error; 4 warnings emitted
```

### PR DESCRIPTION
**Pull request by Dnreikronos**
Fixes rust-lang/rustc_codegen_gcc#829

For an over-aligned argument, rustc selects `PassMode::Indirect { on_stack: true, .. }`. cg_gcc declares the GCC parameter as the aggregate itself and records its index in `on_stack_function_params`, but the call operand is still a pointer to rustc's stack slot.

The direct-call path wasn't reading those indices. It saw a pointer where the GCC function expected an aggregate, fell through to the generic bitcast path, and asked libgccjit to turn an 8-byte pointer into a 64-byte value. That's why libgccjit stops before it writes the object file.

Direct calls now follow the function-pointer path: cast the slot pointer if its backend pointee type differs, then load the aggregate value. imo, this is the right layer for the fix because ABI lowering already records the distinction here, and the callee side already uses the same `on_stack` metadata.

btw, the new `repr(align(64))` run test stays close to the original reproducer and makes sure the direct-call ABI path compiles for an over-aligned value.

### DIFF
--- a/tests/run/overaligned_arg.rs
+++ b/tests/run/overaligned_arg.rs
@@ -1,0 +1,26 @@
+// Compiler:
+//
+// Run-time:
+//   status: 0
+
+#![feature(no_core)]
+#![no_std]
+#![no_core]
+#![no_main]
+
+extern crate mini_core;
+
+#[repr(align(64))]
+struct AlignedF32(f32);
+
+#[inline(never)]
+extern "C" fn write<T>(value: T, destination: &mut T) {
+    *destination = value;
+}
+
+#[no_mangle]
+extern "C" fn main(_argc: i32, _argv: *const *const u8) -> i32 {
+    let mut output = AlignedF32(0.0);
+    write(AlignedF32(6.0), &mut output);
+    output.0 as i32 - 6
+}
--- a/tests/run/overaligned_arg.rs
+++ b/tests/run/overaligned_arg.rs
@@ -14,13 +14,11 @@ extern crate mini_core;
 struct AlignedF32(f32);
 
 #[inline(never)]
-extern "C" fn write<T>(value: T, destination: &mut T) {
-    *destination = value;
-}
+extern "C" fn write<T>(_value: T, _destination: &T) {}
 
 #[no_mangle]
 extern "C" fn main(_argc: i32, _argv: *const *const u8) -> i32 {
-    let mut output = AlignedF32(0.0);
-    write(AlignedF32(6.0), &mut output);
-    output.0 as i32 - 6
+    let output = AlignedF32(0.0);
+    write(AlignedF32(6.0), &output);
+    0
 }
--- a/src/builder.rs
+++ b/src/builder.rs
@@ -190,13 +190,32 @@ impl<'a, 'gcc, 'tcx> Builder<'a, 'gcc, 'tcx> {
             return Cow::Borrowed(args);
         }
 
+        let on_stack_param_indices =
+            self.on_stack_function_params.borrow().get(&func).cloned().unwrap_or_default();
         let casted_args: Vec<_> = param_types
             .into_iter()
             .zip(args.iter())
-            .map(|(expected_ty, &actual_val)| {
+            .enumerate()
+            .map(|(index, (expected_ty, &actual_val))| {
                 let actual_ty = actual_val.get_type();
                 if expected_ty != actual_ty {
-                    self.bitcast(actual_val, expected_ty)
+                    if on_stack_param_indices.contains(&index) {
+                        // The ABI passes this argument by value, while rustc gives us a pointer to
+                        // its stack slot. Cast the pointer when the slot uses a different backend
+                        // type, then load the value expected by the function parameter.
+                        if let Some(pointee_ty) = actual_ty.get_pointee()
+                            && pointee_ty != expected_ty
+                        {
+                            self.context
+                                .new_cast(self.location, actual_val, expected_ty.make_pointer())
+                                .dereference(self.location)
+                                .to_rvalue()
+                        } else {
+                            actual_val.dereference(self.location).to_rvalue()
+                        }
+                    } else {
+                        self.bitcast(actual_val, expected_ty)
+                    }
                 } else {
                     actual_val
                 }

### CODE REVIEW

File: tests/run/overaligned_arg.rs [hunk 1]
```diff
@@ -0,0 +1,24 @@
+// Compiler:
```
antoyo, line 1: This test passes on the master branch and the original reproducer from https://github.com/rust-lang/rustc_codegen_gcc/issues/829#issue-3851194500 seems to work on the master branch as well, so it seems the issue is already fixed on the master branch.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustc_codegen_gcc/927/d1f73f59a2cf60126036b28a4feaf29640beb346..461ea2190be53a17bc0fbd575af5af20b3201771)*

Dnreikronos: Yep, you’re right. Both the new test and the original #829 reproducer pass on the current master. I didn’t notice that when I opened the PR.

btw, the direct-call path still has the old bitcast behavior, so it looks like some rustc or libgccjit update made the issue disappear. idk which one yet.

imo, we shouldn’t merge an ABI fix when the test already passes without it. I can bisect the rustc and libgccjit updates to find out what changed, or close the PR if you don’t think it’s worth chasing. ltm what you prefer.

antoyo: This was likely the same issue as #297 and #711, which was likely fixed by #873.
If you think there's still something to fix, you can continue your work on this PR.
But I'm thinking we should probably remove the direct call code since it's barely used.

Otherwise, we can close this PR.

Dnreikronos: yep, I will close it!
Ty for the guidance, @antoyo!

PR #929, opened 2026-07-25
Automatically install `rustfmt` if not installed when running `y.sh fmt` command

GuillaumeGomez:


build_system/src/fmt.rs
```diff
@@ -31,7 +33,29 @@ pub fn run() -> Result<(), String> {
...
+            && let Some(line) = iter.next()
+            && line.contains("run `rustup component add")
+            && let Some(cmd) = line.split('`').nth(1)
+        {
+            println!("`rustfmt` is not installed for this toolchain, installing it...");
+            // A weird round-about way to get a `&&str` so I can get a `&dyn AsRef<OsStr>` but
+            // as long as it works...
+            let cmd = cmd.split(' ').collect::<Vec<_>>();
+            let cmd = cmd.iter().map(|s: &&str| s as &dyn AsRef<OsStr>).collect::<Vec<_>>();
+            run_command_with_output(cmd.as_slice(), Some(Path::new(".")))?;
```
antoyo, line 50: Wouldn't it be simpler to hard-code the `rustfmt` command instead of parsing the error message which could change in the future?

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustc_codegen_gcc/929/b544d99c6476e92e1b55bb826ec009fb0bc7622e..7a90d915581bceba12fa67fcd56a9f66e7a34e76)*

GuillaumeGomez: I need to provide the toolchain, which is provided by the error (I could parse our `rust-toolchain` file, but lazyness++).

antoyo (approved):

Issue #930, opened 2026-07-26
Quickstart can't build hello world: librustc_codegen_gcc.so not found


xobs:
I'm trying to build librustc_codegen_gcc.so from this repo (rather than from Rust) to make it easier to upstream patches. However, the Quickstart instructions do not appear to work.

Following the instructions, I've:

1. Copied `config.example.toml` to `config.toml`
2. Ran `./y.sh prepare`
3. Ran `./y.sh build --sysroot --release`

Now I'm trying to run the fourth step:

```
$ ./y.sh cargo build --manifest-path tests/hello-world/Cargo.toml
[BUILD] build system
    Finished [`release` profile [optimized]](https://doc.rust-lang.org/cargo/reference/profiles.html#default-profiles) target(s) in 0.01s
Using `/home/seancross/Code/rustc_codegen_gcc/build/libgccjit/b794cb99508e7eccd40fa42bb310d5781e09899b` as path for libgccjit
   Compiling hello_world v0.0.0 (/home/seancross/Code/rustc_codegen_gcc/tests/hello-world)
error: couldn't load codegen backend /home/seancross/Code/rustc_codegen_gcc/target/debug/librustc_codegen_gcc.so: cannot open shared object file: No such file or directory

error: could not compile `hello_world` (bin "hello_world") due to 1 previous error
$
```

It appears to be looking for something under `target/debug` even though the instructions say to run `./y.sh build --sysroot --release`. If I re-run step 3 with `./y.sh build --sysroot` without `--release`, then attempting to build `hello-world` gives a different error:

```
$ ./y.sh cargo build --manifest-path tests/hello-world/Cargo.toml
[BUILD] build system
    Finished [`release` profile [optimized]](https://doc.rust-lang.org/cargo/reference/profiles.html#default-profiles) target(s) in 0.00s
Using `/home/seancross/Code/rustc_codegen_gcc/build/libgccjit/b794cb99508e7eccd40fa42bb310d5781e09899b` as path for libgccjit
   Compiling hello_world v0.0.0 (/home/seancross/Code/rustc_codegen_gcc/tests/hello-world)
error[E0463]: can't find crate for `mylib`
 --> src/main.rs:1:5
  |
1 | use mylib::my_func;
  |     ^^^^^ can't find crate

For more information about this error, try `rustc --explain E0463`.
error: could not compile `hello_world` (bin "hello_world") due to 1 previous error
$
```

antoyo:
[...]

PR #931, opened 2026-07-26
int: implement signed comparison for unsigned integers

xobs:
Implement signed comparison for unsigned integers, which is very similar to unsigned integers but uses `to_signed()` rather than `to_unsigned()`. This fixes a case where niche optimization causes the discriminant to sometimes compare signed and unsigned values.

This was caught when running a test suite for an embedded device compiled under the gcc backend. A reproducer when compiling under -O for x86-64 is:

```rust
use std::hint::black_box;

#[repr(u8)]
enum Inner { L0 = 0, H255 = 255 }
enum O { A(Inner), B, C }

#[inline(never)]
fn which(o: &O) -> &'static str {
    match o { O::A(_) => "a", O::B => "b", O::C => "c" }
}

fn main() {
    assert_eq!(which(black_box(&O::A(Inner::H255))), "a"); // 0xFF
    println!("all ok");
}
```

The problem only shows up with `-O`, not by default.

```
thread 'main' (1223) panicked at /home/seancross/work/testcase/testcase-3.rs:13:5:
assertion `left == right` failed
  left: "c"
 right: "a"
```

### DIFF
--- a/src/int.rs
+++ b/src/int.rs
@@ -462,9 +462,15 @@ impl<'a, 'gcc, 'tcx> Builder<'a, 'gcc, 'tcx> {
                     lhs_high = self.context.new_cast(self.location, lhs_high, unsigned_type);
                     rhs_high = self.context.new_cast(self.location, rhs_high, unsigned_type);
                 }
-                // FIXME(antoyo): we probably need to handle signed comparison for unsigned
-                // integers.
-                _ => (),
+                IntPredicate::IntSGT
+                | IntPredicate::IntSGE
+                | IntPredicate::IntSLT
+                | IntPredicate::IntSLE => {
+                    let signed_type = native_int_type.to_signed(self.cx);
+                    lhs_high = self.context.new_cast(self.location, lhs_high, signed_type);
+                    rhs_high = self.context.new_cast(self.location, rhs_high, signed_type);
+                }
+                IntPredicate::IntEQ | IntPredicate::IntNE => (),
             }
 
             let condition = self.context.new_comparison(
@@ -602,9 +608,17 @@ impl<'a, 'gcc, 'tcx> Builder<'a, 'gcc, 'tcx> {
                         rhs = self.context.new_cast(self.location, rhs, unsigned_type);
                     }
                 }
-                // FIXME(antoyo): we probably need to handle signed comparison for unsigned
-                // integers.
-                _ => (),
+                IntPredicate::IntSGT
+                | IntPredicate::IntSGE
+                | IntPredicate::IntSLT
+                | IntPredicate::IntSLE => {
+                    if !a_type.is_vector() {
+                        let signed_type = a_type.to_signed(self.cx);
+                        lhs = self.context.new_cast(self.location, lhs, signed_type);
+                        rhs = self.context.new_cast(self.location, rhs, signed_type);
+                    }
+                }
+                IntPredicate::IntEQ | IntPredicate::IntNE => (),
             }
             self.context.new_comparison(self.location, op.to_gcc_comparison(), lhs, rhs)
         }
--- a/tests/run/int.rs
+++ b/tests/run/int.rs
@@ -319,4 +319,29 @@ fn main() {
         const VAL5: T = 73236519889708027473620326106273939584_i128;
         check_ops128!();
     }
+
+    {
+        #[allow(dead_code)]
+        #[repr(u8)]
+        enum Inner {
+            L0 = 0,
+            H255 = 255,
+        }
+        #[allow(dead_code)]
+        enum O {
+            A(Inner),
+            B,
+            C,
+        }
+
+        #[inline(never)]
+        fn which(o: &O) -> &'static str {
+            match o {
+                O::A(_) => "a",
+                O::B => "b",
+                O::C => "c",
+            }
+        }
+        assert_eq!(which(black_box(&O::A(Inner::H255))), "a");
+    }
 }


xobs:
Let me know if you'd like me to open an issue in the `rust` repo, or if there's a place I can add a testcase here. I'm manually driving `rustc` due to #930.

I have a few other minor fixes, and I can send PRs with them once I know how you'd prefer them formatted.

src/int.rs
```diff
@@ -602,9 +602,17 @@ impl<'a, 'gcc, 'tcx> Builder<'a, 'gcc, 'tcx> {
...
-                // integers.
-                _ => (),
+                IntPredicate::IntSGT
+                | IntPredicate::IntSGE
+                | IntPredicate::IntSLT
+                | IntPredicate::IntSLE => {
+                    if !a_type.is_vector() {
+                        let signed_type = a_type.to_signed(self.cx);
+                        lhs = self.context.new_cast(self.location, lhs, signed_type);
+                        rhs = self.context.new_cast(self.location, rhs, signed_type);
```
antoyo, line 618: Please add the same code for non-native integers around line 465.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustc_codegen_gcc/931/b544d99c6476e92e1b55bb826ec009fb0bc7622e..bc6d9786022bc565b56f79500e15572a8d4f8955)*

antoyo:
> Let me know if you'd like me to open an issue in the `rust` repo, or if there's a place I can add a testcase here. I'm manually driving `rustc` due to #930.

No need to open an issue: you can open a PR right away.
You can add a test in [this directory](https://github.com/rust-lang/rustc_codegen_gcc/tree/master/tests/run).

> I have a few other minor fixes, and I can send PRs with them once I know how you'd prefer them formatted.

Anything else you need to know about my preferences here?

Thanks for your contribution!

src/int.rs
xobs: Alright, I've added it there.

tests/run/int.rs
```diff
@@ -319,4 +319,29 @@ fn main() {
         const VAL5: T = 73236519889708027473620326106273939584_i128;
         check_ops128!();
     }
+
+    {
+        #[allow(dead_code)]
```
antoyo, line 324: Is this needed because L0 is never constructed?

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustc_codegen_gcc/931/b544d99c6476e92e1b55bb826ec009fb0bc7622e..5c533bc664a4003a47828b0b82bae328252668f0)*

xobs:
That should be it. Now I know how to add and run tests, I can use that to make sure the changes I push actually fix things.

There are a few bigger changes I've accumulated (LTO improvements, link section fixes, clobber support), and a couple of them require patches to gcc (and consequently `gccjit` support), but I'll start with these small patches first.

tests/run/int.rs
xobs: It's needed because not all of the enum variants are constructed. However, when I added that I also didn't have libubsan installed, so I saw both errors and warnings in the compile failure.

It's not strictly speaking _needed_, so I can remove it to make the test easier to follow.

antoyo: Please keep it as it.
I just wanted to make sure.

xobs:
Hmm... The test doesn't seem to work without int128. Or on m68k. Think that is a real bug, or should I gate the test on x86 only?

antoyo:
> Hmm... The test doesn't seem to work without int128. Or on m68k. Think that is a real bug, or should I gate the test on x86 only?

It seems like a real bug to me.
Without int128, the error is:
```
gcc_jit_context_new_cast: cannot cast param0[(int)1] from type: __int64_t  __attribute__((aligned(8))) to type: __int64_t  __attribute__((aligned(8)))[2]
```
which means it seems you try to cast a low/high value to the complete 128-bit representation (which is an array of i64).
I guess the code needs to be adjusted to call `to_signed` on the inner element type instead of the array type.

xobs:
Yep, sorry, I just needed to figure out how to use `gcc-15-without-int128.deb`. Now I see I just need to extract it somewhere and point config.toml:gcc-path at it.

Hopefully works now. So happy CI and test coverage exists.

src/int.rs
```diff
@@ -462,9 +462,17 @@ impl<'a, 'gcc, 'tcx> Builder<'a, 'gcc, 'tcx> {
...
                     rhs_high = self.context.new_cast(self.location, rhs_high, unsigned_type);
                 }
-                // FIXME(antoyo): we probably need to handle signed comparison for unsigned
-                // integers.
-                _ => (),
+                IntPredicate::IntSGT
+                | IntPredicate::IntSGE
+                | IntPredicate::IntSLT
+                | IntPredicate::IntSLE => {
+                    if !a_type.is_vector() {
```
antoyo: For non-native integers, `a_type` will be an array, so this condition is not necessary, I believe.

antoyo (commented):
One last thing, then we can merge this:

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustc_codegen_gcc/931/3aea6dce9935004372b634ff7cff896c26c73137..2a023ce4bdb02989039624f334efb74f03f5f291)*

src/int.rs
xobs: Alright, I've never hit that condition so I'm not sure how I'd trigger it. But I've removed the vector check and updated the PR.

antoyo: I still see the condition.
Did you forget to push it?

xobs: Thanks, yeah. `jj git push` didn't automatically push to the remote, and just said "Nothing changed". I forced it by specifying the remote. (It's a good tool, just still getting used to it.)

antoyo:
Thanks for your contribution!

PR #932, opened 2026-07-26
asm: seed the in_value for inout

xobs:
When performing an "inout" operation, the "in" value isn't assigned. This results in incorrect asm being generated. Make sure the value is assigned to the input register.


antoyo (approved):

antoyo:
Thanks for your contribution!

PR #935, opened 2026-07-31
Only keep minimum required info for `cpuid.def`

GuillaumeGomez:
From this [zulip discussion](https://rust-lang.zulipchat.com/#narrow/channel/231349-licensing/topic/Intel.20Simplified.20Software.20License/with/613705051).


GuillaumeGomez:
Ah right, gonna need to make the other update (removed a libgccjit API) before CI can pass again.

antoyo (approved):

PR #920, opened 2026-07-23
gcc: locate codegen-backends via host_tuple()

xobs:
`file_path` builds the `lib/rustlib/<host>/codegen-backends` path used to load the codegen backend. It passed `&sess.host.llvm_target`, but that is the host's LLVM target string, not its rustc target tuple.  The two can differ (e.g. `arm64-apple-macosx` vs `aarch64-apple-darwin`), so the lookup can miss the directory that actually holds the backend. Use `rustc_session::config::host_tuple()` instead.

This fixes an issue when building custom toolchains. In particular, I'm using the gcc backend to build a custom SH4 compiler. Without this patch, this command fails to find the backend:

```
$ BOOTSTRAP_SKIP_TARGET_SANITY=1 python3 x.py install library --target sh4-dreamcast-none-elf
```

PR #954, opened 2026-08-09
Use rust target triple instead of LLVM's

antoyo:


bjorn3:
This would break with custom targets. `TargetTuple::tuple()` returns the basename of the json file for custom targets, which has no relation to the underlying target that the codegen backend would generate code for.

antoyo:
> This would break with custom targets. `TargetTuple::tuple()` returns the basename of the json file for custom targets, which has no relation to the underlying target that the codegen backend would generate code for.

Using the LLVM target caused [issues](https://github.com/rust-lang/rustc_codegen_gcc/pull/920) in the past.
Do you mean we should fix something on the bootstrap side as well?

bjorn3:
I'm not sure there is a great fix. I guess keeping this code as is and accepting that libgccjit is found based on the json filename works for now.

antoyo:
> I'm not sure there is a great fix. I guess keeping this code as is and accepting that libgccjit is found based on the json filename works for now.

Isn't the current code broken for all arch where the LLVM target triple is different than Rust's target triple?

bjorn3:
I meant keeping the code in this PR as is and merging this PR.

antoyo:
Thanks.

PR #957, opened 2026-08-15
Add a new environment variable for adding a custom libgccjit path

techmetx11:
This PR adds a environment variable, that lets the user choose where should the driver look for `libgccjit` in.

This is better (in my opinion) than making a symlink in the Rust toolchain directory, because you don't have to assume which toolchain and host CPU architecture the user is using

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -220,6 +220,10 @@ impl CodegenBackend for GccCodegenBackend {
             && let Ok(true) = fs::exists(custom_path)
         {
             load_libgccjit_if_needed(&PathBuf::from(custom_path));
+
+            if !gccjit::is_loaded() {
+                println!("Could not load gccjit from CG_GCCJIT_LIBRARY_PATH: {}.", custom_path);
+            }
         }
 
         // We use all_paths() instead of only path() in case the path specified by --sysroot is
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -222,7 +222,7 @@ impl CodegenBackend for GccCodegenBackend {
             load_libgccjit_if_needed(&PathBuf::from(custom_path));
 
             if !gccjit::is_loaded() {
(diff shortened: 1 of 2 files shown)


src/lib.rs
```diff
@@ -210,6 +210,10 @@ impl CodegenBackend for GccCodegenBackend {
                     lib_path.join(sess.target.llvm_target.as_ref()).join("libgccjit.so");
                 paths.push(llvm_target_path);
             }
+            if let Ok(custom_path) = env::var("CG_GCCJIT_SEARCH_PATH").as_deref() {
+                paths.push(PathBuf::from(custom_path).join("libgccjit.so"));
```
antoyo: Since this is added last, this means that if a library is found in the sysroot, the library from the env var won't be used.
I believe it would make more sense to use the one from the env var if it is exists.
Doing that means we can probably move that check outside this function.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustc_codegen_gcc/957/c3199014914275ad92037e145267ca9ec31ec976..f6d8d38e4837e063a412f99e03195478b6b8ef07)*

antoyo: Do you think it would be better to have the env var point directly at the library file?
Doing so, we could support libraries named `libgccjit.so.0`.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustc_codegen_gcc/957/c3199014914275ad92037e145267ca9ec31ec976..f6d8d38e4837e063a412f99e03195478b6b8ef07)*

techmetx11: I think that would be better.

techmetx11: Oh yeah, I haven't thought of that. I'll move the check outside the nested function.

PR #958, opened 2026-08-15
Ignore code model setting for m68k

techmetx11:
`m68k-unknown-none-elf`'s target spec has [the code model set to `medium`](https://github.com/rust-lang/rust/blob/main/compiler/rustc_target/src/spec/targets/m68k_unknown_none_elf.rs#L15) which causes rustc_codegen_gcc to insert `-mcmodel=medium` in the GCC command-line options. Since GCC m68k doesn't support this option, it prematurely errors with `libgccjit.so: error: unrecognized command-line option '-mcmodel=medium'`

This PR fixes the issue by making rustc_codegen_gcc ignore the code model setting if the target architecture is m68k

### DIFF
--- a/src/gcc_util.rs
+++ b/src/gcc_util.rs
@@ -204,13 +204,16 @@ pub fn new_context<'gcc>(sess: &Session) -> Context<'gcc> {
     if let Some(model) = sess.code_model() {
         use rustc_target::spec::CodeModel;
 
-        context.add_command_line_option(match model {
-            CodeModel::Tiny => "-mcmodel=tiny",
-            CodeModel::Small => "-mcmodel=small",
-            CodeModel::Kernel => "-mcmodel=kernel",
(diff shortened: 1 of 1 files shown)


antoyo:
Would it make more sense to fix the issue upstream in the Rust repo?
(i.e. can cg_llvm set the code model on that target?)
Perhaps by removing the code model from this target spec?

techmetx11:
> Would it make more sense to fix the issue upstream in the Rust repo? (i.e. can cg_llvm set the code model on that target?) Perhaps by removing the code model from this target spec?

Probably would, since the `m68k-unknown-linux-gnu` target spec has no code model set. I don't know why the code model is set specifically for this target

antoyo:
I see you opened an issue upstream.
If this gets rejected, please ping me here so that I can take another look.

PR #959, opened 2026-08-21
Sync from rust 2026/08/21

antoyo:


antoyo:
@folkertdev: The CI hit the [error you added here](https://github.com/rust-lang/rust/pull/160715/changes#diff-44ed94419ab7173a0a684b47ac44fb45e37caca8a7ab23e2fe49313c5797e6d3R874).
Does your PR respect the `--sysroot` flag?

folkertdev:
I don't know so I'd guess not? I just look at the path that showed up in the error messages.

antoyo:
> I don't know so I'd guess not? I just look at the path that showed up in the error messages.

Do you know how to fix this in compiletest?
I believe it would make more sense to first look in the sysroot provided by `--sysroot` if the flag is present.

folkertdev:
Hmm actually looking into it, that change uses 

```rust
    let gcc_supported_target_tuples = match default_codegen_backend {
        CodegenBackend::Gcc => {
            directives::find_gcc_supported_targets(&args.sysroot_base, &args.host)
        }
        CodegenBackend::Llvm | CodegenBackend::Cranelift => vec![],
    };
```

which is documented as 

```rust
    /// Directory containing the compiler sysroot.
    #[arg(long)]
    sysroot_base: Utf8PathBuf,
```

I'm not really sure what to look for here beyond that. What is the path you expect the sysroot to be at? For reference the code currently tries to look at 

```
/home/runner/.rustup/toolchains/nightly-2026-08-21-x86_64-unknown-linux-gnu/lib/rustlib/x86_64-unknown-linux-gnu/codegen-backends/lib
```

antoyo:
> What is the path you expect the sysroot to be at?

We provide our own sysroot at this path: `<cg_gcc repo root>/build/build_sysroot/sysroot`.

folkertdev:
Hmm, I think the answer is to use `query_rustc_output` to get the actual sysroot, and also check for `libgccjit.so` there. 

@jieyouxu does that sound right? If so, what would be the best way to go about that? During `Config` construction, or just delay and add e.g. a `gcc_supported_target_tuples` method on `Config` that can do the rustc query and then call `find_gcc_supported_targets`. 

I'm not really sure whether the sysroot rustc reports is sufficient or whether we should still also check the `sysroot_base`.

jieyouxu:
(I'll try to take a look on Friday)

PR #962, opened 2026-08-23
Fix EII

AsakuraMizu:
Fix support for EII functions with defaults and EII statics. Now all `tests/ui/eii` tests pass.

### DIFF
--- a/src/mono_item.rs
+++ b/src/mono_item.rs
@@ -153,6 +153,11 @@ impl<'gcc, 'tcx> CodegenCx<'gcc, 'tcx> {
 
         attributes::from_fn_attrs(self, fn_decl, instance, Some(fn_abi));
 
+        #[cfg(feature = "master")]
+        if linkage == Linkage::WeakAny {
+            fn_decl.add_attribute(FnAttribute::Weak);
+        }
+
         // If we're compiling the compiler-builtins crate, e.g., the equivalent of
         // compiler-rt, then we want to implicitly compile everything with hidden
         // visibility as we're going to link this object all over the place but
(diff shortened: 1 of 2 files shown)


antoyo:
It seems there's some interactions with the inline attribute that should be handled as well.
For instance:
```rust
#![crate_type = "rlib"]
#![feature(extern_item_impls)]
#[eii(eii1)]
#[inline]
pub fn decl1(x: u64) {
    println!("default {x}");
}
```
panics with this warning:
```
libgccjit.so: warning: : inline function ‘_RNvCslJaV4vr0Uxj_9test_rust5decl1’ declared weak [-Wattributes]
```
Please add this as a test in `tests/run` and fix this issue.

AsakuraMizu:
> It seems there's some interactions with the inline attribute that should be handled as well.

Hi, thanks for your review.

I investigated this further and found that this isn't only a attribute interaction issue with cg_gcc; it could also lead to unexpected behavior on cg_llvm. I've reported this upstream: https://github.com/rust-lang/rust/issues/161650.

bjorn3 suggested denying `#[inline]` on EII declarations altogether: https://github.com/rust-lang/rust/issues/161650#issuecomment-5393156891

Therefore if we add this check in rustc, the given example will produce a compile error before reaching cg_gcc instead of triggering an ICE.

Would you prefer to wait for the upstream fix for now?

antoyo (approved):

antoyo:
I'm OK with merging this PR now.
Is this ready to merge?

AsakuraMizu:
> Is this ready to merge?

Yes it's ready. Thanks! :smiley:

antoyo:
Thanks for your contribution!

GuillaumeGomez:
Nice work!

